PR #78, opened 2015-01-12
Aragorn output to useful GFF3

erasche:
This is really awful horrible code, I'm aware. However, their output is likewise, really awful and horrible.

We make the decision to run aragorn twice, once with forced output formatting if GFF3 is requested, as it's easier than the alternative. (The alternative is, of course, to run once with 'full' format, and then generate the fasta version if they'd rather have that from the 'full').


bgruening:
Nice thanks.

PR #87, opened 2015-02-06
Fix uniq.xml typo by removing redundant help value

peterjc:
On a related point, the README file says "the uniq.xml and the sort_rows tools are moved to the unix tools" but I'm not sure where that is?


nsoranzo:
@peterjc I think this refer to text_processing/text_processing/unsorted_uniq.xml

peterjc:
Thanks @nsoranzo - I agree.

bgruening:
Thanks @peterjc and @nsoranzo. I will for a little bit more feedback on these tools and will release a new version soon.

PR #89, opened 2015-02-11
Fix bug in bismark_methylation_caller.xml

cche:
A # symbol in front of --bismark_path messed up the command line so the caller didn't run in galaxy.

All the best
Cristian


bgruening:
Thanks!

PR #95, opened 2015-03-19
Add new tool to find subsequences and output the regions in BED format

bgruening:
This PR will add one new tool to find for example all restrictions enzymes as requested by Ralf Gilsbach.
@smithcr @TorHou please review! :)


erasche:
we have a tool to run restriction digests on DNA, would that be something you'd want in here? (It also produces SVG diagrams like ApE). It didn't feel like a great fit for tools-iuc since that's mostly large suite and this is a "home made" tool.

bgruening:
@erasche exactly we want to use it for restriction enzymes, but not only for this use-case. In MethylC-seq your are also interested in finding CG, or CT and so on ...

My plan for later was to add a select-box with predefined restriction enzymes but offer the option to put a user-string in as well.

If your tool is superior I'm for sure want to use it :)

erasche:
@bgruening our's is here: https://github.com/erasche/restriction-digest from one of my undergrads last semester. We actually use the entire [NEB REBASE](https://github.com/erasche/restriction-digest/blob/master/dnadigest/rebase.yaml) database for doing digestions. A lot like your select box, but with 200+ choices (and we've listed all [isoschizomers](https://en.wikipedia.org/wiki/Isoschizomer) in case anyone knows them by the other names).

bgruening:
@erasche this is a monster! But a very nice one! 
I feel this is a little bit to much of an overhead for what we are going to do. Do you really need such dependencies? I was thinking of include rebase, but was running our of time. I think adding a simple selectbox should do the job, isn't it?

erasche:
Sure, if it's too much overhead for what you want, feel free to not use it. If you want it included with this repo, I'll be happy to PR it, that's all. 

We could definitely cut down the enzyme list though, we just wanted to include as much as possible the first time through. I don't do bench work, I have no clue which enzymes are common/etc.

"Entire Rebase" is actually incorrect:

- enzymes without cut sites were removed (~3k/4k)
- cut sites that were different on 5' to 3' strands were removed (so things like `5' NNNN^T 3'`, `3' NNN^NA 5'` types were ignored)
- list was further collapsed by creating a giant dict based on cut site signature, and then outputting that.

bgruening:
@erasche is it ok for you if I use your rebase.yaml and extract with the following snippet all enzymes+recognition_sequence (5')?

`[(key,a[key]['recognition_sequence'][0].split()[1]) for key in a.keys() if key == a[key]['enzyme']]`

I think is would be enough to make this small tool really usable.
Thanks!

Btw. if you want to include your tool here, feel free to do so. I also plan to add a few more annotation tools.

erasche:
Please! Go ahead! :)

tor. 19. mar. 2015 kl. 18.13 skrev Björn Grüning <notifications@github.com>:

erasche:
https://github.com/erasche/restriction-digest/blob/master/scripts/rebase.py was used to parse the rebase database formatted for bioperl. I don't remember exactly which file that was but it was similarly formatted to ftp://ftp.neb.com/pub/rebase/withref.503

PR #97, opened 2015-04-08
Added install recipes for bioformats version 1.0.4, imagej version 1.48

gregvonkuster:
and javabridge version 1.0.11.  I didn't add .shed.yml files because I don't see them in your galaxy tools repo.  Let me know if you would like them added.  I've set the owners to iuc, but will change to bgruening if you'd rather have that - just let me know.  Also, please let me know if you see any issues and I'll get them fixed up. I've tested these with a simple tool, but will be enhancing the tools from here.


packages/package_bioformats_1_0_4/tool_dependencies.xml
```diff
@@ -0,0 +1,40 @@
...
+                    <repository name="package_javabridge_1_0_11" owner="iuc">
+                        <package name="javabridge" version="1.0.11" />
+                    </repository>
+                </action>
+                <action type="shell_command">
+                    export PYTHONPATH=$PYTHONPATH:$INSTALL_DIR/lib/python &amp;&amp; 
+                    python setup.py install --install-lib $INSTALL_DIR/lib/python
+                </action>
+                <action type="set_environment">
+                    <environment_variable action="prepend_to" name="PYTHONPATH">$INSTALL_DIR/lib/python</environment_variable>
```
bgruening: Can you add a BIOFORMATS_ROOT_PATH to the environment variables? This is kind of a standard know, to access to base directory from other packages if needed. Some holds for the other packages.

bgruening:
This look very nice. Thanks Greg! IUC for packages is fantastic, we can use `galaxytools` as incubator and migrate it over if the IUC repository is interested in image processing tools.

bgruening:
@gregvonkuster do you think `bioformats` is a to general name? Maybe `java_bioformats`? Not sure just want to keep the TS clean.

packages/package_bioformats_1_0_4/tool_dependencies.xml
gregvonkuster: Sure!  I'll also rename to python_bioformats if you agree.  Probably a better name since it downloads the python-bioformats module.  https://pypi.python.org/pypi/python-bioformats  Let me know if you'd rather name it java_bioformats as you stated below.

erasche:
@bgruening there's a section for imaging in https://github.com/galaxyproject/tools-iuc/issues/33 so I'd say we're interested :)

packages/package_bioformats_1_0_4/tool_dependencies.xml
bgruening: @gregvonkuster sure python_bioformats sounds good. it's just a gut feeling that `bioformats` is to general.

bgruening:
@erasche; I forgot this issue completely. @gregvonkuster will try to include ImageJ, if this works out I'm happy to migrate it.

gregvonkuster:
Yeah, missed some files - still too used to hg.  I hope this is better.

bgruening:
Nice thanks a lot!

PR #99, opened 2015-04-12
Trim galore update

bgruening:
Changes
========
* update to latest version
* use data collection
* include tests
* more parameters
* best practice updates

@jmchilton I think I abused a `paired data collection` to create a `unpaired` `data collection` ;). 
If this still looks ok to you I will try to fix bowtie2.


tools/trim_galore/trim_galore_wrapper.xml
```diff
@@ -248,97 +208,144 @@
...
-                  <option type="from_param" name="singlePaired.input_singles" param_attribute="ext" />
-                </action>
-          </actions>
         </data>
 
-        <data format="fastq" name="trimmed_reads_pair1" label="${tool.name} on ${on_string}: trimmed reads pair 1">
+        <collection name="trimmed_reads_paired_collection" type="paired" label="${tool.name} on ${on_string}: trimmed reads">
+            <filter>singlePaired['sPaired'] == "paired_collection"</filter>
+            <data name="forward" format="fastqsanger" from_work_dir="input_mate1_val_1.fq" />
+            <data name="reverse" format="fastqsanger" from_work_dir="input_mate2_val_2.fq" />
```
jmchilton: It is possible to derive this from the inputs without requiring an action. Do you want to support multiple formats?

tools/trim_galore/trim_galore_wrapper.xml
```diff
@@ -248,97 +208,144 @@
...
+            <filter>
+            ((
+                params['settingsType'] == "custom" and
+                params['retain_unpaired']['settingsType'] == "retain_unpaired_output" and
+                singlePaired['sPaired'] == "paired_collection"
+            ))
+            </filter>
+            <data name="forward" format="fastqsanger" from_work_dir="input_mate1_unpaired_1.fq" />
+            <data name="reverse" format="fastqsanger" from_work_dir="input_mate2_unpaired_2.fq" />
+        </collection>
```
jmchilton: A paired collection of unpaired reads - not sure if this better than individual outputs - it may be though I really don't know.

tools/trim_galore/trim_galore_wrapper.xml
bgruening: I removed the support for multiple file formats. Fastqsanger is nowadays standard, if there are old files, these can be converted. So I guess it's fine as it is now and makes the wrapper more maintainable.

jmchilton: Great, if you ever need the trick let me know though.

tools/trim_galore/trim_galore_wrapper.xml
bgruening: Yes, this scares me as well. But the alternative is to have many many files in your history with strange labels and no way to track it back.

bgruening: @nsoranzo do you have any advice here? We will have the same problem in bowtie2/bwa I guess.

nsoranzo: I think this is fine since this paired collection is created only if the input reads where already in a paired collection.

bgruening: Ok, if no one complains I will merge this :)

Thanks!

bgruening:
@jmchilton which of these changes will prevent `trim galore` to work in workflow mode? Trim galore shows up but is grey and not click able.

jmchilton:
Yeah - the collection output prevents that. Sorry I didn't warn about that sooner - I really need to fix that in this release but am running out of time...

PR #100, opened 2015-04-14
OpenMS 2.0

TorHou:
A few tools were deleted in OpenMS 2.0.
- IDEvaluator
- ImageCreator
- Resampler
 At least they were not in the list of binaries of:
http://share.gruenings.eu/openms-2.0-Linux-x86_64.tar.gz


bgruening:
Thanks!

PR #110, opened 2015-05-05
Add .shed.yml files for package_fiji_20141125, package_imagej_1_48,

gregvonkuster:
package_javabridge_1_0_11 and package_python_bioformats_1_0_4.  Also add
an additional environment variable for the package_imagej_1_48 recipe
(although the recipe may not be used since we now have fiji, the
environment variable is very usefule).


bgruening:
&#x1f60e; very nice!

PR #111, opened 2015-05-05
Add version 1.0.0 of the imagej2_convert_format tool.

gregvonkuster:
Let me know if you see any issues, and I'll take care of them asap.  Thanks!


tools/image_processing/imagej2/imagej2_convert_format/imagej2_convert_format.xml
```diff
@@ -0,0 +1,66 @@
+<?xml version='1.0' encoding='UTF-8'?>
+<tool id="imagej2_convert_format" name="Convert image format" version="1.0.0">
+    <description>with bioformats</description>
+    <macros>
+        <import>imagej2_macros.xml</import>
+    </macros>
+    <expand macro="requirements" />
+    <command interpreter="python">
+        imagej2_convert_format.py
+        -i $input
```
bgruening: The `$input` and `$output` needs to be quoted, the parameters doesn't need to. More or less all strings, paths needs to be quoted.

tools/image_processing/imagej2/imagej2_convert_format/imagej2_convert_format.xml
```diff
@@ -0,0 +1,66 @@
...
+      	</test>
+      	<test>
+          	<param name="input" value="dot_blot.tiff" ftype="tiff" />
+          	<param name="input_datatype" value="tiff" />
+          	<output name="output_datatype" value="jpg" /> 
+          	<output name="output" file="dot_blot.jpg" ftype="jpg" />
+      	</test>
+  	</tests>
+    <help>
+**What it does**
```
bgruening: You should use CDATA where ever you can to avoid `&lt;` this.
I would recommend `planemo lint` to you, this can catch such not-best-practice stuff.

gregvonkuster:
Bjorn, thanks for the feedback!

tools/image_processing/imagej2/imagej2_macros.xml
```diff
@@ -0,0 +1,55 @@
+<?xml version='1.0' encoding='UTF-8'?>
+<macros>
+    <xml name="requirements">
+        <requirements>
+            <requirement type="package" version="3.10.2">atlas</requirement>
```
bgruening: This does not need to be included. It will be automatically sourced with numpy.

tools/image_processing/imagej2/imagej2_macros.xml
```diff
@@ -0,0 +1,55 @@
...
+            </when>
+        </conditional>
+    </xml>
+	<token name="@HELP_CITATIONS@">
+
+------
+
+**Citation**
+
+If you use this tool in Galaxy, please cite Blankenberg D, et al. *In preparation.*
```
bgruening: Copy & paste?

tools/image_processing/imagej2/imagej2_convert_format/tool_dependencies.xml
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0"?>
+<tool_dependency>
+    <package name="atlas" version="3.10.2">
+        <repository name="package_atlas_3_10" owner="iuc" />
```
bgruening: I don't think we need this here explicitly. Numpy will install it for us.

tools/image_processing/imagej2/imagej2_macros.xml
gregvonkuster: Yes, I actually do not know the latest way to cite Galaxy in tools, and I saw this in many of yours.  Is there a more appropriate citation?

bgruening: I don't think you need to cite Galaxy in your tools. You can cite yourself, or a github repository if you like.

bgruening: Please, let me know where you found this citation we need to remove this.

gregvonkuster: samtools_sort.xml is the only one in galaxytools, but it is in gatk2_macros.xml in tools-iuc and in many tools in tools-devteam.

bgruening:
Thanks Greg, looks good now. From now on planemo will test all tools with a `.shed.yml` file.

gregvonkuster:
Awesome - thanks!

PR #116, opened 2015-05-07
Aragorn

erasche:
Apparently there's a `-w` option which I wasn't previously aware of, which produces a useful table, meaning that we can get rid of the bloody mess that was the original aragorn to gff3 parser.

@bgruening, @sjackman any comments?


erasche:
Rebased + forced push because my copy of master was out of date.

bgruening:
Thanks @erasche. 
I hope we have at some point proper gff handling as core part of biopython to makes this prettier. For the genome annotation pipeline I will update all wrappers to the latest version.

erasche:
@bgruening agreed, would love for BCBio's gff parser to be part of core biopython.

PR #118, opened 2015-05-08
aragorn_out_to_gff3.py: Make the GFF ID unique

sjackman:
The ID output by ARAGORN in the first column starts over at 1 with each
sequence-region, and so is not unique.

I've qualified each ID with the sequence it came from. The other way to fix this would be to simply assign each tRNA a sequential number.


erasche:
LGTM.

sjackman:
Thanks again, Eric.

erasche:
Happy to help, glad someone is getting some use out of it! And thanks for letting me know about the `-w` option, that's a heck of a lot simpler than parsing the entire, barely-structured report.

PR #121, opened 2015-05-13
Update OpenSwathWorkflow.xml

jj-umn:
options -out_features and -out_tsv  are mutually exclusive, so have the tool wrapper enforce this.


bgruening:
@jj-umn this looks really good! Please keep in mind that all these wrappers are generated automatically, so we need to somehow keep track of these manually changes.

bgruening:
I updated the repository: https://testtoolshed.g2.bx.psu.edu/view/galaxyp/openms/

PR #122, opened 2015-05-14
Update OpenSwathWorkflow.xml

jj-umn:
Create symbolic links to input datasets.  OpenSwath uses the filename extension to determine whether the input is in mzML or mzXML format.

### DIFF
--- a/tools/openms/OpenSwathWorkflow.xml
+++ b/tools/openms/OpenSwathWorkflow.xml
@@ -7,12 +7,22 @@
   </macros>
   <expand macro="stdio"/>
   <expand macro="requirements"/>
-  <command>OpenSwathWorkflow
-
--in
-  #for token in $param_in:
-    $token
-  #end for
+  <command>
+#import re
+#set $input_files = []
+#for infile in $param_in:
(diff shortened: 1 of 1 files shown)


tools/openms/OpenSwathWorkflow.xml
```diff
@@ -7,12 +7,22 @@
...
+  <command>
+#import re
+#set $input_files = []
+#for infile in $param_in:
+  #if str($infile.ext) == 'mzxml':
+    #set $file_extension = '.mzXML'
+  #else:
+    #set $file_extension = '.mzML'
+  #end if
+  #set $input_name = '"' + re.sub('\W','_',$infile.display_name) + $file_extension + '"'
```
bgruening: Do we really need the display name here? I'm a little bit worried about special characters ...

bgruening: @jj-umn any comment on this? Do we need the display name here?

bgruening:
We integrated these changes in https://github.com/bgruening/galaxytools/pull/182.

Thanks.

PR #123, opened 2015-05-20
rna_tools: Add DoRiNA data source tool definition.

rekado:
Add a basic data source tool definition for the public DoRiNA
deployment.  When accessed with a GALAXY_URL parametre, DoRiNA offers a
link back to the provided Galaxy instance enabling it to fetch the
results in tabular form.

### DIFF
--- a/tools/rna_tools/dorina/LICENSE
+++ b/tools/rna_tools/dorina/LICENSE
@@ -1,0 +1,28 @@
+License note for "dorina.xml"
+-----------------------------
+
+Copyright © 2015 Ricardo Wurmus <ricardo.wurmus@mdc-berlin.de>
+This file is part of Webdorina.
+
+Webdorina is free software; you can redistribute it and/or modify it
+under the terms of the GNU General Public License as published by the
+Free Software Foundation; either version 3 of the License, or (at your
+option) any later version.
+
+Webdorina is distributed in the hope that it will be useful, but
+WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+General Public License for more details.
+
+You should have received a copy of the GNU General Public License
+along with Webdorina.  If not, see <http://www.gnu.org/licenses/>.
+
+To enable combination with Galaxy, dorina.xml is also released under
+the Academic Free License version 3.0.
+
+
+License note for "data_source.py"
+---------------------------------
+
+As part of Galaxy, data_source.py is released under the Academic Free
+License version 3.0.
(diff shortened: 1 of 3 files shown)


bgruening:
@rekado thanks a lot! 
Could you please add `data_source.py` to this repository. I think this was one error we encountered during testing.

rekado:
Isn't this part of any Galaxy deployment?  I did not write the code in `data_source.py` and it appears to be used by other data sources as well.

jmchilton:
@bgruening is correct - it needs to live beside the tool. If you look at the tools in Galaxy they are all in folders that also contain ``data_source.py`` files (I think there is just one such folder actually).

rekado:
Okay then.  I'll add `data_source.py` tomorrow then.  (I wonder why it works in our Galaxy installation, but I can't investigate right now.)

TorHou:
@rekado thank you a lot :+1: I'm very glad you did this so quickly. I hope the de.NBI members will take notice :smirk: 
If you install the tool from the toolshed it will put it in a different shed-tools directory.. with no `data_source.py` present. It shouldn't work if you delete dorina from you tools/data_source directory and just use the toolshed version.

rekado:
I added `data_source.py`, though I'm not sure if that's sufficient (I didn't have the time to set up the dorina data source as it would be when installing from the tool shed).

I could not find any license headers on `data_source.py` and I don't know where upstream for this file would be.  Before merging I would like to make sure that `data_source.py` is in fact available under a free software license.

(Do you want me to add license headers to `dorina.xml`?  As part of webdorina it is released under GPLv3+.)

blankenberg:
@rekakdo the file located at tools/data_source/data_source.py in the Galaxy distribution is Academic Free License version 3.0: https://raw.githubusercontent.com/galaxyproject/galaxy/dev/LICENSE.txt

Although, ideally, we should probably build a separate dependency package containing this python script and attempt to decouple it from the Galaxy code as much as possible. I wouldn't consider that a blocker at this point, however.

blankenberg:
Also, I do not think you need to add that license to the xml, the xml doesn't contain any part of the dorina code.

rekado:
Well, as the author of dorina.xml I can publish it under dual licenses to make it possible to combine it with the AFL3.0 in the context of Galaxy and with my preferred license, the GPLv3+, so this is not an insurmountable problem :)

bgruening:
@blankenberg is it ok for you to put  `data_source.py` into a IUC repository - `package_galaxy_data_source_0_1`?
@rekado do you want to add anything? Otherwise I will merge this and add a tool dependency as soon as I have created the IUC repository for `data_source.py`.

rekado:
I have nothing more to add.  The last commit adds license notes for both files in a separate LICENSE file, indicating that for the XML file a license exception to combine with Galaxy is granted.  (AFL and GPL are not compatible, unfortunately, so I had to make this dual-licensing explicit.)

bgruening:
Great! Thanks a lot!

PR #131, opened 2015-06-07
add protein property calculation tool

bgruening:


tools/protein_properties/protein_properties.py
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/env python
+
+import sys
+from Bio import SeqIO
+from Bio.SeqUtils.ProtParam import ProteinAnalysis
+
+with open(sys.argv[2], 'w+') as result:
```
erasche: Just a thought, I've been writing a lot of tools lately and I've been printing, or writing to `sys.stdout` in lieu of opening extra files. Having more pipeable tools is nice.

tools/protein_properties/test-data/protein_properties_result1.bed
```diff
@@ -0,0 +1,3 @@
+ID	Sequence	MW	IP	gravy	length	instability	monoisotpoic
+sp|Q9LHB9|PER32_ARATH	MNFSYSSLSTWTTLMTLGCLLLHSSISSAQLTPTFYDNTCPSVFTIVRDTIVNELRSDPRIAASILRLHFHDCFVNGCDASILLDNTTSFRTEKDAAPNANSARGFPVIDRMKAAVETACPRTVSCADILTIAAQQAVNLAGGPSWRVPLGRRDSLQAFFALANTNLPAPFFTLPQLKASFQNVGLDRPSDLVALSGGHTFGKNQCQFIMDRLYNFSNTGLPDPTLNTTYLQTLRGQCPRNGNQTVLVDFDLRTPTVFDNKYYVNLKELKGLIQTDQELFSSPNATDTIPLVREYADGTQKFFNAFVEAMNRMGNITPLTGTQGQIRQNCRVVNSNSLLHDVVEIVDFVSSM	38846.6361	6.20635986328	-0.0943181818182	352	31.0301420455	0.0880681818182
```
erasche: is it useful to have the sequence in this? Makes the output harder to read for long protiens...

erasche: Oh, just saw that it's `.bed`, nevermind.

bgruening: Yeah thought about this too, but then it's very easy to get the FASTA file back (for example after filtering) with tabular-to-faster.
On the other hand, the user could extract al sequences from the original FASTA with the filtered ID's.
To make it short I'm not sure :) Moving it as last column? Or removing it?

erasche: Last column is an excellent compromise :)

søn. 7. jun. 2015 kl. 16.17 skrev Björn Grüning <notifications@github.com>:

tools/protein_properties/protein_properties.py
bgruening: Ok, this is a good argument. `sys.argv`will still be needed, or I also read from stdin.

erasche: Reading from stdin is a great option if you aren't using `multiple=True`. :)

erasche: http://biopython.org/wiki/Reading_from_unix_pipes the simplicity is nice

bgruening: Yes, biopython is really flexible here.

bgruening:
@erasche changed, thanks!

bgruening:
@erxleben maybe interesting for you as well.

PR #134, opened 2015-06-08
Added the imagej2_create_image tool that provides a model for analyzing

gregvonkuster:
images using version 20141125 of the fiji package in headless mode from
the command line.  This tool creates a new image of a user-specified
type, format, width, height and depth (i.e. stack size).

This pull request also includes a few changes to the imagej2_convert_format tool that are needed due to some of the changes to the utility methods that are shared between the two tools.


bgruening:
Wonderful!

PR #154, opened 2015-06-16
initial commit

smithcr:
work in progress - need to fill the custom parameter select


smithcr:
@nsoranzo Hi, I added a tool dependency and took a different approach in the wrapper i.e. utilizing the "flavours"/aliases that mafft provides. Thanks for pointing out https://bitbucket.org/crs4/orione-tools/src/1003fff61befb17d3c8153c1fbef2121db78f6ab/custom/mafft/mafft.xml?at=bb-orione-tools. I'm keen at some point to fill the custom parameter select, any suggestions going forward?

Cheers,

nsoranzo:
Hi Cameron,
nice work! Looks already more advanced than CRS4 one. Only thing missing from the old one is `--thread \${GALAXY_SLOTS:-1}`.

smithcr:
ah, right thanks!

smithcr:
@bgruening merge?

tools/mafft/tool_dependencies.xml
```diff
@@ -0,0 +1,21 @@
...
+            <actions>
+                <action type="download_by_url">https://raw.githubusercontent.com/bgruening/download_store/master/MAFFT/mafft-7.221-with-extensions-src.tgz</action>
+                <action type="change_directory">./core</action>
+                <action type="shell_command">sed -i "s|\(PREFIX = \).*$|\1$INSTALL_DIR|" Makefile</action>
+                <action type="shell_command">make clean; make; make install</action>
+                <action type="change_directory">../extensions</action>
+                <action type="shell_command">sed -i "s|\(PREFIX = \).*$|\1$INSTALL_DIR|" Makefile</action>
+                <action type="shell_command">make clean; make; make install</action>
+                <action type="set_environment">
+                    <environment_variable name="PATH" action="prepend_to">$INSTALL_DIR/bin</environment_variable>
```
bgruening: Can you please add MAFFT_ROOT_PATH here?

tools/mafft/mafft.xml
```diff
@@ -0,0 +1,152 @@
...
+        ## specify threads to use
+        --thread \${GALAXY_SLOTS:-1}
+        
+        #if $outputFormat.value == 'clustalw'
+            --clustalout
+        #end if
+        
+        $inputSequences > 
+        
+        #if $outputFormat.value == 'fasta'
```
bgruening: @smithcr you could also do the following `> output` and later in the output section use `from_work_dir="output"`. This way you move the filter logic entirely into the output section.

bgruening: Note this will use an additional copy operation. It's your decision if you want to go for cleaner code or for better performance :)

smithcr: okidoke: I will leave it as is for performance and compatibility with autogenerated wrappers...

smithcr:
@bgruening should be good to merge (after travis is finished?)

PR #156, opened 2015-06-17
Add the imagej2_noise tool and wrapper that runs ImageJ in headless mode

gregvonkuster:
to add or remove specified nosie from an image via a Jython script.


tools/image_processing/imagej2/imagej2_noise/imagej2_noise.py
```diff
@@ -0,0 +1,100 @@
+# Galaxy wrapper for use with Imagej2 by Greg Von Kuster
+"""
```
bgruening: @gregvonkuster I'm wondering if it wouldn't be easier to put this all into cheetah. You would save a lot of boilerplate generating the argparse.

tools/image_processing/imagej2/imagej2_noise/imagej2_noise_macros.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version='1.0' encoding='UTF-8'?>
+<macros>
+    <xml name="insertion_select">
```
bgruening: You can also define macros inside of your tool.xml. This comes very handy in such situations.

bgruening:
@gregvonkuster this looks good to me. Please decide yourself if you want to use more cheetah and less python files. This makes probably only sense if it stays readable. But in a few cases it's not worth the overhead of argparse, imho.

xref: https://github.com/galaxyproject/tools-devteam/pull/103

tools/image_processing/imagej2/imagej2_noise/imagej2_noise.py
gregvonkuster: @bgruening I could give it a shot.  Do you want me to try to eliminate the imagej2_noise.py file altogether?   I assume I can make all of the same (or similar) calls from lines # 56 - 100 in Cheetah, but I'll need to see.

bgruening: @gregvonkuster this is just a suggestion. Imho, this is always a trade-off between readability and produced overhead. If you can avoid extra python files go for it, as long as the readability does not degrade too much.

bgruening: `#silent some_function()#` is your cheetah friend.

gregvonkuster: @bgruening Just saw your note letting me decide for this one.  How about this approach - I'll have several more tools coming that follow this model and are probably about as complex as this one.  For the next tool, I'll take the Cheetah approach, and see how that goes.  If all works well, I'll retrofit this tool to use Cheetah as well. Will that work, or would you rather not have this tool committed in its current format and then have it (possibly) significantly changed later.

bgruening: This is totally fine with me. I also don't know if this will work out, so I like this idea. Just wanted to point out this other possibility.

PR #158, opened 2015-06-17
Fixes for functional tests for 2 imagej tools - I somehow got the wrong

gregvonkuster:
copy of the tool config in my initial commit.  I assume some tests have
failed somehwere - very sorry!

### DIFF
--- a/tools/image_processing/imagej2/imagej2_convert_format/imagej2_convert_format.xml
+++ b/tools/image_processing/imagej2/imagej2_convert_format/imagej2_convert_format.xml
@@ -39,19 +39,19 @@
             <param name="input" value="dot_blot.jpg" />
             <param name="input_datatype" value="jpg" />
             <param name="output_datatype" value="png" />
-            <output name="output" file="dot_blot.png" />
+            <output name="output" file="dot_blot.png" compare="sim_size" />
         </test>
         <test>
             <param name="input" value="dot_blot.png" />
             <param name="input_datatype" value="png" />
             <param name="output_datatype" value="tiff" />
-            <output name="output" file="dot_blot.tiff" />
+            <output name="output" file="dot_blot.tiff" compare="sim_size" />
         </test>
         <test>
             <param name="input" value="dot_blot.tiff" />
             <param name="input_datatype" value="tiff" />
             <param name="output_datatype" value="jpg" />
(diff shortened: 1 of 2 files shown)


tools/image_processing/imagej2/imagej2_convert_format/imagej2_convert_format.xml
```diff
@@ -36,24 +36,22 @@
     </outputs>
     <tests>
         <test>
-            <param name="input" value="dot_blot.jpg" ftype="jpg" />
-            <param name="max_heap_size_type" value="megabytes" />
-            <param name="max_heap_size" value="256" />
+            <param name="input" value="dot_blot.jpg" />
             <param name="input_datatype" value="jpg" />
             <param name="output_datatype" value="png" />
-            <output name="output" file="dot_blot.png" ftype="png" />
```
bgruening: specifying `ftype` is recommended.

gregvonkuster: @bgruening This is what I thought, but it seems only some tests will pass if image types are used in tags.  If I remember correctly, gif and tiff do not work and I think there are others.  Maybe my configuration is wrong?

bgruening: Have you tried `compare="sim_size"` instead of diff? I'm not aware of any filetype specific bug. But this does not mean anything :)

gregvonkuster: @bgruening The tests pass without requiring the use of "sim_size", although the use of "lines_diff" is required.  I haven't tried replacing it with "sim_size", although I will if you'd like.  The problem is that using "type" in the tag results in errors like this:

Exception: {u'message': {u'type': u'error', u'data': {u'file_type': u"An invalid option was selected for file_type, u'jpg', please verify.", u'files_metadata': [u"An invalid option was selected for file_type, u'jpg', please verify."]}}}

gregvonkuster: @bgruening Ok, I changed test to use sim_size, and tests are passing.  Thanks!

PR #163, opened 2015-06-19
Add the imagej2_bunwarpj_compare_elastic tool.  This change includes

gregvonkuster:
some miscellaneous cleanup of the current set of imagej tools.


tools/image_processing/imagej2/imagej2_bunwarpj_compare_elastic/imagej2_bunwarpj_compare_elastic.py
```diff
@@ -0,0 +1,72 @@
...
+    error_message = imagej2_base_utils.get_stderr_exception( tmp_err, tmp_stderr, tmp_stdout )
+    imagej2_base_utils.stop_err( error_message )
+
+# Example contents of tmp_output_path:
+# Target image : ~/tmpYOkRxE.jpg
+# Source image : ~/tmpVVtDiE.gif
+# Target Transformation file : ~/tmpi5338f.txt
+# Source Transformation file : ~/tmplVWe3l.txt
+# Warping index = 14.87777347388348
+results = open( tmp_output_path, 'r' ).readlines()
```
bgruening: @gregvonkuster how large can such a file be? Maybe it is safer to iterate over the file and take the last line?

tools/image_processing/imagej2/imagej2_bunwarpj_compare_elastic/imagej2_bunwarpj_compare_elastic.xml
```diff
@@ -0,0 +1,74 @@
...
+the **Save Transformation** option in the **Align two images with bUnwarpJ** tool.
+
+]]>
+
+    </help>
+    <citations>
+        <citation type="bibtex">
+            @InProceedings(Arganda-Carreras2006,
+                author =     "Ignacio Arganda-Carreras and
+                                    Carlos Oscar S{\'a}nchez Sorzano and
```
bgruening: @gregvonkuster this can also move to the macro right? I think they are the same for many tools. If a special tool needs an extra citation we can yield it form the macro.

tools/image_processing/imagej2/imagej2_bunwarpj_compare_elastic/imagej2_bunwarpj_compare_elastic.py
gregvonkuster: @bgruening The file will always be about the size of the example contents I listed - just 5 lines, although the paths will be longer, so I felt deadlines() would be safe.  I will definitely change this to iterate if you'd like me to.  Please confirm.  ;)

bgruening: No way, 5 lines. Go with this! :) Sorry for the disturbance.

tools/image_processing/imagej2/imagej2_bunwarpj_compare_elastic/imagej2_bunwarpj_compare_elastic.xml
gregvonkuster: @bgruening Yes, it could, although this only applies to the bunwarpj tools.  There are enough of them that this should (and will) go to the macro file.  I'll make that change for all bunwarpj tools in my next commit if that''s ok.

tools/image_processing/imagej2/imagej2_bunwarpj_compare_elastic/imagej2_bunwarpj_compare_elastic.xml
```diff
@@ -0,0 +1,74 @@
...
+<![CDATA[
+    python $__tool_directory__/imagej2_bunwarpj_compare_elastic.py
+    --target_image "$target_image"
+    --target_image_format $target_image.ext
+    --source_image "$source_image"
+    --source_image_format $source_image.ext
+    --target_transformation "$target_transformation"
+    --source_transformation "$source_transformation"
+    --max_heap_size_type $set_max_heap_size.max_heap_size_type
+    --max_heap_size $set_max_heap_size.max_heap_size
```
bgruening: And because macros are cool, you could also put the cheetah corresponding to `max_heap_size_type_conditional` into a macro token.

tools/image_processing/imagej2/imagej2_bunwarpj_compare_elastic/imagej2_bunwarpj_compare_elastic.xml
bgruening: Sure!

tools/image_processing/imagej2/imagej2_bunwarpj_compare_elastic/imagej2_bunwarpj_compare_elastic.xml
gregvonkuster: @bgruening I definitely will!  Thanks for the pointers!!

PR #165, opened 2015-06-22
Enhance macro use in the imagej2 tools, fix an intermittent bug in the

gregvonkuster:
imagej2_base_utils.get_max_heap_size_value() function, and fix the id of
the bunwarpj_compare_elastic tool.

@bgruening sorry about screwing up the id of the imagej2_bunwarpj_compare_elastic tool.  :(


bgruening:
@gregvonkuster no worries. Everything is in flux and only in TTS :)

PR #166, opened 2015-06-22
Add the imagej2_bunwarpj_compare_raw tool.

gregvonkuster:


gregvonkuster:
@bgruening Should I wait for acceptance of pull requests before triggering builds in docker, or should I do it when I create the pull request?  Thanks for helping a rookie!  :)

bgruening:
@gregvonkuster new tools needs to be added to https://github.com/bgruening/galaxy-imaging/blob/master/Dockerfile before and it needs to be add the TTS regardless of the PR status ;)

PR #167, opened 2015-06-22
Add the imagej2_bunwarpj_compare_elastic_raw tool and change the

gregvonkuster:
variable names in the imagej2_bunwarpj_compare_raw tool's underlying
code to more closely match that of it's associated tools (i.e., this
tool ans the imagej2_bunwarpj_compare_elastic tool).  The labels remain
the same to align with the bUnwarpJ documentation, which is a bit
confusing since it doesn't closely align with the plugin code.


bgruening:
I will update everything this night :) Thanks!

PR #170, opened 2015-06-23
Add the imagej2_bunwarpj_adapt_transform tool (this completes the

gregvonkuster:
bunwarpj tool suite for the current time), cleaniup imports, fix command
line defaults and comments in several recent imagej tools.


bgruening:
I guess we should link a lot more files. For example the tool_dependencies.xml files or even the test data?

gregvonkuster:
@bgruening I was thinking about this, but I wasn't sure if it may make things messier.  I can give it a try though.  Should we have a tool_dependencies subdirectory at the image root level that contains various names for specific recipes?  then we could link a specific tool's tool_dependencies.xml file to a specified file name in that directory?  Similarly, should we have a test-data directory at the root level, allowing for the same approach for the various tools?

PR #174, opened 2015-07-02
Add new test-data and tool_dependencies directories at the imagej2 root

gregvonkuster:
directory and replace all copies of the same files across imagej2 tools
with symlinks to files in these new directories.


bgruening:
Very nice, this saves a lot of space.

gregvonkuster:
@bgruening I'm still not completely clear on the complete semi-automated process for handling updates to existing repositories like this, so I was hoping these changes would not pose problems.  

There were 2 options from which I could choose: 1: using different directory names and 2: using different file names.  

To clarify, let's look at using soft links to share a tool_dependencies.xml file:

1. Using different directory names:
~/tool_dependencies
    /fiji_20141125/tool_dependencies.xml
    /fiji_future_release/tool_dependencies.xml

Using this approach, tool authors would create symlinks like this:

ln -s ../tool_dependencies/fiji_20141125/tool_dependencies.xml .

Pro: This approach allows the file names to remain the same (i.e., always named tool_dependencies.xml).

Con: Since I had many of the same image and transformation matrix files for tool functional tests named differently, this approach would have resulted in many more changes to tool configs to account for using the same file name rather than the different names I had implemented the tool configs to use (I probably should have standardized these file names, but wasn't paying attention at the time).

2. Using different file names (this is the approach I chose)
~/tool_dependencies
    /fiji_20141125.xml
    /fiji_future_release.xml

Using this approach, tool authors would create symlinks like this:

ln -s ../tool_dependencies/fiji_20141125.xml ./tool_dependencies.xml

Pro: This approach did not force me to make any changes to file names defined in tool configs.

Con: Symlink names are different than the names of the files to which they link.

If you think my choice of the 2nd option will result in any issues, now or in the future, I'll create another pull request that will replace the use of option 2 with option 1.

Sorry for the long discussion here.  Please let me know what you think about this.

Thanks!

PR #175, opened 2015-07-05
very basic fasta splitting function

smithcr:
fast and nasty


TorHou:
you probably don't want the 
;blah > blah
in the command

tools/hacked/splitFasta/splitfasta.xml
```diff
@@ -0,0 +1,32 @@
+<tool id="rbc_splitfasta" name="SplitFasta" version="0.1.0">
+    <requirements>
+    </requirements>
+    <stdio>
+        <exit_code range="1:" />
+    </stdio>
+
+    <command interpreter="python">
+    <![CDATA[
+        splitFasta.py $inputFile; echo 'blah' > blah
```
bgruening: cleanup?

tools/hacked/splitFasta/splitfasta.xml
```diff
@@ -0,0 +1,32 @@
...
+    ]]></command>
+    <inputs>
+        <param name="inputFile" type="data" format="txt" label="Fasta file to split"/>
```
bgruening: format?

tools/hacked/splitFasta/splitfasta.xml
```diff
@@ -0,0 +1,32 @@
...
+            <discover_datasets pattern="(?P&lt;designation&gt;.*)" directory="splits" ext="fasta" visible="false"/>
+        </collection>
+        <data name="output">
+            <filter>outputFormat == 'separate'</filter>
+            <discover_datasets pattern="(?P&lt;designation&gt;.*)" directory="splits" ext="fasta" visible="true"/>
+        </data>
+    </outputs>
+    <help><![CDATA[
+        Takes an input file and writes each consecutive two lines to a separate file, in a dataset collection.
+    ]]></help>
```
bgruening: Reference to: https://gist.github.com/bgruening/ec1e9d2031c83cfe8fa5 ?

smithcr: cool

bgruening:
@smithcr can you fix this?
@jmchilton this tool would be obsolete if we can offer splitting and merging functions to every datatype, derived from the datatype definition `splti()` and `merge()` function.

bgruening:
@smithcr any updates here?

smithcr:
@bgruening not as yet, I'll look at it today

bgruening:
I rebased and fixed the comments. Still not convinced that the Python code is good enough and that this is the correct approach. If anyone wants such a tool we could reuse the code from the split() function included in the datatype FASTA definion in Galaxy.
Let me know.

tools/hacked/splitFasta/splitFasta.py
```diff
@@ -0,0 +1,14 @@
...
+    file_contents = open(inpath, 'r').readlines()
+    os.makedirs('splits')
+    inname = os.path.basename(inpath)
+    for i in range(0, len(file_contents), 2):
+        headline = file_contents[i]
+        outname = headline[1:headline.index(' ')]+'.fa'
+        outfile = open(os.path.join('splits',outname), 'w')
+        outfile.write(file_contents[i])
+        outfile.write(file_contents[i+1])
+        outfile.close()
```
erasche: The script breaks on non-perfect input... I had a user who ran this and now has >2k items in their history because of fixed width fasta format.

So, I rewrote this a little, hope it is useful.

```python
50.24.16.65 - esr [16/Oct/2015:14:19:06 -0500] "GET /galaxy/api/histories/e94e536920e6936f/contents HTTP/1.1" 200 - "https://cpt.tamu.edu/galaxy/root" "Mozillimport sys                                                                                                                                                    
from Bio import SeqIO                                                                                                                                         
                                                                                                                                                              
if __name__ == "__main__":                                                                                                                                    
    inpath = sys.argv[1]                                                                                                                                      
                                                                                                                                                              
    with open(inpath, 'r') as handle:                                                                                                                         
        for record in SeqIO.parse(handle, 'fasta'):                                                                                                           
            header = record.id + '.fa'                                                                                                                        
            with open(header, 'w') as handle2:                                                                                                                
                SeqIO.write([record], handle2, 'fasta')                                                                                                       
```

bgruening:
And open again. @erasche told me this was uploaded to the MTS. Not good!
I will work on this and maintain it.

erasche:
Trivial to fix if you can depend on biopython like I did in my comment.

tools/hacked/splitFasta/splitFasta.py
bgruening: @erasche what do you think about this code:
https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/datatypes/sequence.py#L424

Without the biopython dependency ...

bgruening: But you are right, lets take biopython!

erasche: That code would work just fine, agreed.

bgruening:
@erasche ready for review

erasche:
@bgruening LGTM. I would raise issue with removal of individual files, but I think that's OK since someone said galaxy would eventually support selecting individual files from within dataset collections, so this should be fine.

bgruening:
This option seems to me to dangerous :) and I think we should push for a better UI to extract datasets from collections as you pointed out.

erasche:
agreed, dangerous. xref https://github.com/galaxyproject/galaxy/issues/740

PR #176, opened 2015-07-06
A better implementation of 7b90b1e - all symlink names now remain the

gregvonkuster:
same as the linked file names.  The previous implementation broke manual
single file uploads to TS repositories (its unclear whether planemo
uploads were adversely affected).


bgruening:
Thanks Greg, thanks and greetings from the GCC!

PR #177, opened 2015-07-06
Add the imagej2_skeletonize3d tool.

gregvonkuster:
Say hello to everyone at the GCC for me!


gregvonkuster:
@bgruening I spent some time attempting to eliminate the need for the underlying Python scripts in these tools and using a combination of only a Cheetah template and a Jython script.  I was not able to do that even for simpler tools like this one.  Cheetah is still not very good at importing local Python files which I do a lot using the various utility files.  I also bumped into other difficulties with the more restricted Cheetah syntax, so I felt it best to keep this pattern for these tools where I use underlying Python scripts to call the Jython scripts.  I think maintaining the tools will also be easier.

bgruening:
This is ok for me! What ever is more maintainable! Thanks!

PR #178, opened 2015-07-08
RNA visualization and tools use dbn & ct from core

yhoogstrate:
The data type definitions for DotBracket and ConnectivityTable are now in galaxy.
This updates the visualization and tools to use those.


bgruening:
Great! Thanks! How do you do this during the conference :)

PR #198, opened 2015-08-01
Add sparse - a multiple sequence aligner for RNA structures

bgruening:
This tool was developed by Milad during our GalaxyDevDay. This is still WIP. The dependency needs to be tested and the wrapper needs to be adapted to our guidelines, e.g. the boolean switch for conditionals need to die and the documentation needs some love.

Move the package into it's own repository. Locarna will later depend on it.


bgruening:
@s-will any interest to work on this with some locarna love :)

mmiladi:
@bgruening @s-will : Do we have exparna-p and sparse in the new locarna wrapper?

s-will:
Hi @mmiladi. There is already a sparse wrapper contained in my new locarna wrappers, actually there are two: one in the pairwise and one in the multiple alignment wrapper (however still without the sparse-specific cut-offs, which I postponed -- could you add them??). 
The exparna-p wrapper is still in planning, but I am going to add this together with the other (pairwise) wrappers for the locarna package -- note that it makes little sense to add this as a mode of the pairwise aligner (locarna, sparse), so I plant to write a separate wrapper for exparna-p (and also locarna-p).

mmiladi:
@s-will  I see. Are these the two files the ones that I should change to add sparse specific params?

https://github.com/bgruening/galaxytools/blob/master/tools/rna_tools/locarna/locarna_multiple.xml
https://github.com/bgruening/galaxytools/blob/master/tools/rna_tools/locarna/locarna_pairwise.xml

PS: I am back into the locarna train :-D update you this week about aligner project.

s-will:
Hi Milad, yes, in principle. Moreover, there is a macro file, which you
should use -if possible- to avoid redundancy:
https://github.com/bgruening/galaxytools/blob/master/tools/rna_tools/locarna/macros.xml

mmiladi:
For the updated xml files, what tests should be checked? planemo lint?

s-will:
You should check (in this order)
  planemo lint
  planemo test
  planemo serve
before pushing. Travis does also check the first two (lint+test).

s-will:
of course, planemo test is most useful if you add a /specific/ test case as well

mmiladi:
Features of this PR will be covered by the new PR: https://github.com/bgruening/galaxytools/pull/495

bernt-matthias:
Seems that this PR can be closed?

PR #202, opened 2015-08-03
Update SortMeRNA to v. 2.0 .

nsoranzo:
Planemo tests pass, but needs more testing on cached databases and the output filters.

There are still a number of options not included, but porting to v. 2.0 was my top priority.

This is using the test data kindly provided by @ekopylova.


bgruening:
Awesome work @nsoranzo!
Thanks, also to @ekopylova for providing the test data!

bgruening:
@nsoranzo I uploaded it to the TS and will rebuild the RNA workbench. If everything goes fine I'm happy to upload it to the TS tomorrow if this ok with you.

nsoranzo:
Thanks @bgruening for the merge and for pushing to the test TS!

One thing to consider before pushing to the main TS is the format of the tool data table `rRNA_databases.loc`: previously `buildtrie` was creating its index files in the `automata` directory, instead now `indexdb_rna` asks to specify the index basename as a path. So we should:
- either add a fourth column to the table for the basename path
- or clarify how to run `indexdb_rna`, e.g. `indexdb_rna --ref /path/to/db.fasta,/path/to/db` .

What do you think?

tools/rna_tools/sortmerna/sortmerna.xml
```diff
@@ -18,145 +18,169 @@
...
+            <when value="cached">
+                <param name="input_databases" label="rRNA database" type="select" display="checkboxes" multiple="true">
+                    <options from_data_table="rRNA_databases" />
+                    <validator type="no_options" message="Select at least one database"/>
+                </param>
+            </when>
+            <when value="history">
+                <repeat name="input_databases" title="Database" min="1">
+                    <param name="database_name" type="data" format="fasta" label="rRNA database"
+                        help="Your database will be indexed first, which may take up to several minutes."/>
```
bgruening: Just, saw this now. Is there any reason not to use `multiple=True` here?

bgruening:
This new database build seems to be very confusing in the new version: https://github.com/biocore/sortmerna/issues/51
If possible I would like to avoid changing the .loc files. And as far as I saw you already used your second approach, anything wrong with it?

tools/rna_tools/sortmerna/sortmerna.xml
nsoranzo: Not specifically, it was this way before. Do you have time to update this?

bgruening: Sure will do it.

nsoranzo:
> If possible I would like to avoid changing the .loc files. And as far as I saw you already used your second approach, anything wrong with it?

Yes, I'm using the second approach, the only downside is that it is a bit less flexible (you need to create the index files in the same directory of the FASTA file). But whichever we choose, it needs to be documented in the `rRNA_databases.loc` file.

If you are ok with the second option (i.e. only documentation), I can directly commit an update.

bgruening:
Let's think a little bit ahead. At some point it would be nice to have data managers and here we will have fasta and index in one directory, right? I think this is ok and a save assumption.

ekopylova:
Is creating the index files in the same directory as the FASTA files a constraint in the wrapper, because you can use `indexdb_rna` to index databases in a different directory. The format is,

`indexdb_rna --ref /path1/to/reference1.fasta,/path2/to/reference1.idx`

where path1 does not necessarily need to be path2. Then, the same parameter is passed during the index search,

`sortmerna --ref /path1/to/reference1.fasta,/path2/to/reference1.idx --reads /path/to/reads.fast{a}{q} ..`

nsoranzo:
@ekopylova: That's exactly what I meant in https://github.com/bgruening/galaxytools/pull/202#issuecomment-127566504 , thanks for clarifying!

All the database indexing is completely hidden from the Galaxy users. We are discussing if it is better for the Galaxy admins to have all the common database files (which can be symlinks, BTW) in the same directory (keeping the `rRNA_databases.loc` config file simple) or leave more flexibility adding a fourth column to the `rRNA_databases.loc` file.

ekopylova:
Great! Let me know if there's anything else I can help with :)

PR #203, opened 2015-08-04
adding JAMM Galaxy Wrapper v1.0.7rev1

mahmoudibrahim:
adding Galaxy wrapper was written by Clemens Messerschmidt


bgruening:
@mahmoudibrahim I don't think we want to ship JAMM within the wrapper. We need to write a `tool_dependency.xml` file. This will download a specific version of JAMM and install it.
We want to keep the wrappers and diffs clean. Moreover, JAMM, the dependency, could be used by additional tools as dependency.
Can you please create a different PR with only the wrapper/test-data and so on? Thanks!

bgruening:
@mahmoudibrahim will you work further on this on or can I extract the relevant parts?

PR #208, opened 2015-08-10
Miscellaneous fixes and changes to ImageJ tools:

gregvonkuster:
1) Eliminate the ability to set the JVM memory from tool interfaces.
2) Tools that require binary images as input used to check the image
type against ImagePlus.GRAY8.  This check worked in many cases, but not
all, and the check should have been ImageProcessor.isBinary().  This has
been corrected in all tools that require binary images as input.
3) Fixing the above issue uncovered another bug in the
imagej2_analyze_skeleton tool in that the input images were not always
being converted to binary prior to analysis, so the value for "Slab
Voxels" was not being properly defined.  This bug has also been
corrected, and the tools test data has been updated.


gregvonkuster:
@bgruening I am a bit concerned about the imagej2_analyze_skeleton tool's option to "Show detailed info".  The results are correct, but images that contain more than 1 "skeleton" (i.e., graph) will have details for each graph when this option is set to "yes".  However, the current version of the plugin on which this tool is based does not have the ability to label these skeletons, so it is not possible to know which details are associated with which skeletons in the skeletonize image.  I am thinking it may be best to eliminate this option for the current time so that detailed information is not available until some future time when it is possible to generate an additional output image that displays labels for skeletons in the image.  This would clarify the skeleton IDs in the detailed information.  I believe this information will be available in the next Fiji release, which I understand is coming around September.  Let me know what you think, and if you agree that we should comment out this option for now, I'll do that in another PR. Thanks!

bgruening:
@gregvonkuster I thinks is better to comment this option out and fill a github issue to keep track of it.
Thanks!

gregvonkuster:
@bgruening Done!  Thanks!

PR #211, opened 2015-08-13
Add imagej2_watershed_binary and imagej2_analyze_particles_binary tools

gregvonkuster:
and temporarily eliminate the show_detailed_info option from the imagej2_analyze_skeleton tool.


tools/image_processing/imagej2/imagej2_analyze_particles_binary/imagej2_analyze_particles_binary.xml
```diff
@@ -0,0 +1,134 @@
...
+            <param name="exclude_edges" value="yes" />
+            <param name="include_holes" value="yes" />
+            <output name="output" file="analyze_particles_masks.gif" compare="sim_size" />
+        </test>
+    </tests>
+    <help>
+.. class:: warningmark
+
+This tool works on binary images, so other image types will automatically be converted to binary
+before they are analyzed.  This step is performed using the ImageJ2 **Make Binary** command with
```
bgruening: I saw this text elsewhere as well, so maybe move it to the macro?

gregvonkuster: Yes!  In my next PR.  Thanks for spotting this!

tools/image_processing/imagej2/imagej2_watershed_binary/imagej2_watershed_binary.xml
```diff
@@ -0,0 +1,51 @@
+<?xml version='1.0' encoding='UTF-8'?>
+<tool id="imagej2_watershed_binary" name="Watershed segmentation" version="1.0.0">
+    <description>of binary image with ImageJ2</description>
+    <macros>
+        <import>imagej2_macros.xml</import>
+    </macros>
+    <expand macro="fiji_20141125_requirements" />
```
bgruening: General comment: I think this should be renamed to something without a version number. This would have the advantage that we don't need to change every tool if we increase the dependency.
Similar to the tool-version number. This should also move into a macro similar to: https://github.com/galaxyproject/tools-iuc/blob/master/tools/bedtools/macros.xml#L8

This will make migration to new fiji versions easy :)

gregvonkuster: Ok, I will change it to be just fiji_requirements.  If I remember correctly, I believe I named it this way since the Fiji site states that Fiji is still in beta, and things can change.  But clearly having to change all tools when the next version of Fiji is released is not optimal if changes don't affect tool results.

gregvonkuster: @bgruening Can you clarify your advice here:

expand macro="fiji_20141125_requirements"
General comment: I think this should be renamed to something without a version number. This would have the advantage that we don't need to change every tool if we increase the dependency.

If we use a generic requirements tag, how can we ensure that tools can get access to this specific requirement combination (requirement type="package" version="20141125">fiji) when future fiji versions are released?

Is it safe to assume that we will always be able to update the fiji dependency with the latest fiji release and all tools will produce the same results that they did in the past?  If so, ok - perhaps I'm not thinking clearly on this... ;)

bgruening: I don't think they will produce the same results.
At some point we release this version to MTS. All these tool will depend on version 20141125 and get an installable revision. After a few month we will upgrade the dependency (add the test files ;)) and upload everything. We will get a new installable revision and everything will be good :)

gregvonkuster: Ah, I get it..  Thanks!

PR #212, opened 2015-08-13
General cleanup of ImageJ2 tools:

gregvonkuster:
1) Eliminate unnecessary imports from some Jython scripts
2) Make tool version string a macro
3) Eliminate version string from fiji and bioformats macro requirements
4) Move alert text for tools that require binary image inputs to a macro
5) Eliminate the ability to set output image formats - outputs will have
the same format as inputs
6) Move the adjust threshold command line arguments and parameters from
the macros to the tool config - I originally thought I may use these
across tools, but have since determined that is not needed.


bgruening:
Nice cleanup!

PR #214, opened 2015-08-15
Add the black_background param to all imagej2 binary image tools and fix

gregvonkuster:
setting binary parameters in the tools.  ImageJ2 handles binary
parameters in a different way than macro and plugin parameters.


bgruening:
Looks good to me! Thanks!

PR #221, opened 2015-08-26
Add the imagej2_shadows tool.

gregvonkuster:


tools/image_processing/imagej2/imagej2_shadows/imagej2_shadows.xml
```diff
@@ -0,0 +1,56 @@
+<?xml version='1.0' encoding='UTF-8'?>
+<tool id="imagej2_shadows" name="Add shadow effect" version="@WRAPPER_VERSION@.0">
+    <description>to images with ImageJ2</description>
```
bgruening: Greg, what do think should be mention ImageJ in the description of every tool?

bgruening:
:+1:

tools/image_processing/imagej2/imagej2_shadows/imagej2_shadows.xml
gregvonkuster: @bgruening This is another one of this items I've wondered about.  I put the "with ImageJ2" string in the initial few tools to make more visible what the wrapper is based on, and, of course, it has just snowballed.  I would happily eliminate this string from all tools though if you think it would be better.

bgruening: I guess we will install them into a category, called ImageJ, so this does not need to be part of the description. But I'm also not sure.

gregvonkuster: @bgruening The fact that we've both questioned whether it is necessary is a sign, I think, that it should be removed.  So I'll remove it in my next PR.  ;)

PR #224, opened 2015-08-27
marking duplicates and optionally output split reads and/or discordan…

AGoepfert:
…t pairs


AGoepfert:
Here are the installation steps for tool_dependencies:
Installation

samblaster is self contained and therefore has no installation dependencies beyond g++ and make.

samblaster can be downloaded from the releases tab or manually downloaded via git clone. Afterwards, simply use make and copy samblaster to a directory in your path. For example:

git clone git://github.com/GregoryFaust/samblaster.git
cd samblaster
make
cp samblaster /usr/local/bin/.

bgruening:
I already added this here: https://github.com/bgruening/galaxytools/blob/master/tools/samblaster/tool_dependencies.xml

Thanks!

AGoepfert:
ah - sry!! that is great

PR #226, opened 2015-08-27
Add the imagej2_smooth, imagej2_sharpen and imagej2_find_edges tools,

gregvonkuster:
and eliminate the use of the "with ImageJ2" string in several tool
config description tags.


bgruening:
What a fast turnaround!

PR #228, opened 2015-08-31
Add the imagej2_enhance_contrast tool.

gregvonkuster:


gregvonkuster:
@bgruening I've come up with an approach in the 2nd commit on this PR that eliminates the need for image metadata enhancements in Galaxy for this tool.  I have similar changes to the imagej2_find_maxima tool that I'll submit as a separate PR.  I'll then submit a PR for yet another tool that also uses this approach.  This approach ensures the tools will not have to be changed even if we add enhancements for image metadata into Galaxy at some point.

bgruening:
Ha, great!

PR #230, opened 2015-09-03
Change the imagej2_find_maxima tool to ignore features only used with

gregvonkuster:
RGB images if the input image is not RGB.  This eliminates the use of
the conditional in the original implementation, simplifying the tool
interface.


gregvonkuster:
@bgruening here's the change to the imagej2_find_maxima tool.  ;)

bgruening:
:beer:

PR #231, opened 2015-09-03
Add the imagej2_math_tool.

gregvonkuster:
@bgruening This tool is pretty cool!  ;)  I'll leave you alone after this.  Hope all is going exceptional well for you!


tools/image_processing/imagej2/imagej2_math/jython_script.py
```diff
@@ -0,0 +1,78 @@
...
+    elif operation == 'Max':
+        # Max does not work without using the ImageProcessor.
+        image_processor_copy.max( float_constant )
+    elif operation == 'Abs':
+        if bit_depth not in [ 16, 32 ]:
+            # Convert the image to 32-bit.
+            IJ.run( input_image_plus_copy, "32-bit", "" )
+            IJ.run( input_image_plus_copy, new_operation, "" )
+    elif operation == 'Reciprocal':
+        if bit_depth != 32:
```
bgruening: is the check for 16 bit missing here?

tools/image_processing/imagej2/imagej2_math/jython_script.py
```diff
@@ -0,0 +1,78 @@
...
+        # Min does not work without using the ImageProcessor.
+        image_processor_copy.min( float_constant )
+    elif operation == 'Max':
+        # Max does not work without using the ImageProcessor.
+        image_processor_copy.max( float_constant )
+    elif operation == 'Abs':
+        if bit_depth not in [ 16, 32 ]:
+            # Convert the image to 32-bit.
+            IJ.run( input_image_plus_copy, "32-bit", "" )
+            IJ.run( input_image_plus_copy, new_operation, "" )
```
bgruening: Greg, what happens if this condition is not satisfied, than the operation is not applied?

tools/image_processing/imagej2/imagej2_math/jython_script.py
gregvonkuster: No, this particular option is only applicable to 32-bit, while the other applies to both 16-bit unsigned as well as 32-bit.

tools/image_processing/imagej2/imagej2_math/jython_script.py
gregvonkuster: The ImageJ2 docs make this statement, which I've included in the tool form: This option works only with 32-bit or signed 16-bit images.  When using Fiji, selecting this option for an image that is not 16-bit or 32-bit displays an error message and the user must manually change the image to one of these types.  Using this Galaxy tool the image ia automatically changed to a 32-bit image, which is explained on the tool form.  Changing the image in this way does noting more than changing the bit depth so that it can work for this option.  I think making this automat is a better approach than even Fiji uses.  But with Fiji everything happens in the same session" (i.e., the image is not loaded and saved for each tool run like with Galaxy.  So it is easy for the user to manually change the image type to 32-bit and then execute this option.

bgruening: Got it! Thanks!

bgruening:
Thanks Greg, not sure what you mean you will leave, but I hope you referring to holidays :)

gregvonkuster:
@bgruening I just meant I won't bug you with a pull request for a bit longer now!  I know you are busy, so this is all I'll bug you with at least for the rest of today!!  ;)

PR #234, opened 2015-09-06
Quast

AGoepfert:
Correction of tests and title!

You also mentioned the missing requirements. I didn't write those because I am not sure how.

I installed quast as described on the webpage http://quast.bioinf.spbau.ru/manual.html#sec1 (see the following lines)

1. Installation
QUAST can be run on Linux or Mac OS.

It requires:
python 2 (2.5 or higher)
perl 5.6.0 or higher
g++
make
sh
csh
sed
awk
ar
All those tools are usually preinstalled on Linux.
Mac OS, however, initially misses make, g++ and ar, so you will have to install Xcode (or only Command Line Tools for Xcode) to make them available. 

It is also highly recommended to install the Matplotlib Python library for drawing plots. We recommend to use Matplotlib version 1.0 or higher. Tested with Matplotlib v.1.3.1. 
Installation can be done with Python pip-installer:
    pip install matplotlib
Or with the Easy Install Python module:
    easy_install matplotlib
Or on Ubuntu by typing:
    sudo apt-get install python-matplotlib

To download the QUAST source code tarball and extract it, type:
    wget https://downloads.sourceforge.net/project/quast/quast-3.1.tar.gz
    tar -xzf quast-3.1.tar.gz
    cd quast-3.1
QUAST automatically compiles all its sub-parts when needed (on the first use). Thus, there is no special installation command for QUAST.

_____________

I installed quast and matplotlib - not the other tools like python, perl, g++, etc.

My question is know, do I have to write requirements for those tools as well?


bgruening:
I already added this here: https://github.com/bgruening/galaxytools/blob/master/packages/package_quast_3_1/tool_dependencies.xml but need to wait for a PR to main Galaxy before I can test it intensively.

This PR looks like this was intended so I will close it.

PR #235, opened 2015-09-06
Pileometh

AGoepfert:
Cameron solved my main problem and I made some tiny changes! It should work without any problems! But I am not sure if the tool_dep is working.


tools/pileometh/test-data/cg100.fa.fai
```diff
@@ -0,0 +1,837 @@
+
```
bgruening: This file looks strange and it not a `fai` file. It looks like you tried to download it from github?

AGoepfert:
My mistake! I didn't commit the deleted files - Is this the reason why they were visible for you?

bgruening:
Yes this can be. Can you please rebase against master? Thanks!

AGoepfert:
Everything looks perfect! and unnecessar files are not there anymore!

PR #240, opened 2015-09-10
Changed the output of blockbuster to real bed-format.

TorHou:


TorHou:
I'm not so sure about this anymore. Patrick's follow up depends on the old structure ..

yhoogstrate:
Hi @TorHou,

Can you convert:

```<param name="input" type="data" format="bed" label="BED file containing read expressions"/>```

into something like

```<param name="input" type="data" format="bed" label="BED file containing the read density for every position in the alignment" help="A BED file that aggregates the read density for each position in the alignment. (tip: you should normalize this value by the number of times you can map this tag to the genome)"/>```

Are there tools available to convert BAM files into such BED files?

If so, please mention them, because it's not clear how to get these files.
If not, is it not more convenient to use segemehl's output over BED files?

TorHou:
Hi,

I just talked to Pavan about the tool you are looking for and he suggested:
http://dario.bioinf.uni-leipzig.de/example/map2bed.pl
it is used in this example 
http://dario.bioinf.uni-leipzig.de/help.py#analysis

I wanted to look at blockbuster output (with format 2) directly in IGV or Trackster which this change would make possible.
The problem is: we have workflows which rely on the current format .. so probably one should use a small awk call after blockbuster and get real bed that way. So probably this change is not going in anyway.

yhoogstrate:
Hi @TorHou,

I think I understand what you mean. The output from blockbuster is converted by the wrapper, which for compatibility is not really a desired feature. What looks like a handy solution to me is to add two extra tools to this package:
 - One with map2bed.pl, that will make it very practical to use because input files can be generated from the de facto alignment standard BAM.
 - A small shell script with 'blockbuster2bed' to convert the output of blockbuster directly to BED or BEDGRAPH to be compatible with trackster, igv etc.
 
The only thing you would need to change in the pipeline is to add the blockbuster2bed wrapper at the end.

bgruening:
As we discussed in out pipeline hackathon we will go with this and try to get similar functionality upstream. 
ping @s-will

PR #242, opened 2015-09-13
Upload of all files for uniprot

AGoepfert:
Hey Bjoern,

you find here all files for uniprot!
the tool is running on planemo without any problems - all tests are correct!

Please have a look at:
- requirements (I didn't test them on docker)
- Citation - I was not sure who I should mention (uniprot or the author of the script?) 
- Please run planemo s uniprot.xml with:
-- input: id_uniprot.tab
-- tool: map
-- from: uniprot - ID
-- to: uniprot - gene name
--> I had the problem that I couldn't view the resulting output - the download link showed up by pressing "view data"

Cheers


bgruening:
This looks perfect. The uniprot tools is MIT licensed so I would include it in this repository directly. So no installation is needed.

AGoepfert:
I copied the python scripts directly to the folder and removed the tool requirements

PR #248, opened 2015-09-21
Test case for bigWigToBedGraph that fails on old versions

yhoogstrate:
Hi @bgruening,

I figured out that the v287 bigWigToBedGraph is sensitive to long (medium long?) chromosome names. Sometimes the output is completely empty, contains a few lines or crashes with a segfault. Therefore I think it's important to add a test-case that fails if the old version is being used. I have added a bigWig file, converted from this bedgraph:

```
chr1	210	220	46
chr2	210	220	46
chr3	0	210	47
gacvbdefgrthjinhgacvbdefgrthjinhgacvefgrthjinhga	210	220	46
```

With the following chrom.sizes:

```
chr1	100000
gacvbdefgrthjinhgacvbdefgrthjinhgacvefgrthjinhga	100000
chr2	100000
chr3	100000
```

I have tested it using the v287 bigWigToBedGraph (hard coded in the wrapper) and I got indeed the following error:

image

The error does not appear with the latest binary (http://hgdownload.cse.ucsc.edu/admin/exe/linux.x86_64/bigWigToBedGraph,  01-Sep-2015 23:00, 2db7eb4926769a79167f60b8193505a4)


bgruening:
Great! Thanks! It's time to add a real dependency to it and make sure the correct binary is provided.

yhoogstrate:
Thanks! I think the dependency is already available: https://toolshed.g2.bx.psu.edu/view/iuc/package_ucsc_tools_312
I will test it for the issue stated above asap.

bgruening:
Yes, it is we build them now by our own. Thanks!

PR #252, opened 2015-10-08
Support for basic and extended grep

yhoogstrate:
Hi @bgruening,

I'm used to work with egrep / grep -E so I thought why not including it. I don't know what the additional value to you is, but if you like it you can merge.

I am missing grep in Galaxy. Could you get it into the toolshed or is there still work that to be done apart from the small list in the readme?

thanks in advance,

Youri

### DIFF
--- a/tools/text_processing/text_processing/grep.xml
+++ b/tools/text_processing/text_processing/grep.xml
@@ -13,7 +13,7 @@
             GREP_COLOR='1;34'
             grep
                 --color=always
-                -P
+                $regex_type
                 -A $lines_after
                 -B $lines_before
                 $invert
@@ -22,7 +22,7 @@
                 '${infile}' | $__tool_directory__/ansi2html.sh > "${output}"
         #else:
             grep
-                -P
+                $regex_type
                 -A $lines_after
                 -B $lines_before
                 $invert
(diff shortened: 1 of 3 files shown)


bgruening:
This is fantastic. One question does the -P works for you? I got one user where the compilation failed so that gnu_grep way compiled without perl support. Still investigation. But this PR looks nice!

I need to spend much more time on these tools. The Galaxy team expressed interest in using this on usegalaxy.org.

So many thanks!
Bjoern

yhoogstrate:
Only the second test seems to go wrong using the tool shed grep installation:
```
History item  different than expected, difference (using diff):
( /usr/local/src/galaxytools-bgruening/tools/text_processing/text_processing/test-data/grep_results2.html v. /tmp/tmpLmFOV5/tmp/tmpnlUZAUgrep_results2.html )
--- local_file
+++ history_data
@@ -290,17 +290,11 @@
 <body class="f9 b9">
 <pre>
 GCTATAG<span class="bold"><span class="f4">AAATGT</span></span>TAACATCGAATGTACATTATAAC
-<span class="f6">--</span>
 CAGCTAACAATC<span class="bold"><span class="f4">AAGCGT</span></span>TACAGATTAGCCCCCCCC
-<span class="f6">--</span>
 GAACTTGCGTAACGTACAAAAATGCAAGCA<span class="bold"><span class="f4">AAAAGT</span></span>
-<span class="f6">--</span>
 GCTCTGTTAATCTAGA<span class="bold"><span class="f4">AAATGT</span></span>GTCTCCCCCCCCCC
-<span class="f6">--</span>
 <span class="bold"><span class="f4">AATCGT</span></span>ATAGCTCGGGCCGGATACTAGTACACCCCC
-<span class="f6">--</span>
 GATATAGTGGATAACTAATGCTCCCCCAG<span class="bold"><span class="f4">AACTGT</span></span>T
-<span class="f6">--</span>
 GAACGGACTATAGCCGGTATCCAAACAT<span class="bold"><span class="f4">AAATGT</span></span>TC
 </pre>
 </body>
```
 This test does work with my local version of grep (grep (GNU grep) 2.20)

PR #253, opened 2015-10-09
Have tp_split_on_column use a tmp directory for output. 

blankenberg:
Prevents other cruft in the job directory from being included in the collection list.


bgruening:
Thanks @blankenberg! Do you need an update or can we wait for more fixes?

blankenberg:
I have a local instance running with these modifications, so no need to rush an update for me.

bgruening:
Awesome, let's wait than. Maybe you will find more :)
Let me know if you need an update.

PR #254, opened 2015-10-10
Sailfish update

bgruening:
Latest version of sailfish.


bgruening:
Still not sure if it's worth to write a data manager for it, as the transcriptome changes too frequently and a fixed index does not make so much sense. Currently, the admin can link in manually a reference transcriptome. I think it makes sense to make the history-transcriptome default.
Any opinions? 

ping @nsoranzo @nekrut @jennaj

jennaj:
All makes sense. @blankenberg is a good person to help make the final decision.

nsoranzo:
Never used sailfish, sorry, I have no informed opinion!

bgruening:
Release early release often. Hopefully, this is good enough.

PR #261, opened 2015-10-23
add targetfinder

dyusuf:
Hi Bjoern,

Here is the wrappers for tagetfinder tools.

This is my first galaxy tool so please be accommodating for any silly mistake.

One notification. I found it difficult to test targetfinder_threads. The output of the tool is not constant in terms of order of predicted target sites. This makes difficult to provide a test output for the comparison. In any cases, I put ath_miRNAs_predicted_targets.txt in test-data. For the comparison, this file and the output from galaxy should be both sorted.

cheers,
Dilmurat


bgruening:
@dyusuf this is awesome! 
If you want to learn more about wrapping tools I could review this one carefully and postpone the merge until it's perfect. Or do you want me to merge this and fix stuff afterwards?

dyusuf:
@bgruening Thanks a lot!
I want to learn. Please have a careful review to tell me where the problems are.

tools/rna_tools/targetfinder/TargetFinder.xml
```diff
@@ -0,0 +1,66 @@
...
+            -c $c
+            -t $t
+            -p "$p"
+            $r
+            > $output1
+    ]]></command>
+    <inputs>
+        <param label="Small RNA sequence (RNA or DNA, 5'->3')" help="(-s)" name="s" type="text"/>
+        <param label="Target sequence database file (FASTA-format)" help="(-d)" name="d" type="data" format="fasta" />
+        <param label="Query sequence name (DEFAULT = 'query')" help="(-q)" name="q" type="text" value="query" />
```
bgruening: (DEFAULT = 'query') should go into the help="" text. Help can be really long and educative. label should be short and informative.

tools/rna_tools/targetfinder/TargetFinder.xml
```diff
@@ -0,0 +1,66 @@
...
+            -d $d
+            -q "$q"
+            -c $c
+            -t $t
+            -p "$p"
+            $r
+            > $output1
+    ]]></command>
+    <inputs>
+        <param label="Small RNA sequence (RNA or DNA, 5'->3')" help="(-s)" name="s" type="text"/>
```
bgruening: If your `name` is the same as your `--argument` you can use <param argument="-s">.
name is in this case not needed and the help "(-s)" will be created automatically.

tools/rna_tools/targetfinder/TargetFinder.xml
```diff
@@ -0,0 +1,66 @@
+<tool id="targetfinder" name="targetfinder" version="0.1.0">
```
bgruening: the <description> tag is missing. It would be nice to make this a little bit more informative.
E.g. name="RNA target prediction" <description>tool for plants (targetfinder)</description>

tools/rna_tools/targetfinder/.shed.yml
```diff
@@ -0,0 +1,9 @@
+name: targetfinder
+owner: dyusuf
+description: Plant small RNA target prediction tool
+homepage_url: https://github.com/carringtonlab/TargetFinder.git
+long_description: TargetFinder will computationally predict small RNA binding sites on target transcripts from a sequence database. This is done by aligning the input small RNA sequence against all transcripts, followed by site scoring using a position-weighted scoring matrix.
```
bgruening: You can break long line with the following syntax:

```
long_description: |
  foo
  bar
  
  link
categories: 
```

tools/rna_tools/targetfinder/.shed.yml
```diff
@@ -0,0 +1,9 @@
+name: targetfinder
+owner: dyusuf
```
bgruening: owner: rna_team
This is a shared account between a few peoples that are interested in RNA and Galaxy.

tools/rna_tools/targetfinder/TargetFinder.xml
```diff
@@ -0,0 +1,66 @@
...
+            -p "$p"
+            $r
+            > $output1
+    ]]></command>
+    <inputs>
+        <param label="Small RNA sequence (RNA or DNA, 5'->3')" help="(-s)" name="s" type="text"/>
+        <param label="Target sequence database file (FASTA-format)" help="(-d)" name="d" type="data" format="fasta" />
+        <param label="Query sequence name (DEFAULT = 'query')" help="(-q)" name="q" type="text" value="query" />
+        <param label="Prediction score cutoff value (DEFAULT = 4)" help="(-c)" name="c" type="float" value="4.0" />
+        <param label="Threads for parallel Smith-Waterman searches (DEFAULT = 1)" help="(-t)" name="t" type="integer" value="1" />
```
bgruening: We don't want to offer this option to a user. Please note that this tool will be distributed on a cluster, you normally don't know the exact number of CPU's available. But the scheduler or Galaxy knows. So this should be set from Galaxy: 
See this example:
https://github.com/bgruening/galaxytools/blob/master/tools/diamond/diamond.xml#L19

tools/rna_tools/targetfinder/TargetFinder.xml
```diff
@@ -0,0 +1,66 @@
...
+        <param label="Output format for small RNA-target pairs." help="(-p)" name="p" type="select" >
+                <option value="classic">classic</option>
+                <option value="gff">gff</option>
+                <option value="json">json</option>
```
bgruening: I'm not sure json makes sense in a Galaxy context. We probably don't have a tool that can consume it.
classic would be nice to have a better describtion, that is it, what can a user expect.

tools/rna_tools/targetfinder/TargetFinder.xml
```diff
@@ -0,0 +1,66 @@
...
+                <option value="table">table</option>
```
bgruening: Same for table ... what is it?

tools/rna_tools/targetfinder/TargetFinder.xml
```diff
@@ -0,0 +1,66 @@
...
+        </param>
+        <param label="Search reverse strand for targets? Use this option if the database is genomic DNA." help="(-r)" name="r" type="boolean" falsevalue="" truevalue="-r" checked="false" />
```
bgruening: into the help:  Use this option if the database is genomic DNA.

tools/rna_tools/targetfinder/TargetFinder.xml
```diff
@@ -0,0 +1,66 @@
...
+    </inputs>
+    <outputs>
+        <data label="Predicted targets by targetfinder" name="output1" format="data"/>
```
bgruening: I would remove the label and let Galaxy handle this, but this is only my opinion :)

tools/rna_tools/targetfinder/TargetFinder.xml
```diff
@@ -0,0 +1,66 @@
...
+            <param name="p" value="classic"/>
+            <output name="output1" file="miR399a_predicted_targets.txt"/>
+        </test>
+    </tests>
+    <help><![CDATA[
+**What it does**
+
+    targetfinder: Plant small RNA target prediction tool.
+
+**ATTRIBUTION**
```
bgruening: I don't think this is needed, you have all the references, right?

tools/rna_tools/targetfinder/TargetFinder.xml
```diff
@@ -0,0 +1,66 @@
...
+            <param name="q" value="miR399a"/>
+            <param name="r" value="-r"/>
+            <param name="c" value="4.0"/>
+            <param name="t" value="1"/>
+            <param name="p" value="classic"/>
+            <output name="output1" file="miR399a_predicted_targets.txt"/>
+        </test>
+    </tests>
+    <help><![CDATA[
+**What it does**
```
bgruening: The help should be really improved. More is better ... please note that this is how we can train our users to do the correct thing and help them. As better the help is the less questions they have.

tools/rna_tools/targetfinder/TargetFinder_threads.xml
```diff
@@ -0,0 +1,57 @@
+<tool id="targetfinder_threads" name="targetfinder_threads" version="0.1.0">
```
bgruening: Why do we have 2 tools? There is not real difference from a user point of view right?

tools/rna_tools/targetfinder/tool_dependencies.xml
```diff
@@ -0,0 +1,40 @@
+<?xml version="1.0"?>
+<tool_dependency>
+    <package name="targetfinder" version="1.07">
+        <install version="1.0">
+            <actions>
+                <action type="shell_command">git clone https://github.com/carringtonlab/targetfinder.git</action>
```
bgruening: Download the github release if you can:
https://github.com/carringtonlab/TargetFinder/archive/v1.7.tar.gz

With this we don't have a dependency on git and do not need to clone the entire repository. (This repo is small but it can matter for large ones.)

tools/rna_tools/targetfinder/TargetFinder.xml
dyusuf: that's one of formats provided by targetfinder.

bgruening: :) Just saying that we should help the user here. Describe the output format a little bit, what a user can expect.
Or mention this in the help.

bgruening:
@dyusuf this is a great start!
My only concerns are from a UI point of view. If we provide a usable wrapper with many informations and help this can increase the usability and accessibility a lot!

tools/rna_tools/targetfinder/TargetFinder_threads.xml
dyusuf: That is because of what is provided by targetfinder package. 

the major difference is the input - targetfinder only allows a string of sequence while targetfinder_threads takes a fasta file.

dyusuf:
@bgruening  thanks a lot for the effort.
I will make the mortification accordingly. cheers.

tools/rna_tools/targetfinder/TargetFinder_threads.xml
bgruening: Than lets go with this tool here and abandon the other tool. From a user point of view this makes no difference and everything should be a file in Galaxy to reuse this tool as one piece in a workflow.

dyusuf:
@bgruening  sorry for the delayed response. occupied with a MS. I did modifications following most of your comments. check it out.

tools/rna_tools/targetfinder/TargetFinder_threads.xml
```diff
@@ -0,0 +1,77 @@
...
+        targetfinder_threads.pl
+            -f $f
+            -d $d
+            -c $c
+            -t "\${GALAXY_SLOTS:-12}"
+            $r
+            -o $output1
+    ]]></command>
+    <inputs>
+        <param label="Input small RNA sequences file" help="(-f) FASTA-format" name="f" type="data" format="fasta" />
```
bgruening: Not mandatory, but probably good to know is the `argument=""` attribute. If you use `argument="-c"` you don't need to specify a `name=""` attribute and the `(-f)` in your help attribute is automatically attached.

tools/rna_tools/targetfinder/TargetFinder_threads.xml
```diff
@@ -0,0 +1,77 @@
...
+        <param label="Search reverse strand for targets?" help="(-r) Use this option if the database is genomic DNA." name="r" type="boolean" falsevalue="" truevalue="-r" checked="false" />
+    </inputs>
+    <outputs>
+        <data name="output1" format="data"/>
+    </outputs>
+    <tests>
+        <test>
+            <!-- Order of predicted targets are not constant in output per run with identical parameter. -->
+            <!-- This makes test fail while the wrapper actually works (external validation). -->
+            <param name="f" value="ath_miRNAs_test.fa"/>
```
bgruening: please test with `<output name="output1" file="ath_miRNAs_predicted_targets.txt" compare="contains"/>`.
`ath_miRNAs_predicted_targets.txt` needs to be a subset of the result file.

tools/rna_tools/targetfinder/TargetFinder_threads.xml
```diff
@@ -0,0 +1,77 @@
+<tool id="targetfinder_threads" name="Plant small RNA target prediction tool" version="0.1.0">
```
bgruening: Since you only have now one tool, I would call it `targetfinder` and use `targetfinder_threads.pl`. No one will care about this implementation detail, but will wonder why this tool has such a strange ID or name (here I mean file name) :)

tools/rna_tools/targetfinder/.shed.yml
```diff
@@ -0,0 +1,12 @@
+name: targetfinder
+owner: rna_team
+description: Plant small RNA target prediction tool
+homepage_url: https://github.com/carringtonlab/TargetFinder.git
+long_description: |
+  TargetFinder will computationally predict small RNA binding sites on target transcripts from a sequence database.
+  This is done by aligning the input small RNA sequence against all transcripts,
+  followed by site scoring using a position-weighted scoring matrix.
+remote_repository_url:
```
bgruening: https://github.com/bgruening/galaxytools/tree/master/tools/rna_tools/targetfinder/

tools/rna_tools/targetfinder/tool_dependencies.xml
```diff
@@ -0,0 +1,40 @@
...
+        </install>
+    <readme>
+Required Software Packages: Perl (v5.8+) 
+    </readme>   
+    </package>   
+    <package name="fasta" version="35.4.12">
+        <install version="1.0">
+            <actions>
+                <action type="download_by_url">http://faculty.virginia.edu/wrpearson/fasta/fasta33-35/fasta-35.4.12.tar.gz</action>
+                <action type="shell_command">cd src; make -f ../make/Makefile.linux_sse2 ssearch35_t</action>
```
bgruening: `cd src && make -f ...``

bgruening:
@dyusuf this looks great. I only have small adjustments ... sorry to b so nitpicking!
Please also check our automated linting tests to catch a few errors.

tools/rna_tools/targetfinder/.shed.yml
```diff
@@ -0,0 +1,12 @@
+name: targetfinder
+owner: rna_team
```
bgruening: rnateam

tools/rna_tools/targetfinder/TargetFinder_threads.xml
dyusuf: neat. compare="contains" is what I have been looking for. any books to recommend to improve skills of this kind.

bgruening: A book, maybe not a bad idea :)
For the time being please look at this project:
https://planemo.readthedocs.org/en/latest/writing.html
 and this wiki page
https://wiki.galaxyproject.org/Admin/Tools/ToolConfigSyntax

bgruening: In addition, I learned a lot from PR like this one and reviewing other PR ... to learn new thing. And this one is already quite good! So thanks a lot!

tools/rna_tools/targetfinder/TargetFinder_threads.xml
dyusuf: thanks for the cool tip :) 

yet, I like the current configuration better since it is more transparent to me while passing arguments.

dyusuf:
@bgruening  update is done. cheers. :)

Issue #136, opened 2015-06-08
Lighter


smithcr:
https://github.com/mourisl/Lighter

bgruening:
Start here: https://github.com/bgruening/galaxytools/pull/263

PR #263, opened 2015-10-23
WIP: Initial version of lighter

bgruening:
xref: https://github.com/bgruening/galaxytools/issues/136

Created initial PR, maybe someone ones to pick up this branch.
There is much left to do ...


tools/lighter/lighter.xml
```diff
@@ -0,0 +1,124 @@
+<tool id="lighter" name="lighter" version="1.0">
+    <description>can correct the reads containing sequence errors</description>
+    <command>
+       lighter
+       #for $file in $r:
+           -r $file
+       #end for
+       #if str($kmer.k_select) == "smallk":
+           -k $kmer.k_parameter_string
```
yhoogstrate: Do text type params need to be escaped by quotes? (if so, this also occurs at L11 and L22)

bgruening: Yes they do! Thanks for all your work today! :) Reading through my mails now ...

PR #270, opened 2015-11-25
Fix missing and wrong <when>s.

nsoranzo:
These show up as WARNINGs in the Galaxy log at every server restart.


bgruening:
Thanks @nsoranzo!

Issue #187, opened 2015-07-23
sortmeRNA missing arguments


erxleben:
* [ ]     -L  length of the sliding window (the seed) (default: 18)

nsoranzo:
This is an option of buildtrie, used to create the database indexes.

nsoranzo:
Fixed in #273.

PR #273, opened 2015-11-30
Upgrade the Sortmerna wrapper

### LINKED ISSUE #272: SortMeRNA wrapper
Hello,

I tried SortMeRNA with ToolShed wrapper. But, it does not take into account all parameters I need (seed length for `indexdb_rna` for example). 

I developed a [new wrapper](https://github.com/ASaiM/galaxytools/tree/master/tools/sortmerna), currently available on test Tool Shed.
@bgruening Are you interesting to merge your wrapper with mine?

bebatut:
To fix #272:
* Add parameters that was lacking
* Add specific commands to install on MacOSX

### DIFF
--- a/tools/rna_tools/sortmerna/sortmerna.xml
+++ b/tools/rna_tools/sortmerna/sortmerna.xml
@@ -27,8 +27,6 @@
             #set $ref += $sep + str($db) + ',' + $os.path.splitext($os.path.basename(str($db)))[0]
             #set $sep = ':'
         #end for
-        \${SORTMERNADIR}/indexdb_rna --ref $ref
-        &&
     #else:
         ## databases path is not directly accessible, must match by hand with LOC file contents
         #set $data_table = dict([(_[0], _[2]) for _ in $databases_type.input_databases.input.options.tool_data_table.data])
@@ -37,24 +35,60 @@
             #set $sep = ':'
         #end for
     #end if
-    \${SORTMERNADIR}/sortmerna --ref $ref --reads $input_reads --aligned aligned
-    #if str( $sequencing_type.sequencing_type_selector ) == 'paired'
-        $sequencing_type.paired_type
-    #end if
-    $strand_search
-    $aligned_fastx.aligned_fastx_selector
-    #if $aligned_fastx.aligned_fastx_selector == '--fastx'
-        #if $aligned_fastx.other
-            --other other_file
+    \${SORTMERNADIR}/indexdb_rna --ref $ref -L $seed_length --max_pos $max_pos
+        &&
+    \${SORTMERNADIR}/sortmerna 
+	--ref $ref 
+	--reads $input_reads 
+	--aligned aligned
+	
+	#if str( $sequencing_type.sequencing_type_selector ) == 'paired'
+        	$sequencing_type.paired_type
+	#end if
+
+    	$strand_search
+    	$aligned_fastx.aligned_fastx_selector
+    	#if $aligned_fastx.aligned_fastx_selector == '--fastx'
+        	#if $aligned_fastx.other
+            		--other other_file
+        	#end if
+    	#end if
+    	$aligned_sam.aligned_sam_selector
+    	#if $aligned_sam.aligned_sam_selector == '--sam'
+        	$aligned_sam.sq
+    	#end if
+    	$aligned_blast
+
+	
+    	$log
+	
+	#if $report.report_type == 'best'
+            #if $report.report_best.report_best_type == '0'
+                --best 0
+            #else if $report.report_best.report_best_type == '1'
+                --best 1
+                --min_lis $report.report_best.report_best_min_lis
+            #else
+                --best $report.report_best.report_best_value
+                --min_list $report.report_best.report_best_min_lis
+            #end if
+        #else
+            #if $report.report_num_alignments.report_num_alignments_type == '0'
+                --num_alignments 0
+            #else if $report.report_num_alignments.report_num_alignments_type == '1'
+                --num_alignments 1
+            #else
+                --num_alignments $report.report_num_alignments.report_num_alignments_value
+            #end if
         #end if
-    #end if
-    $aligned_sam.aligned_sam_selector
-    #if $aligned_sam.aligned_sam_selector == '--sam'
-        $aligned_sam.sq
-    #end if
-    $aligned_blast
-    $log
-    -a \${GALAXY_SLOTS:-1}
+
+        -e $e_value
+        --match $match
+        --mismatch $mismatch
+        --gap_open $gap_open
+        --gap_ext $gap_ext
+        -N $ambiguous_letter
+    	-a \${GALAXY_SLOTS:-1}
 ]]>
     </command>
     <inputs>
@@ -102,17 +136,17 @@
 
         <!-- Outputs -->
         <conditional name="aligned_fastx">
-            <param name="aligned_fastx_selector" type="select" label="Include aligned reads in FASTA/FASTQ format">
+            <param name="aligned_fastx_selector" type="select" label="Include aligned reads in FASTA/FASTQ format?">
                 <option value="--fastx">Yes (--fastx)</option>
                 <option value="">No</option>
             </param>
             <when value="--fastx">
-                <param name="other" type="boolean" label="Include rejected reads file" help="(--other)" />
+                <param name="other" type="boolean" label="Include rejected reads file?" help="(--other)" />
             </when>
             <when value="" />
         </conditional>
         <conditional name="aligned_sam">
-            <param name="aligned_sam_selector" type="select" label="Include alignments in SAM format">
+            <param name="aligned_sam_selector" type="select" label="Include alignments in SAM format?">
                 <option value="--sam">Yes (--sam)</option>
                 <option value="">No</option>
             </param>
@@ -131,6 +165,56 @@
         <param name="log" type="boolean" checked="False" truevalue="--log" falsevalue="" label="Generate statistics file"
                help="Generates statistics for the rRNA content of reads, as well as rRNA subunit distribution. (--log)">
         </param>
+	<conditional name="report">
+            <param name="report_type" type="select" display="radio" label="Parameters for filtering and read mapping" help="">
+                <option value="best" selected="true">Report best alignments per read reaching E-value</option>
+                <option value="num_alignments">Report first alignements per read reaching E-value</option>
+            </param>
+            <when value="best">
+                <conditional name="report_best">
+                    <param name="report_best_type" type="select" display="radio" label="Number of searched alignments" help="Only the best alignment is reported">
+                        <option value="0">All high-candidate reference sequences are searched for alignments (very slow)</option>
+                        <option value="1" selected="true">Only one high-candidate reference sequence is searched for alignments (fast). The high-candidate sequences are determined heuristically using a LIS of seed matches)</option>
+                        <option value="other_value">A custom number of reference sequences are searched for alignments (speed decrease for high value)</option>
+                    </param>
+                    <when value="other_value">
+                        <param name="report_best_value" type="integer" min="0" max="100" value="1" label="Number of alignments to be made" help="Only the best one is reported. The computation speed decrease with high value"/>
+                        <param name="report_best_min_lis" type="integer" min="0" max="100" value="2" label="Number of longest LIS an alignement needs to be searched" help="The alignements having the first INT longest LIS. LIS stands for Longest Increasing Subsequence, it is computed using seeds' positions to expand hits into longer matches prior to Smith-Waterman alignment."/>
+                    </when>
+                    <when value="1">
+                        <param name="report_best_min_lis" type="integer" min="0" max="100" value="2" label="Number of longest LIS an alignement needs to be searched" help="The alignements having the first INT longest LIS. LIS stands for Longest Increasing Subsequence, it is computed using seeds' positions to expand hits into longer matches prior to Smith-Waterman alignment."/>
+                    </when>
+                </conditional>
+            </when>
+            <when value="num_alignments">
+                <conditional name="report_num_alignments">
+                    <param name="report_num_alignments_type" type="select" display="radio" label="Number of output alignments" help="">
+                        <option value="0">All alignments reaching the E-value threshold are reported (very slow, this option is not suggested for high similarity rRNA databases)</option>
+                        <option value="1" selected="true">The first alignment passing E-value threshold are reported (very fast, best choice if only filtering is needed)</option>
+                        <option value="other_value">A custom number of alignments are made and reported (speed decrease for high value)</option>
+                    </param>
+                    <when value="other_value">
+                        <param name="report_num_alignments_value" type="integer" min="0" max="100" value="1" label="Number of alignments to be made and reported" help=""/>
+                    </when>
+                </conditional>
+            </when>
+        </conditional>
+
+        <param name="e_value" type="float" min="0" max="10" value="1" label="E-value threshold" help=""/>
+        <param name="match" type="integer" min="0" max="10" value="2" label="SW score for a match" help=""/>
+        <param name="mismatch" type="integer" min="-10" max="0" value="-3" label="SW penalty for a mismatch" help=""/>
+        <param name="gap_open" type="integer" min="0" max="10" value="5" label="SW penalty for introducing a gap" help=""/>
+        <param name="gap_ext" type="integer" min="0" max="10" value="2" label="SW penalty for extending a gap" help=""/>
+        <param name="ambiguous_letter" type="integer" min="-10" max="0" value="-3" label="SW penalty for ambiguous letters (N's)" help=""/>
+
+        <param name="strand" type="select" display="radio" label="Search on" help="">
+            <option value="both" selected="true">Both strands</option>
+            <option value="forward" >Only forward strand</option>
+            <option value="reverse" >Only reverse-complementary strand</option>
+        </param>
+
+        <param name="seed_length" type="integer" min="0" max="100" value="18" label="Seed length for database indexing" help=""/>
+        <param name="max_pos" type="integer" min="0" max="100000" value="10000" label="Maximum number of positions to store for each k-mer for database indexing" help="With 0, all positions are stored"/>
     </inputs>
     <outputs>
         <data format_source="input_reads" name="output_fastx" from_work_dir="aligned.dat"
@@ -248,6 +332,26 @@ The user may also choose to use their own rRNA databases.
 
 Note that your personal databases are indexed each time, and that
 this may take some time depending on the size of the given database.
+
+**Parameters**
+
+The database index can be modulated by:
+
+    - Seed length
+    - Maximum number of positions to store for each k-mer for database indexing
+
+For RNA sorting, the parameters are:
+    
+    - Test to output files in fasta or fastq, in sam and/or in blast format
+    - Test for conservation of rejected sequences
+    - Choice in blast format
+    - Test to add SQ tags in sam file
+    - Filtering and read mapping parameters
+        - Test for conservation of best alignment or first alignment
+        - Number of searched, conserved alignments
+    - E-value threshold
+    - SW score for a match, for a mismatch, for introducing a gap, for extending a gap, for ambigous letters
+    - Strand to search
 ]]>
     </help>
 
--- a/tools/rna_tools/sortmerna/sortmerna.xml
+++ b/tools/rna_tools/sortmerna/sortmerna.xml
@@ -38,13 +38,13 @@ sortmerna --version 2>&1|grep 'SortMeRNA version'
     indexdb_rna --ref $ref -L $seed_length --max_pos $max_pos
     &&
     sortmerna 
-	--ref $ref 
-	--reads $input_reads 
-	--aligned aligned
+    	--ref $ref 
+    	--reads $input_reads 
+    	--aligned aligned
 	
-	#if str( $sequencing_type.sequencing_type_selector ) == 'paired'
-        	$sequencing_type.paired_type
-	#end if
+    	#if str( $sequencing_type.sequencing_type_selector ) == 'paired'
+            	$sequencing_type.paired_type
+    	#end if
 
     	$strand_search
     	$aligned_fastx.aligned_fastx_selector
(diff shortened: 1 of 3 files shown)


tools/rna_tools/sortmerna/tool_dependencies.xml
```diff
@@ -2,17 +2,46 @@
...
+                        </action>
+                        <action type="set_environment">
+                            <environment_variable name="SORTMERNADIR" action="set_to">$INSTALL_DIR/sortmerna-2.0/</environment_variable>
+                            <environment_variable name="PATH" action="prepend_to">$INSTALL_DIR/bin</environment_variable>
+                        </action>
+                    </actions>
+                <actions architecture="x86_64" os="darwin">
+                        <action type="download_by_url" target_filename="sortmerna-2.0.tar.gz">https://github.com/biocore/sortmerna/archive/2.0.tar.gz</action>
+                        <action type="shell_command"><![CDATA[
+                            export CC=clang && export CXX=clang++ && ./configure && make && make install
```
bgruening: @bebatut this uses Clang from LLVM, right? We do not have LLVM as dependency for Galaxy as we are using GCC. Is this working with GCC as well? Are the exports really needed? Shouldn't the system be smart enough to pick the correct compiler?

The IUC needs to decide which compiler we want to use in the future or how we can support both.

tools/rna_tools/sortmerna/sortmerna.xml
```diff
@@ -15,7 +15,7 @@
     </stdio>
```
bgruening: We should increase the tool version number.
In general every change that affects reproducibility should yield in an increase of the tool version number.

tools/rna_tools/sortmerna/sortmerna.xml
```diff
@@ -131,6 +165,56 @@ sortmerna --version 2>&1|grep 'SortMeRNA version'
         <param name="log" type="boolean" checked="False" truevalue="--log" falsevalue="" label="Generate statistics file"
                help="Generates statistics for the rRNA content of reads, as well as rRNA subunit distribution. (--log)">
         </param>
+	<conditional name="report">
+            <param name="report_type" type="select" display="radio" label="Parameters for filtering and read mapping" help="">
```
bgruening: I guess the IUC would recommend a normal select box here:
http://galaxy-iuc-standards.readthedocs.org/en/latest/best_practices/tool_xml.html

I admit it's not clear and I'm fine with a radiobox if this increases usability.

tools/rna_tools/sortmerna/sortmerna.xml
```diff
@@ -131,6 +165,56 @@ sortmerna --version 2>&1|grep 'SortMeRNA version'
...
+                <option value="best" selected="true">Report best alignments per read reaching E-value</option>
+                <option value="num_alignments">Report first alignements per read reaching E-value</option>
+            </param>
+            <when value="best">
+                <conditional name="report_best">
+                    <param name="report_best_type" type="select" display="radio" label="Number of searched alignments" help="Only the best alignment is reported">
```
bgruening: Here I'm slightly more in favour of a select box. I like unified user interfaces and from other tools I would expect a normal drop down box and not radio buttons.

tools/rna_tools/sortmerna/sortmerna.xml
```diff
@@ -131,6 +165,56 @@ sortmerna --version 2>&1|grep 'SortMeRNA version'
...
+                    </param>
+                    <when value="other_value">
+                        <param name="report_best_value" type="integer" min="0" max="100" value="1" label="Number of alignments to be made" help="Only the best one is reported. The computation speed decrease with high value"/>
+                        <param name="report_best_min_lis" type="integer" min="0" max="100" value="2" label="Number of longest LIS an alignement needs to be searched" help="The alignements having the first INT longest LIS. LIS stands for Longest Increasing Subsequence, it is computed using seeds' positions to expand hits into longer matches prior to Smith-Waterman alignment."/>
+                    </when>
+                    <when value="1">
+                        <param name="report_best_min_lis" type="integer" min="0" max="100" value="2" label="Number of longest LIS an alignement needs to be searched" help="The alignements having the first INT longest LIS. LIS stands for Longest Increasing Subsequence, it is computed using seeds' positions to expand hits into longer matches prior to Smith-Waterman alignment."/>
+                    </when>
```
bgruening: <when value="0" /> is missing

tools/rna_tools/sortmerna/sortmerna.xml
```diff
@@ -131,6 +165,56 @@ sortmerna --version 2>&1|grep 'SortMeRNA version'
...
+                <conditional name="report_num_alignments">
+                    <param name="report_num_alignments_type" type="select" display="radio" label="Number of output alignments" help="">
+                        <option value="0">All alignments reaching the E-value threshold are reported (very slow, this option is not suggested for high similarity rRNA databases)</option>
+                        <option value="1" selected="true">The first alignment passing E-value threshold are reported (very fast, best choice if only filtering is needed)</option>
+                        <option value="other_value">A custom number of alignments are made and reported (speed decrease for high value)</option>
+                    </param>
+                    <when value="other_value">
+                        <param name="report_num_alignments_value" type="integer" min="0" max="100" value="1" label="Number of alignments to be made and reported" help=""/>
+                    </when>
+                </conditional>
```
bgruening: <when value="0" /> is missing
<when value="1" /> is missing

tools/rna_tools/sortmerna/sortmerna.xml
```diff
@@ -131,6 +165,56 @@ sortmerna --version 2>&1|grep 'SortMeRNA version'
...
+        </conditional>
+
+        <param name="e_value" type="float" min="0" max="10" value="1" label="E-value threshold" help=""/>
+        <param name="match" type="integer" min="0" max="10" value="2" label="SW score for a match" help=""/>
+        <param name="mismatch" type="integer" min="-10" max="0" value="-3" label="SW penalty for a mismatch" help=""/>
+        <param name="gap_open" type="integer" min="0" max="10" value="5" label="SW penalty for introducing a gap" help=""/>
+        <param name="gap_ext" type="integer" min="0" max="10" value="2" label="SW penalty for extending a gap" help=""/>
+        <param name="ambiguous_letter" type="integer" min="-10" max="0" value="-3" label="SW penalty for ambiguous letters (N's)" help=""/>
+
+        <param name="strand" type="select" display="radio" label="Search on" help="">
```
bgruening: drop down box

tools/rna_tools/sortmerna/sortmerna.xml
```diff
@@ -37,24 +35,60 @@ sortmerna --version 2>&1|grep 'SortMeRNA version'
...
+                --best 0
+            #else if $report.report_best.report_best_type == '1'
+                --best 1
+                --min_lis $report.report_best.report_best_min_lis
+            #else
+                --best $report.report_best.report_best_value
+                --min_list $report.report_best.report_best_min_lis
+            #end if
+        #else
+            #if $report.report_num_alignments.report_num_alignments_type == '0'
```
bgruening:
```
#if $report.report_num_alignments.report_num_alignments_type in ['0','1']:
    --num_alignments $report.report_num_alignments.report_num_alignments_type
#else:
```

or you can revert the if else ...

tools/rna_tools/sortmerna/tool_dependencies.xml
```diff
@@ -2,17 +2,46 @@
...
+            <actions_group>
+                <actions architecture="x86_64" os="linux">
+                        <action type="download_by_url" target_filename="sortmerna-2.0.tar.gz">https://github.com/biocore/sortmerna/archive/2.0.tar.gz</action>
+                        <action type="autoconf"/>
+                        <action type="change_directory">../</action>
+                        <action type="make_directory">$INSTALL_DIR/sortmerna-2.0/</action>
+                        <action type="move_directory_files">
+                            <source_directory>sortmerna-2.0/</source_directory>
+                            <destination_directory>$INSTALL_DIR/sortmerna-2.0/</destination_directory>
+                        </action>
```
nsoranzo: What files/directories do you need? The `autoconf` action already installs the files in `$INSTALL_DIR/bin/` and `$INSTALL_DIR/include/` .
I think that any extra file/dir should be moved explicitly.

tools/rna_tools/sortmerna/sortmerna.xml
```diff
@@ -15,7 +15,7 @@
     </stdio>
     <version_command>
 <![CDATA[
-sortmerna --version 2>&1|grep 'SortMeRNA version'
+\${SORTMERNADIR}/sortmerna --version 2>&1|grep 'SortMeRNA version'
```
nsoranzo: The added `\${SORTMERNADIR}/` should not be needed. Same below.

tools/rna_tools/sortmerna/tool_dependencies.xml
```diff
@@ -2,17 +2,46 @@
...
+                    </actions>
+                <actions architecture="x86_64" os="darwin">
+                        <action type="download_by_url" target_filename="sortmerna-2.0.tar.gz">https://github.com/biocore/sortmerna/archive/2.0.tar.gz</action>
+                        <action type="shell_command"><![CDATA[
+                            export CC=clang && export CXX=clang++ && ./configure && make && make install
+                        ]]>
+                        </action>
+                        <action type="change_directory">../</action>
+                        <action type="make_directory">$INSTALL_DIR/sortmerna-2.0/</action>
+                        <action type="move_directory_files">
```
bgruening: Is there any need to create a separate directory? Why not move the content into $INSTALL_DIR/bin? If you have it in `bin` and `bin` is on your $PATH you don't need to prefix it in cheetah: `\${SORTMERNADIR}/sortmerna`

tools/rna_tools/sortmerna/tool_dependencies.xml
```diff
@@ -2,17 +2,46 @@
...
+                    </actions>
+                </actions_group>
         </install>
         <readme>
-        SortMeRNA requires g++ 4.3 or later. Installation may take a moment since ribosomal databases have to be indexed.
+        SortMeRNA requires g++ 4.3 or later. 
+
+        Note: the Clang compiler on Mac (distributed through Xcode) does not 
+        support multithreading. The user is recommended to install the original 
+        GCC compiler via MacPorts
```
bgruening: Can you use GCC if you exported clang previously?

bgruening:
Thanks @bebatut I have added a few comments inline.

tools/rna_tools/sortmerna/sortmerna.xml
```diff
@@ -131,6 +165,56 @@ sortmerna --version 2>&1|grep 'SortMeRNA version'
...
+                        <param name="report_best_min_lis" type="integer" min="0" max="100" value="2" label="Number of longest LIS an alignement needs to be searched" help="The alignements having the first INT longest LIS. LIS stands for Longest Increasing Subsequence, it is computed using seeds' positions to expand hits into longer matches prior to Smith-Waterman alignment."/>
+                    </when>
+                    <when value="1">
+                        <param name="report_best_min_lis" type="integer" min="0" max="100" value="2" label="Number of longest LIS an alignement needs to be searched" help="The alignements having the first INT longest LIS. LIS stands for Longest Increasing Subsequence, it is computed using seeds' positions to expand hits into longer matches prior to Smith-Waterman alignment."/>
+                    </when>
+                </conditional>
+            </when>
+            <when value="num_alignments">
+                <conditional name="report_num_alignments">
+                    <param name="report_num_alignments_type" type="select" display="radio" label="Number of output alignments" help="">
```
nsoranzo: `help="(--num_alignments)"`

tools/rna_tools/sortmerna/sortmerna.xml
```diff
@@ -37,24 +35,60 @@ sortmerna --version 2>&1|grep 'SortMeRNA version'
...
+	
+	#if $report.report_type == 'best'
+            #if $report.report_best.report_best_type == '0'
+                --best 0
+            #else if $report.report_best.report_best_type == '1'
+                --best 1
+                --min_lis $report.report_best.report_best_min_lis
+            #else
+                --best $report.report_best.report_best_value
+                --min_list $report.report_best.report_best_min_lis
```
nsoranzo: s/--min_list/--min_lis/

tools/rna_tools/sortmerna/sortmerna.xml
nsoranzo: Add `(--best)` at the end of the `help` attribute value.

tools/rna_tools/sortmerna/sortmerna.xml
```diff
@@ -131,6 +165,56 @@ sortmerna --version 2>&1|grep 'SortMeRNA version'
...
+                        <option value="other_value">A custom number of alignments are made and reported (speed decrease for high value)</option>
+                    </param>
+                    <when value="other_value">
+                        <param name="report_num_alignments_value" type="integer" min="0" max="100" value="1" label="Number of alignments to be made and reported" help=""/>
+                    </when>
+                </conditional>
+            </when>
+        </conditional>
+
+        <param name="e_value" type="float" min="0" max="10" value="1" label="E-value threshold" help=""/>
```
nsoranzo: `help="(-e)"`

tools/rna_tools/sortmerna/sortmerna.xml
bebatut: There is no parameters to configure when "0" is checked

tools/rna_tools/sortmerna/sortmerna.xml
bebatut: There is no parameters to configure when 0 or 1 is checked. Do I have to add an empty ``<when value="1" />`` ?

tools/rna_tools/sortmerna/sortmerna.xml
bgruening: Yes, but this is needed. Otherwise Galaxy will generate a lot of warnings.

tools/rna_tools/sortmerna/sortmerna.xml
nsoranzo: Yes, that's preferable, because otherwise Galaxy complains in the server logs.

tools/rna_tools/sortmerna/tool_dependencies.xml
bebatut: The question is about the data in ``rRNA_databases`` directory. Do we need to move it explicitily?

bgruening: Yes this needs to be moved explicitly I guess. How large is this directory.

nsoranzo: It's 84 MB (just the FASTA file, no indexes).

bebatut:
I still have an issue with databases. When I instal from ToolShed, the databases in SortMeRNA rRNA_database directory can not be checked. 
Any suggestion?

bgruening:
@bebatut can it be that the binaries needs to be next to the `rRNA_databases` directory?
I don't have a installation by hand, sorry.

bebatut:
I think I don't understand your point. Can you explain it?

nsoranzo:
@bebatut After installing the tool you need to follow the instructions contained in tool-data/rRNA_databases.loc.sample

bgruening:
Btw. here is an example on how to write tests for *.loc files:
https://github.com/galaxyproject/tools-devteam/pull/297

Not saying that it is a must to integrate. But probably nice to know and to play with.

tools/rna_tools/sortmerna/tool_dependencies.xml
```diff
@@ -2,17 +2,46 @@
...
+                        <action type="download_by_url" target_filename="sortmerna-2.0.tar.gz">https://github.com/biocore/sortmerna/archive/2.0.tar.gz</action>
+                        <action type="autoconf"/>
+                        <action type="change_directory">../</action>
+                        <action type="make_directory">$INSTALL_DIR/sortmerna-2.0/rRNA_databases/</action>
+                        <action type="move_directory_files">
+                            <source_directory>sortmerna-2.0/rRNA_databases/</source_directory>
+                            <destination_directory>$INSTALL_DIR/sortmerna-2.0/rRNA_databases/</destination_directory>
+                        </action>
+                        <action type="set_environment">
+                            <environment_variable name="SORTMERNADIR" action="set_to">$INSTALL_DIR/sortmerna-2.0/</environment_variable>
```
bgruening: This is not needed anymore, right?

tools/rna_tools/sortmerna/tool_dependencies.xml
```diff
@@ -2,17 +2,46 @@
...
-                    <environment_variable name="SORTMERNADIR" action="set_to">$INSTALL_DIR</environment_variable>
-                    <environment_variable name="PATH" action="prepend_to">$INSTALL_DIR/bin</environment_variable>
-                </action>
-            </actions>
+            <actions_group>
+                <actions architecture="x86_64" os="linux">
+                        <action type="download_by_url" target_filename="sortmerna-2.0.tar.gz">https://github.com/biocore/sortmerna/archive/2.0.tar.gz</action>
+                        <action type="autoconf"/>
+                        <action type="change_directory">../</action>
+                        <action type="make_directory">$INSTALL_DIR/sortmerna-2.0/rRNA_databases/</action>
```
nsoranzo: I'd use just `$INSTALL_DIR/rRNA_databases`, but that's just a personal preference.

tools/rna_tools/sortmerna/tool_dependencies.xml
nsoranzo: It's needed because it's referenced in tool-data/rRNA_databases.loc.sample .

bebatut:
@nsoranzo Is it not supposed to be automatically done with installation from the ToolShed?

nsoranzo:
@bebatut No, because the Galaxy admin can decide to install all, some or none of these databases. Also building the indexes takes time and a lot of disk space (1.3 Gb here).

nsoranzo:
This will also fix #187.

nsoranzo:
About the last 2 commits, I fear changing the environment variable from SORTMERNADIR to SORTMERNADBDIR may create problems for people updating from the current version. May be better to revert them?

bebatut:
@nsoranzo Once it is installed using the ToolShed, how the Galaxy admin can decide to install these databases? We have to add the information in the help.
I added database indexation each time sortmerna is called. Some indexation parameters such as seed length were added and I did not know how to manage if an user wants non standard indexation parameters for an already indexed database.

bebatut:
@nsoranzo I can revert the last 2 commits

nsoranzo:
@bebatut 
> Once it is installed using the ToolShed, how the Galaxy admin can decide to install these databases? We have to add the information in the help.

This is documented in the loc.sample file: https://github.com/bgruening/galaxytools/blob/master/tools/rna_tools/sortmerna/tool-data/rRNA_databases.loc.sample#L20 , but we can also add a note in `readme.md`

> I added database indexation each time sortmerna is called. Some indexation parameters such as seed length were added and I did not know how to manage if an user wants non standard indexation parameters for an already indexed database.

I haven't noticed that, sorry. I think that most Galaxy tools using loc files expect the Galaxy admin to create the indexes just once (with default parameters) to save execution time. If a user wants to use different indexing parameters, she usually needs to upload the reference, but we can also give three options:
- Use built-in pre-indexed database
- Index built-in database with different parameters
- Index database from your history

Opionions? @bgruening?

> I can revert the last 2 commits

Yes, please!

bebatut:
In last commit, I had the option of reindexing built-in databases. And I limit call to `indexdb_rna` only to reindexing of built-in databases and indexing a history databases

tools/rna_tools/sortmerna/sortmerna.xml
```diff
@@ -131,6 +187,50 @@ sortmerna --version 2>&1|grep 'SortMeRNA version'
...
+                        <option value="1" selected="true">Only one high-candidate reference sequence is searched for alignments (fast). The high-candidate sequences are determined heuristically using a LIS of seed matches)</option>
+                        <option value="other_value">A custom number of reference sequences are searched for alignments (speed decrease for high value)</option>
+                    </param>
+                    <when value="0" />
+                    <when value="1">
+                        <param name="report_best_min_lis" type="integer" min="0" max="100" value="2" label="Number of longest LIS an alignement needs to be searched" help="The alignements having the first INT longest LIS. LIS stands for Longest Increasing Subsequence, it is computed using seeds' positions to expand hits into longer matches prior to Smith-Waterman alignment. (--min_lis)"/>
+                    </when>
+                    <when value="other_value">
+                        <param name="report_best_value" type="integer" min="0" max="100" value="1" label="Number of alignments to be made" help="Only the best one is reported. The computation speed decrease with high value"/>
```
nsoranzo: Does it make sense to have the special case for `--best 1`? If it needs special documentation, that can be added to the help of `report_best_value`.

tools/rna_tools/sortmerna/sortmerna.xml
```diff
@@ -246,8 +346,9 @@ The user may also choose to use their own rRNA databases.
 
 .. class:: warningmark
 
-Note that your personal databases are indexed each time, and that
+Note that the databases are indexed each time given the choosed indexation parameters, and that
```
nsoranzo: s/indexation/indexing/

Also this should be updated to reflect the latest changes.

nsoranzo:
Thanks @bebatut, the latest changes are top-notch!

bgruening:
@bebatut wow this looks really good now! Thanks a bunch!

tools/rna_tools/sortmerna/sortmerna.xml
bebatut: I think so. 
The range for `--best` is not clear. But `--best 1` is recommended for fast execution, we have to insist on it. Maybe I can juste change the range for `report_best_value` to differentiate the cases (0,1, and other values)

tools/rna_tools/sortmerna/sortmerna.xml
bebatut: Done. @nsoranzo thanks

nsoranzo:
:+1:

Thanks @bebatut!

bgruening:
!!! :+1:

PR #276, opened 2015-12-06
WIP: Biopython and gffDB work

bgruening:
I promised for so long to put this online. Here we go. 
This is our work to streamline gff and NCBI tbl handling.

The idea is to manipulate gff files to make it fit the gffDB format so that we can put them into a unique GFF sqlite database. Additionally we will provide tools to query this database. I will try to work on this over X-mas, I hope.

ping @erasche


erasche:
Wow! Just... wow! So much new code to look over :)

It would be nice to have a list of all the operations we want to be able to do, so I can make sure chado tooling is at feature parity (and then it would be really cool if we could write our XML such that people just select "Talk to Chado" or "GffDB from history")

PR #281, opened 2015-12-15
Upgrade diamond with new diamond package

bebatut:
Hi, 

I changed diamond `tool_dependencies.xml` and version in `diamond.xml` and `diamond_makedb.xml` to use the new `package_diamond_0_7_10`

Bérénice


bgruening:
@bebatut thanks for you help here! I don't think this is enough to merge it. I was afraid of doing it because they changed the output format in version 0.7. Afaik we need to include a manually conversion step now to provide the same output. So we need to change a little bit the parameters! But this is a first great step! Thanks. I hope I can work on this all over Xmas.

bebatut:
Ok :)

bgruening:
@bebatut are you working on this or do you have time to do so?

bebatut:
I did not plan to work on this and I'm not sure I will have time in next days :(

bgruening:
Ok, no problem :) Just want to plan this a little bit my train-tavels.

bebatut:
I started reworking on update of diamond. It is just a beginning... Work in progress

tools/diamond/diamond.xml
```diff
@@ -78,7 +78,7 @@
             <option value="blastx">Align DNA query sequences (blastx)</option>
         </param>
 
-        <param argument="--query" type="data" format="fasta,fastq" label="Input query file in FASTA or FASTQ format" />
+        <param name="query"  argument="--query" type="data" format="fasta,fastq" label="Input query file in FASTA or FASTQ format" />
```
bgruening: In this case you do not need the name attr

tools/diamond/diamond.xml
```diff
@@ -64,83 +94,168 @@
...
-            <option value="BLOSUM80">BLOSUM80</option>
-            <option value="BLOSUM90">BLOSUM90</option>
-            <option value="PAM250">PAM250</option>
-            <option value="PAM70">PAM70</option>
-            <option value="PAM30">PAM30</option>
+        <param name="sensitive" argument="--sensitive" type="boolean" truevalue="--sensitive" falsevalue="" checked="false" label="Trigger the sensitive alignment mode with a 16x9 seed shape configuration?" help=""/>
+
+        <param name="more_sensitive" argument="--more-sensitive" type="boolean" truevalue="--more-sensitive" falsevalue="" checked="false" label="Trigger the more sensitive mode?" help="This mode provides some additional sensitivity compared to the sensitive mode."/>
+
+        <param name="gapopen" targument="--gapopen" type="integer" value="11" label="Gap open penalty" help="" />
```
bgruening: Name can be removed also below

bebatut:
I do not know if a diamond db can be sniffed... There is a `datatypes_conf.xml` here to define the datatype, but that is all.

bgruening:
Ok, let's move this datatype into Galaxy as a subclass one of binary. Do you want me to do this?

bebatut:
I can do this if you explain me how to do...

bgruening:
This one is an example with sniffer: https://github.com/galaxyproject/galaxy/pull/3073

If you can not sniff you just need to include one line into the datatypes_conf.xml file subclassing binary.

bgruening:
Wuhu!!! Thanks a lot!

PR #282, opened 2015-12-17
update metilene wrapper

bgruening:


bgruening:
@s-will thanks a lot for this tool!
@dpryan79 maybe this is interesting for you as well

PR #285, opened 2015-12-22
Added sambamba wrapper

thaze:


thaze:
@bgruening please review.

tools/sambamba/sambamba.xml
```diff
@@ -0,0 +1,222 @@
+<tool id="sambamba" name="sambamba" version="">
+    <description> High performance, modern, robust and fast tool to work with SAM and BAM files</description>
```
bgruening: whitespace in the beginning of the string

tools/sambamba/sambamba.xml
```diff
@@ -0,0 +1,222 @@
+<tool id="sambamba" name="sambamba" version="">
```
bgruening: You need to have a version. Usually the same version as the tool with a `.0` at the end.

bgruening:
Test data is missing :)

tools/sambamba/sambamba.xml
```diff
@@ -0,0 +1,222 @@
...
+               <param name="a" type="boolean" truevalue="-a" falsevalue="" checked="false" label="Add additional column of y/n instead of  skipping records not satisfying the criteria"/>
+               <param name="w" type="integer" value="1" optional="false" label="Breadth of the window, in bp (required)"/>
+               <param name="overlap" type="integer" value="0" optional="false" label="Overlap of successive windows, in bp (default is 0)"/>
+           </when>           
+        </conditional>
+    </inputs>
+    <outputs>
+        <data name="output" format="input" metadata_source="input_file1" />
+          <change_format>
+              <when input="flagstat" value="bam" format="txt" />
```
bgruening: value bam?
input="function_select" value="flagstat" maybe?

jmchilton:
I'd really encourage you consider making this 5-8 tools instead of one. Flagstat is so different from sort for instance that they don't belong in the same Galaxy tool in my opinion.

tools/sambamba/sambamba.xml
```diff
@@ -0,0 +1,226 @@
+<tool id="sambamba" name="sambamba" version="0.5.9.0">
+    <description>Sambamba: Process your BAM data faster!</description>
+    <requirements>
+        <requirement type="package" version="1.0">sambamba</requirement>
```
yhoogstrate: This refers to a `tool_dependencies.xml` which I believe is missing. Could you add it? It should look something like this: https://bitbucket.org/lance_parsons/sambamba_filter_galaxy_wrapper/src/fa4adadc5e8bc444b2278dc4f8ad8dfd1fd1f55f/tool_dependencies.xml?fileviewer=file-view-default

Also I think `1.0` should be `0.5.9` as you did in the first line of this file

davebx:
I'm seconding @jmchilton's recommendation of creating separate tool XML wrappers for the different functions. In addition to his argument, I feel that aesthetics would be improved by having different tools each doing one thing rather than a large and potentially fragile conditional block.

PR #301, opened 2016-02-28
extended travis testing of tools

bgruening:
Enable testing of tools with planemo and conda dependencies


bgruening:
Depends on https://github.com/galaxyproject/galaxy/pull/1816

bgruening:
Ready for merge.

yhoogstrate:
According to the results of Travis it should be fine so I would go for it. The blacklist can be reduced over time. I am wondering if it would be helpful to set `GALAXY_RELEASE=release_16.01` to ` GALAXY_RELEASE=master`, instead of manually updating? On the other hand, it could happen that after a new galaxy release travis will not psdd because of upstream incompatibility.

yhoogstrate:
Oops, it's merged already :)

bgruening:
But thanks for you inputs. Yes this was my thinking as well, sticking to master will yield potential new issues, which is good :)

PR #303, opened 2016-03-01
ViennaRNA: also allow sequences from text input field within RNAFold

yhoogstrate:
I have added the option to directly input a RNA sequence in a text field (and I have set this to default, so please comment on this). I hope you like this feature.


tools/rna_tools/vienna_rna/rnafold.xml
```diff
@@ -17,33 +22,47 @@
...
     > $tabular_file
 ]]>
     </command>
 
     <inputs>
-    <param format="fasta" name="fasta_input" type="data" label="FASTA file"/>
-        <param name="temperature" type="float" value="37.0" label="Temperature [°C]" help="-T"/>
-        <param name="dangling" type="select" label="how to treat dangling end energies" help="-d">
+        <conditional name="input_source">
+            <param name="select_fasta" type="boolean" truevalue="true" falsevalue="false" label="Input from FASTA file" selected="false" />
```
bgruening: Can we have a select type here? I know the old one was a boolean, but if we clean up these wrappers it would be nice to stick to the IUC recommendations.

tools/rna_tools/vienna_rna/rnafold.xml
```diff
@@ -17,33 +22,47 @@
...
+
+            <when value="true">
+                <param format="fasta" name="fasta_input" type="data" label="Sequence to fold (FASTA file)"/>
+            </when>
+            <when value="false">
+                <param name="input_sequence" type="text" label="Input sequence"/>
```
bgruening: Hopefully no one will enter a fasta sequence here. Can we be smarter at the cheetah level and look at the first character? Something like $input.strip()[0] != '>'?

tools/rna_tools/vienna_rna/rnafold.xml
```diff
@@ -54,20 +73,77 @@
...
+                   truevalue="--circ"
+                   falsevalue=""
+                   checked="false"
+                   label="Model as circular RNA structure"
+                   argument="--circ"/>
+        </section>
     </inputs>
     <outputs>
-        <data format="tabular" name="tabular_file"/>
+        <data format="dbn" name="tabular_file"/>
```
bgruening: This is awesome! But I guess we need to adopt the other wrappers as well to consume this filetype.

bgruening:
Any reason why you want this to be the default value? I'm worried about usage in workflows.
I really like this idea but I think as non-default it would make more better. Open for discussions ...

I'm also worried about the other wrappers, I guess we should adopt all at once and make them unique in there behaviour, this also relates with the filetypes. If we produce here a dbn file, this need to be consumable by others as well.

What do you think?
Thanks, your help is very much appreciated.

Btw. have a look at this branch: https://github.com/bgruening/galaxytools/pull/301 
Vienna_rna will be tested automatically with conda.

dmaticzka:
Having text input is nice. I consider file input the default, for that reason I would make the text field optional.

We should  not use dbn since the RNAfold output does not conform to the galaxy dbn definition:

  * dbn allows only a single dot-bracket line, RNAfold has multiple
  * dbn dot-bracket must have same length as sequence, RNAfold adds energy
  * dbn does not allow for comments, has RNAfold additional line with energies

The old value "tabular" is not right either, this should be simply "txt"?!

for reference RNAfold example output:

```
>testn_1
UCACGACGCCCGGGUCAGGGCCUGGGCGCGGCUCGGAGCCGGGUCGUCGCUAUAGCAAUCGGGAACCGCUACCUGUCCAUUCCAAGGAUUCGG
...((((((((((.((.(((((.......))))).)).))))).)))))...((((...........))))((.((((.......))))..)) (-37.70)
...((((((((((.((.(((((.,....,))))).)).))))).)))))...{{{{...,{{...,,)}},{(.((((.......))))..)) [-39.06]
...((((((((((.((.(((((.......))))).)).))))).)))))....(((...........))).((.((((.......))))..)) {-36.20 d=7.11}
frequency of mfe structure in ensemble 0.109794; ensemble diversity 10.21
```

yhoogstrate:
Hi guys,

I added the text command as default because I expected this to be the major use case, but reading from your comments it's not. I will revert this! Thanks for the input.
If text input allows multilines, we should definitely check for the fasta header tag. Nice one @bgruening.

@dmaticzka you are right that there are limitations with the current DotBracket format. With respect to dbn I would like to see all tools (unafold, mfold, vienna, one I wrote myself, and there must be plenty others around) that can create a DotBracket*-like* structure using 1 uniform DotBracket datatype by galaxy such that we can use a vizualization (e.g. https://github.com/bgruening/galaxytools/tree/master/visualisations/drawrnajs) and simplify the handshake with other tools. If we can figure out what the desired format should be I wouldn't mind to update the galaxy datatype definition. If we can generalize this, the result will be highly integrated.
If the derivatives of the DotBracket format are too different we can do a manual conversion from dotbracket to CT which probably gives less problems.

We could leave this PR open until we have resolved the DotBracket, or do a DotBracket migration as a separate PR? That way we can tackle all DotBracket problems at once while the text input can get in already.

bgruening:
@dmaticzka I'm also not so concerned about the strict definition of dbn. We should make sure that Galaxy can recognise it in some way and that tools can exchange formats (in a more strict way than text). If one tool consumes a dbn with comments and can not deal with it, this is bad and we should think about other solutions, like @yhoogstrate mentioned ... new filetypes, conversions ... etc. But this exact problem we already have now, but worse as text can be anything. So I see it as improvement to the current state, not ideal, but a step in the right direction.

Not speaking of the vienna_rna guys that should actually fix the output ;)
@yhoogstrate I would like to highlight your point here, we should do the migration at once. So I'm happy to accept a PR that is changing the datatypes for all tools.

What ever we decide the text input changes can definitely go in. So feel free to decide if you want to split this PR up into two.

Thanks guys!

dmaticzka:
Uh, I simply assumed that not conforming to strict dbn would break stuff.
If that's not the case I'm all for using it. In the current state none of
these tool outputs is usable without at least one custom awk call. Making
the outputs at least somewhat dbn compatible would simplify things greatly.

Here's some thoughts on what to do about the current dbn definition. First
of all this needs to be less strict regarding extra characters in sequence
and structure to enable RNA-RNA hybridizations (RNAcofold uses "&" to
separate the two input sequences, I think).

Before implementing anything we probably should discuss having single vs.
multiple structures per header.

Let's say we keep dbn mostly as it is now as multiples of header, sequence
and structure; then we should be able to cast almost all outputs into it by
putting the meta-information of individual structures into the headers.
This will however hugely inflate anything that predicts multiple structures
(RNALfold, RNAsubopt).

If we extend the format to multiple structures per header, we must put meta
information next to the structures. This probably means modifying dbn to
allow anything after the first whitespace or something similar. To be
usable we would probably have to modify all tools that rely on the current
dbn definition. No idea how large a box this would open :).

What's your take on this?


On Wed, Mar 2, 2016 at 5:33 PM, Björn Grüning <notifications@github.com>
wrote:

> @dmaticzka <https://github.com/dmaticzka> I'm also not so concerned about
> the strict definition of dbn. We should make sure that Galaxy can recognise
> it in some way and that tools can exchange formats (in a more strict way
> than text). If one tool consumes a dbn with comments and can not deal with
> it, this is bad and we should think about other solutions, like
> @yhoogstrate <https://github.com/yhoogstrate> mentioned ... new
> filetypes, conversions ... etc. But this exact problem we already have now,
> but worse as text can be anything. So I see it as improvement to the
> current state, not ideal, but a step in the right direction.
>
> Not speaking of the vienna_rna guys that should actually fix the output ;)
> @yhoogstrate <https://github.com/yhoogstrate> I would like to highlight
> your point here, we should do the migration at once. So I'm happy to accept
> a PR that is changing the datatypes for all tools.
>
> What ever we decide the text input changes can definitely go in. So feel
> free to decide if you want to split this PR up into two.
>
> Thanks guys!
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/bgruening/galaxytools/pull/303#issuecomment-191316356>
> .
>



-- 
--------------------------------------------------
Daniel Maticzka, Dipl. Inf.
Lehrstuhl für Bioinformatik
Institut für Informatik
Albert-Ludwigs-Universität Freiburg

Tel: (+49) 761 203 97529
Fax: (+49) 761 203 7462
Mail: maticzkd@informatik.uni-freiburg.de
--------------------------------------------------

yhoogstrate:
@dmaticzka Great to have such expertise here :)

The only problem with galaxy's stricter datatype definition is that if you would produce a file by Vienna that doesn't fit the galaxy definition, and you would download and re-upload it, it would be marked as 'text'. If we would set the input of downstream tools to `format="dbn,text"` there would be no problem I guess.

The more frustrating problem is the handshake with other tools, like visualization tools. I think in this case it's even worse as the FASTQ derivatives problem.

In the example you showed above there are a few things that I am worried about:
 - Including pseudoknots in dotbracket is a more or less a hack because the format is not designed for this in the first place. The recursion depth is limited by the number of used chars/2 (with chars I mean e.g. curly brackets instead of parenthesis). If you would scale a structure up and increase the number of recusive pseudoknots, you will eventually find a structure that can not be explained in DotBracket format because you run out of characters. Anyway, they're there so we have to come up with some sort of solution I guess, but I would argue that CT would make more sense in such cases.
 - Using parenthesis to separate the energy or probabilities from the structrure. The parenthesis are the most prominent chars of the structure as well so this is confusing. There is the header line, that should be used to describe information with respect to the sequence and structure (as long as we would produce 1 structure per sequence).

> Before implementing anything we probably should discuss having single vs. multiple structures per header.

This is absolutely true. From the one hand it's useless to keep copying the sequence for every derivate structure (*inflate*, as you call it). On the other hand it's terrible to extend the structure with suffixes like (dE -23.00kcal/m) while there is a header line supposed to contain metadata. For 1-sequence-1-structure data it speaks for itself and I would put everything within the header.

> Let's say we keep dbn mostly as it is now as multiples of header, sequence and structure; then we should be able to cast almost all outputs into it by putting the meta-information of individual structures into the headers.

This was what I thought as well and this is what has my preference, because that's what a header should be there for. Just to make sure we're on the same page, you would get something like this?:

```
>seq1
GGGaaaCCC
(((...))) (-6.5)
((....)). (-6.2)
(.......) (-6.1)

```
To be replaced with:

```
>seq1 - possibility 01: (-6.5)
GGGaaaCCC
(((...)))
>seq1 - possibility 02: (-6.2)
GGGaaaCCC
((....)).
>seq1 - possibility 03: (-6.1)
GGGaaaCCC
(.......)
```

> This will however hugely inflate anything that predicts multiple structures (RNALfold, RNAsubopt).

There is always the possiblity to make a choice. For example, if this gives problems with the handshake with other vienna tools, we could add a groomer like program to convert to 'unified' dbn or make two possible outputs; the original ('text') and the re-formatted 'dbn'.

The options I see:
 - One single format: header-sequence-structure triplets. For multiple structures per seqeuence, create redundant entries. (has my preference)
 - Two DBN-like formats, one for single and multiple. For the single we shouldn't allow suffixes to the structure line, for multiple we could.

And for pseudoknots:
 - Update the DBN definition and add chars like curcly brackets to the regex. I expect that different tools use different charsets.
 - Convert to CT

For the pseudoknots I do not really have a preference, but I can imagine that when we're starting to implement visualization tools the first choice will give trouble and will take more work (because there can be multiple ways to define the same structure by using different charsets).

Then for the hybridyzation sequences we should know how different tools implement this. If the one package uses an '&' to separate and the other uses e.g. 2 sequences delimited with a '\n' we're getting into problems. If using '&' is generic, I would say update the galaxy definition. I haven't really worked with this type of data so I don't know what's common practice.

yhoogstrate:
Hey guys,

I have rebased the PRs with the travis branch and I addressed the changes. It seems that the test passes so I think it's good to go.

bgruening:
@yhoogstrate there is something wrong with the rebase, can you please re-check?
Thanks!

yhoogstrate:
@bgruening I don't know what you ment by that something was wrong. However, I merged with master and the PR looks normal to me and there seem to be no conflicts. If you still see any odd behaviour please let me know and I'll try to fix it.

bgruening:
@yhoogstrate last time I saw this PR the commits from the last 14 days where included. Now it looks good! Thanks!

Issue #269, opened 2015-11-19
ViennaRNA tools needs some love


bgruening:
(also bag_ you might rework the viennaRNA tools eventually. I always find "--noSomething; No somethings" and then a yes/no option horribly confusing. Rewording it can make it a bit clearer, e.g. "Allow Somethings? Y/N"

TorHou:
@zwanli is working on this

zwanli:
Labels have been updated

bgruening:
Fixed by #416

jfallmann:
There is still room for love, 
many tools need a workover and we really have to think about converting outputs into something usable within the galaxy framework
Big issues with RNAz and other tools, mostly options not working correctly and not being tested
we also need wrappers for some of the pre and post processing tools

TorHou:
Pre- and post- processing tools that are part of the Vienna package? Or external tools?

I wanted to suggest to get a Hiwi working on this, but I guess at this point one would need somebody with more intimate knowledge of the Vienna tools.

jfallmann:
consider this a we'll take care of that, but it will take some time

PR #304, opened 2016-03-02
add RNALfoldz to vienna

dmaticzka:
* RNALfoldz is RNALfold with -z option set
* created a separate tool because:
  * output format changes
  * the maximum span is 400 compared to 5000 of RNALfold
  * has additional citation

* bonus:
  * 5000 is actually not a sufficient maximum for RNALfold, changed to 10k

not sure how to add the citation in combination with the one already there via macro.xml, should add

```
Gruber, A. R., Bernhart, S. H., Zhou, Y. & Hofacker, I. L. RNALfoldz: efficient prediction of thermodynamically stable, local secondary structures. GI-Edition Lecture No in Informat 12 (2010).
```

### DIFF
--- a/tools/rna_tools/vienna_rna/rnalfoldz.xml
+++ b/tools/rna_tools/vienna_rna/rnalfoldz.xml
@@ -1,0 +1,96 @@
+<tool id="rnalfold" name="RNALfold" version="2.1.6.0">
+    <description>calculates locally stable secondary structures of RNA</description>
+    <expand macro="requirements" />
+    <expand macro="version_command" />
+    <expand macro="stdio" />
+    <macros>
+        <token name="@EXECUTABLE@">RNALfold</token>
+        <import>macros.xml</import>
+    </macros>
+    <command>
+<![CDATA[
+        RNALfold -L $span -T $temperature -d$dangling < $fasta_input | sed -r 's/(^[.()]*) +(\(.*\)) +(.*)/\1\t\2\t\3/' > $out_file
+        #if $varExists('$advancedOptions.noconversion')
+            #if $advancedOptions.noconversion
+                --noconv
+            #end if
+            #if $advancedOptions.gquad
+                --gquad
+            #end if
+            #if $advancedOptions.nolp
+                --noLP
+            #end if
+            #if $advancedOptions.nogu
+                --noGU
+            #end if
+            #if $advancedOptions.noclosinggu
+                --noClosingGU
+            #end if
+            #if $advancedOptions.notetra
+                --noTetra
+            #end if
+        #end if
+]]>
+    </command>
+    <inputs>
+	<param format="fasta" name="fasta_input" type="data" label="FASTA file"/>
+        <param name="span" type="integer" min="0" max="5000" value="150" label="base pair span" help="Maximal distance between two paired bases. (-L)"/>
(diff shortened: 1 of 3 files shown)


tools/rna_tools/vienna_rna/rnalfoldz.xml
```diff
@@ -0,0 +1,97 @@
...
+    <expand macro="requirements" />
+    <expand macro="version_command" />
+    <expand macro="stdio" />
+    <macros>
+        <token name="@EXECUTABLE@">RNALfold</token>
+        <import>macros.xml</import>
+    </macros>
+    <command>
+<![CDATA[
+        RNALfold -L $span -T $temperature -d$dangling -z$zthresh < $fasta_input | sed -r 's/(^[.()]*) +\((.*)\) +(.*) +z= +(.*)/\1\t\2\t\3\t\4/' > $out_file
```
yhoogstrate: The only three arguments of sed that are POSIX standard are `-e`, `-f`, `-n`

http://pubs.opengroup.org/onlinepubs/9699919799/utilities/sed.html

If there is a way to achieve the same without `-r` it would be even better

tools/rna_tools/vienna_rna/rnalfoldz.xml
```diff
@@ -0,0 +1,97 @@
...
+large genomes for short RNA structures*.
+Output consists of a list of secondary structure components of size <= L, one
+entry per line. Each output line contains the predicted local structure its
+energy in kcal/mol, the starting position of the local structure and its z-score.
+
+-----
+
+**Input format**
+
+- RNALfoldz requires one input file in FASTA format
```
yhoogstrate: `s/RNALfoldz/RNALfold/` ?

tools/rna_tools/vienna_rna/rnalfoldz.xml
```diff
@@ -0,0 +1,97 @@
...
+            #end if
+            #if $advancedOptions.notetra
+                --noTetra
+            #end if
+        #end if
+]]>
+    </command>
+    <inputs>
+	<param format="fasta" name="fasta_input" type="data" label="FASTA file"/>
+        <param name="zthresh" size="6" type="float" value="0" label="threshold for z-score computation (-z)" help="-z"/>
```
yhoogstrate: min? max? default (seems to be -2 https://www.tbi.univie.ac.at/RNA/RNALfold.1.html)?

tools/rna_tools/vienna_rna/rnalfoldz.xml
```diff
@@ -0,0 +1,97 @@
...
+            <option value="0">ignore dangling ends (0)</option>
+            <option value="1">unpaired bases participate in one dangling end only (1)</option>
+            <option value="3">allow coaxial stacking (3)</option>
+        </param>
+        <conditional name="advancedOptions">
+            <param name="advancedSelector" type="select"  label="advanced options">
+                <option value="basic">basic Options</option>
+                <option value="advanced">advanced Options</option>
+            </param>
+            <when value="advanced">
```
yhoogstrate: Bjoern indicated that double negatives can be confusing (https://github.com/bgruening/galaxytools/issues/269). I addressed some of them in the other PR you can steal it from there

bgruening:
I have talked to @dmaticzka and we agreed to use macros here to share more code between both tools.

PR #308, opened 2016-03-10
Scipy toolkit

pkohvaei:
Protease warpper folder added.


tools/protease/protease.xml
```diff
@@ -0,0 +1,62 @@
+<tool id="protease" name="Protease Model Driver" version="@VERSION@">
```
bgruening: The name should be changed and the ID as well. Currently it's not clear for the user what this tool is doing.

PR #312, opened 2016-03-15
handle compressed files in trim galore

bgruening:
Patch by @dpryan79

### DIFF
--- a/.tt_blacklist
+++ b/.tt_blacklist
@@ -55,7 +55,6 @@ tools/openms
 tools/bigwig_to_bedgraph
 tools/iprscan5
 tools/segemehl
-tools/trim_galore
 tools/deseq2
 tools/labels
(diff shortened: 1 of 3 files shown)


dpryan79:
I'm testing this currently. Should probably do that before merging :)

dpryan79:
It looks like trim galore renames things if the input is gzipped. That's annoying and breaks the output file inclusion in Galaxy. I'll work on a fix.

dpryan79:
I have a working version of this locally. I'll complete it and make a PR tomorrow. It's still kind of hacky (again with the `endswith(".gz")`, but it's better than nothing until galaxy starts handling compressed fastq files better.

bgruening:
Ok, feel free to push against this branch. It has testing enabled.

PR #313, opened 2016-03-17
Trim Galore! modifications for gzipped file handling

dpryan79:
This is still a bit hacky, but I've tested SE/PE/PE collections with both compressed and uncompressed files locally and everything seems to work. I guess if/when Galaxy starts supporting gzipped fastq files innately that a lot of scripts will need to start doing stuff like this.

### DIFF
--- a/tools/trim_galore/trim_galore.xml
+++ b/tools/trim_galore/trim_galore.xml
@@ -48,8 +48,8 @@
         <requirement type="package" version="1.8.3">cutadapt</requirement>
         <requirement type="package" version="1.8">cutadapt</requirement>
     </requirements>
-    <version_command>
-        perl $__tool_directory/trim_galore --version
+    <version_command interpreter="perl">
+        $__tool_directory/trim_galore --version
     </version_command>
     <command>
 <![CDATA[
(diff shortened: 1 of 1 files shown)


dpryan79:
BTW, this PR is targeted at your `trim_galore_gz` branch.

tools/trim_galore/trim_galore.xml
```diff
@@ -48,8 +48,8 @@
         <requirement type="package" version="1.8.3">cutadapt</requirement>
         <requirement type="package" version="1.8">cutadapt</requirement>
     </requirements>
-    <version_command>
-        perl $__tool_directory/trim_galore --version
```
bgruening: This was not working? Than it's a Galaxy bug. `interpreter` is deprecated.

dpryan79: I'd have to double check. I saw a few instances of it returning `perl $__tool_directory/...` as the version string, which seemed rather odd. Perhaps I had mucked something up, though. I'll test real quick.

dpryan79: Ah, the original `<version_command>` gives the following error message: `Can't open perl script "/trim_galore": No such file or directory`. My guess is that `$__tool_directory` isn't set.

bgruening: Ok, I will try to fix it ...

davebx: I thought it was `$__tool_directory__`? Sorry if I'm missing some obvious reason it wouldn't be.

bgruening: It is Dave, it is ... Oo

PR #317, opened 2016-03-23
Add automatic Tool Shed upload to travis

bgruening:
For now only TTS.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,12 +20,18 @@ install:
   - cat changed_repositories.list
 
 script:
-  - for DIR in `cat changed_repositories.list`; do planemo conda_install --conda_ensure_channels bgruening,bioconda,r --conda_prefix $CONDA_PREFIX $DIR ; done
-  - for DIR in `cat changed_repositories.list`; do planemo test --conda_dependency_resolution --conda_prefix $CONDA_PREFIX --galaxy_branch $GALAXY_RELEASE  --galaxy_source $GALAXY_REPO --skip_venv $DIR || exit 1; done
+  - |
+    if [ -s changed_repositories.list ]
+    then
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -22,3 +22,10 @@ install:
 script:
   - for DIR in `cat changed_repositories.list`; do planemo conda_install --conda_ensure_channels bgruening,bioconda,r --conda_prefix $CONDA_PREFIX $DIR ; done
   - for DIR in `cat changed_repositories.list`; do planemo test --conda_dependency_resolution --conda_prefix $CONDA_PREFIX --galaxy_branch $GALAXY_RELEASE  --galaxy_source $GALAXY_REPO --skip_venv $DIR || exit 1; done
+
+after_success:
+  - | 
+    if [ "$TRAVIS_PULL_REQUEST" == "false" -a "$TRAVIS_BRANCH" == "master" ]
+    then 
+      for DIR in `cat changed_repositories.list`; do planemo shed_upload --shed_target testtoolshed --shed_email $SHED_EMAIL --shed_password $SHED_PASSWORD --force_repository_creation $DIR || exit 1; done
```
nsoranzo: Shouldn't this use `shed_update` instead of the low-level `shed_upload`?

bgruening:
@nsoranzo @davebx does this looks ok for you? I have no clue how I can test this properly so I will use my account.
If this looks ok, we can simply merge it and let's see how it goes with a few tools - sailfish maybe?

Next step would be to add MTS as well and than move this script over to IUC.

Any idea how we can share a travis file, between multiple repositories? Creating scripts out of it and putting these into a submodule?

nsoranzo:
Looks like it worked :+1:

bgruening:
Yes it does! Amazing! We should do a negative test as well, not that some beta stuff is committed.

PR #320, opened 2016-03-29
Switch over to conda based planemo install

bgruening:
I found a conda based planemo much more stable than the pip one. Let's test this on travis as well.


jmchilton:
Why is it different? :confused:

I use the brew locally instead of the pip one so I'm not complaining - but if there is specific problems with what is in pip we should address them :smile:.

PR #324, opened 2016-03-30
numeric_clustering tailored for column_data.

pkohvaei:
Numeric_clustering with coulmn_data in "inputs" configfile.

### DIFF
--- a/tools/sklearn_tools/numeric_clustering.xml
+++ b/tools/sklearn_tools/numeric_clustering.xml
@@ -32,8 +32,6 @@
     </macros>
     <version_command>echo "@VERSION@"</version_command>
     <command><![CDATA[
-        cat "$inputs" >&2
-        &&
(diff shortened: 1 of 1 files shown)


tools/sklearn_tools/numeric_clustering.xml
```diff
@@ -32,6 +32,8 @@
     </macros>
     <version_command>echo "@VERSION@"</version_command>
     <command><![CDATA[
+        cat "$inputs" >&2
```
bgruening: This debug can be removed.

bgruening:
Just as reference, this is only working with latest Galaxy dev.

tools/sklearn_tools/numeric_clustering.xml
pkohvaei: Done!

PR #326, opened 2016-04-04
Scipy toolkit

pkohvaei:
numeric_clustering updated with sparse support and fail-safe mechanism.

### DIFF
--- a/tools/sklearn_tools/numeric_clustering.xml
+++ b/tools/sklearn_tools/numeric_clustering.xml
@@ -1,4 +1,4 @@
-<tool id="sklear_numeric_clustering" name="Numeric Clustering" version="@VERSION@">
+<tool id="sklearn_numeric_clustering" name="Numeric Clustering" version="@VERSION@">
     <description></description>
     <expand macro="python_requirements"/>
     <expand macro="macro_stdio"/>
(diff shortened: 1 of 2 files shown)


tools/sklearn_tools/numeric_clustering.xml
```diff
@@ -50,13 +50,20 @@ from scipy.io import mmread
...
+
+my_class = getattr(sklearn.cluster, selected_algorithm)
 cluster_object = my_class()
 options = params["algorithm_options"]["options"]
+
 cluster_object.set_params(**options)
 
 #if $input_types.selected_input_type == "sparse":
-data_matrix = mmread(open("$infile", 'r'))
+if selected_algorithm in ["Birch","DBSCAN","KMeans","MiniBatchKMeans","SpectralClustering"]:
```
bgruening: You could put the `<conditional name="algorithm_options">` into the sparse conditional and remove/add algorithm that are only supported by sparse ... Not sure if this is easier for the user, but it would catch this before submitting the job.

pkohvaei: All the algorithms which support sparse also support normal arrays...

bgruening: Have a look at your error message. If you want to avoid this you can simply add the algorithm conditional under the input_types one.

pkohvaei:
Now! Looks less troubled I suppose, though now I have an overhead in the algorithm_options section. I think users will like it anyway ;)

Issue #331, opened 2016-04-12
can we somehow capture and show trimgalore's errors?


martenson:
original shared history
https://usegalaxy.org/u/bradlanghorst/h/custom-adapter-trimming

there was an error fetching fastqs and the inputs contained only HTML so trim galore failed, however it provided this useless message:
```
Trim Galore! on data 2 and data 1: trimmed reads pair 2
An error occurred with this dataset:
1.8
Cutadapt terminated with exit signal: '256'.
Terminating Trim Galore run, please check error message(s) to get an idea what went wrong...
```

bgruening:
Not sure how I can improve this. I don't remove any output. Maybe it is written to the repooort?
https://github.com/bgruening/galaxytools/blob/master/tools/trim_galore/trim_galore.xml#L253

bgruening:
ping @bwlang

bwlang:
not sure if warnings will cause any other trouble... seems to be ok in my local instance.

bwlang:
now with invalid input i see:
```
Input file './input_singles' seems to be completely empty. Consider respecifying!

Path to Cutadapt set as: 'cutadapt' (default)
1.5
Cutadapt seems to be working fine (tested command 'cutadapt --version')


AUTO-DETECTING ADAPTER TYPE
===========================
Attempting to auto-detect adapter type from the first 1 million sequences of the first file (>> ./input_singles <<)

Found perfect matches for the following adapter sequences:
Adapter type	Count	Sequence	Sequences analysed	Percentage
```

nsoranzo:
@bgruening report will be filled only if trim_galore succeeds, as `<command>` is written now.

bgruening:
@nsoranzo this is what I'm saying maybe we can improve this. But the `supress_warn` options can also be a way to solve this.
@bwlang not sure why this is failing now :(

bgruening:
@bwlang my best guess is that without supress_warn trim_galore returns an error code != 0. So we might want to use `<stdio>` here.

bwlang:
stdio block rescues a couple of tests... the rest have changed output.
any preference on whether to change the expected output to match what's generated with warnings on vs. a modification to the tests to be more flexible (e.g. re_match?)

bgruening:
re_match :)
Thanks @bwlang!

jennaj:
@bwlang @bgruening Is this still in progress? Would be very helpful to have the reasons. Tx!

bwlang:
[...]

PR #332, opened 2016-04-12
shows warnings to give a user a fighting chance

bwlang:
re #331


bwlang:
I think this resolves the tests...
I applied planemo normalize, hope that's what you want
One potential concern, I don't know what "set compressed = 'no'" does, i just kept it from your upstream changes.

bgruening:
@bwlang planemo normalise does nothing good at the moment, can you revert this commit? It just scrambles the order of attributes and removes &&

bgruening:
Thanks for working on this.

bwlang:
this fails local tests... may gzip differences. i'll push one more to loosen up the size comparisons a bit.

bwlang:
ok... maybe this is ok for you?

bgruening:
Thanks a bunch @bwlang!

PR #333, opened 2016-04-18
minor update to lighter

yhoogstrate:


bgruening:
Thanks! Do you need it in the TS soon?

yhoogstrate:
Nope, I think deseq has a higher priority

PR #339, opened 2016-05-08
ChemicalToolBox updates

bgruening:
This will bulk update the ChemicalToolBox to use conda dependencies and macros.


bgruening:
@gregvonkuster fyi ... huge update to the CTB. I need to fix a few other tools but overall this seems to be huge for CTB.

PR #340, opened 2016-05-09
De-multiplex the ImageJ2 tools

gregvonkuster:
This PR eliminates the imagej2_convert_formats tool since it used a different tool_dependencies.xml recipe from all other tools in this suite (it had a dependency on bioformats).  That tool is really not needed (it was the first tool I wrote for this suite, and I later discovered it was not necessary).  If others would like the tool to be re-added, then it can be slightly tweaked to eliminate the use of Bioformats and use simply ImageJ2 since it provides image formate conversion commands as well.  If it is decided that this tool should be re-added, it should be in a separate PR.


bgruening:
This is great! 800 lines code less! Thanks a lot!

PR #341, opened 2016-05-09
Trigger test for fiji tools

bgruening:


bgruening:
@gregvonkuster we get a timeout here? Can it be that fiji is trying to start in graphical mode and just times out?

gregvonkuster:
@bgruening Hmm. I don't think the problem is that it is trying to start the UI. I'm using --headless, which keeps things from starting up the UI.  However, running the fiji executable even without the UI takes a bit of time.  Is there a way to extend the timeout period?

bgruening:
I guess I know what it is.
https://github.com/bgruening/galaxytools/blob/fiji_test/tools/image_processing/imagej2/imagej2_base_utils.py#L26

To work with conda we need to ensure that the command starts with `fiji` (the bash wrapper) and not with java ....

gregvonkuster:
@bgruening I'm not quite clear on how this needs to be changed.  If I understand correctly, the problem lies only with the bunwarpj tools since the command line to execute them is java rather than the ImageJ executable.  The reason for this is because the bunwarpj tools include an API so they can be executed from the command line outside of the ImageJ utility, while other ImageJ tools were not.  I assume the fiji.sh script needs to be enhanced to call these bunwarpj tools in a special way, but I'm not clear on what needs to be done.  Do you have some pointers?

bgruening:
My bet would be to create a `bunwarpj.sh` wrapper.
Related: https://github.com/bioconda/bioconda-recipes/issues/1519

gregvonkuster:
Thanks for the pointer @bgruening, so we will have a fiji.sh script and a bunwarpj.sh script, with build.sh enhanced to look something like this?

cp $RECIPE_DIR/fiji.sh $outdir/fiji
ln -s $outdir/fiji $PREFIX/bin
chmod 0755 "${PREFIX}/bin/fiji"
cp $RECIPE_DIR/bunwarpj.sh $outdir/bunwarpj
ln -s $outdir/bunwarpj $PREFIX/bin
chmod 0755 "${PREFIX}/bin/bunwarpj"

bgruening:
I guess so. This Java stuff is still way to complicated for me :(
Essentially what this shell script tries is to abstract the `--jar` path away, because you can not know where this is stored in the conda installation. 

ImageJ-linux64 is on $PATH correct? So this should not be a problem for these tools.

bgruening:
@gregvonkuster do you have time to update the wrappers for the new .sh script in conda. I hope these will fix the tests.

gregvonkuster:
@bgruening I'll try to find some time, but I don't know what needs to be changed in the wrappers.  Can you provide some details?  Here is the test error, but I'm not sure if the wrappers need to be updated to use a new version of samtools or something else.  Thanks!  ;)

Error: HTTPError: 404 Client Error: Not Found for url: https://conda.anaconda.org/bioconda/linux-64/samtools-1.3.1-0.tar.bz2: https://conda.anaconda.org/bioconda/linux-64/samtools-1.3.1-0.tar.bz2

bgruening:
I thought you need to change this command at least: https://github.com/bgruening/galaxytools/blob/master/tools/image_processing/imagej2/imagej2_base_utils.py#L26

I rerun the test, this error looks wired.

gregvonkuster:
Should the command now somehow call bunwarpj.sh?  The internals of this conda stuff are still a black box to me.  :(

bgruening:
Why? Anything I can explain. the sh script is just a wrapper around `java -jar PATH` because you can not know the PATH if you install it via conda. On every machine it's different. So the sh scrips adds some magic to automatically find the JAR file and execute `java -jar ` for you.

gregvonkuster:
Sorry, I'm just not clear on how the back-end stuff works yet.  I haven't had time to look too deeply into it.  Did the tests pass for all tools wrapped by the fiji.sh script or have they failed as well?  If they are failing, then I assume the wrappers for those tools need to updated too.  Similar to how the bunwarpj.sh script call Java, the fiji.sh script calls the ImageJ executable.  So do  the tools wrappers need to be changed to call bunwarpj and fiji respectively?

bgruening:
It hanging here I guess: 
```imagej2_adjust_threshold_binary.py --input "/tmp/tmpHchwaQ/files/000/dataset_1.dat" --input_datatype gif --threshold_min 0.0 --threshold_max 129.0 --method Default --display red --black_background no --stack_histogram no --jython_script /home/travis/build/bgruening/galaxytools/tools/image_processing/imagej2/imagej2_adjust_threshold_binary_jython_script.py --output_datatype gif --output "/tmp/tmpHchwaQ/files/000/dataset_2.dat"
```

gregvonkuster:
@bgruening Sorry for the delay on this.  I guess what I need for this is to better understand the way this is being tested and how the tools should be executed from the command line (either within or outside of a Galaxy environment) and how that differs from what we were doing before.

I assume planemo is being used here.  Are the steps for testing defined here: https://github.com/bgruening/galaxytools/blob/master/.travis.yml?  If so, I assume that after execution the conda build step, planemo should be used as described.  Can you confirm this?

Is there some documentation that provides the context about how to now write tools to use the new shell script wrappers for Java?

I'll try hard to set aside some time this week to take another look at this, but I'll definitely need some directions.  :)

bgruening:
> @bgruening Sorry for the delay on this.  I guess what I need for this is to better understand the way this is being tested and how the tools should be executed from the command line (either within or outside of a Galaxy environment) and how that differs from what we were doing before.

1. Travis is using `planemo conda_install ./path/tool.xml` to create a conda environment.
2. planemo test is executed with conda_dependency_resolver enabled
3. Galaxy (from planemo) will see the conda environment from (1)

So what ever is in the conda environment should be used in the wrapper - I guess in this special case here we need to call the bash-script (that wraps java -jar). You can easily test locally by installing fiji from conda and see how to use it.

> I assume planemo is being used here.  Are the steps for testing defined here: https://github.com/bgruening/galaxytools/blob/master/.travis.yml?  If so, I assume that after execution the conda build step, planemo should be used as described.  Can you confirm this?

Yes!

> Is there some documentation that provides the context about how to now write tools to use the new shell script wrappers for Java?

`fiji --parameters -param` should work I guess. 
@gregvonkuster have a look at this issue: https://github.com/bioconda/bioconda-recipes/issues/1519
It is a python port of the bash wrapper, maybe it get's more clear what this thing is doing.

Thanks!

gregvonkuster:
@bgruening Thanks for the info above.  It provides some additional context, and I've taken another look at changing these tools to use conda.  But I am still not clear on why we need some of this, so I need some more input to make it work.

We currently have 2 categories of Galaxy tools for Fiji:
1) Tools that wrap the bunwarpj plugin - these execute java directly from the command line since the bunwarpj plugin provides its own API, and the ImageJ2 executable does not need to be called directly.

Since these tools call java directly, I guess we need the bunwarpj.sh bash script to provide the path to java and the specific jar files needed by bunwarpj.  But it is not clear if / how the tool config should call the bunwarpj.sh bash script.

Here is the command line from the imagej2_bunwarpj_adapt_transform.xml tool config:

```
    python $__tool_directory__/imagej2_bunwarpj_adapt_transform.py
    --target_image "$target_image"
    --target_image_format $target_image.ext
    --source_image "$source_image"
    --source_image_format $source_image.ext
    --input_elastic_transformation "$input_elastic_transformation"
    --image_size_factor $image_size_factor
    --output "$output"
```
When this tool is executed, the imagej2_bunwarpj_adapt_transform.py script builds a different command line that looks something like this:

```
java -cp <fiji jar dir>/ij-1.49k.jar:<fiji plugin dir>/bUnwarpJ_-2.6.1.jar bunwarpj.bUnwarpJ_ -adapt_transform <target image> <source image> <elastic transformation> <output image>
```
The above command is executed using:
```
subprocess.Popen( args=<above command>, stderr=subprocess.PIPE, stdout=subprocess.PIPE, shell=True )
```

So does the above java command need to be changed to call the bunwarpj.sh bash script?  If so, then the current bunwarpj.sh bash script needs to handle parsing the command line correctly.


2) Tools that call the ImageJ2 package - these tools execute Python from the command line, and all Java stuff is encapsulated within a Python script passed as a command line parameter.

I guess the tools in category 2 have to call the fiji.sh bash script.  But the fiji recipe installs the ImageJ2 package and makes it available from the command line (with no need for calling java directly), so what is the reason that we need the fiji.sh bash script?  Can't we just treat these tools like all other tools that have nothing to do with java?

bgruening:
I guess this can be closed now. Thanks Greg!

PR #350, opened 2016-05-30
Text utils condafication

bgruening:


tools/text_processing/text_processing/tool_dependencies.xml
```diff
@@ -1,34 +0,0 @@
-<?xml version="1.0"?>
```
jmchilton: Doesn't this mean by default with the "recommended" approach to tool installs - Galaxy will now use the insecure system version of sed?

bgruening: I have not planned to upload this one to the MTS until we have discussed the "recommended" system at GCC.
I hope this is ok. I could leave the dependencies in, but the version do not match anymore and it would mean new tool_dependencies.

jmchilton: Of course it is okay.

xref https://github.com/galaxyproject/galaxy/issues/2446

PR #352, opened 2016-05-31
Scipy toolkit

pkohvaei:
Ensemble estimator tool added.


bgruening:
Incredible PR! Nice work!
Just one thing I have seen, why do you have two statistic-tools? Is one of them deprecated?

bgruening:
"Ran 67 tests in 1326.089s" :)
Can you please add a `.shed.yml` to `tools/statistics` this will also trigger tests for these tools.

tools/statistics/macros.xml
```diff
@@ -0,0 +1,262 @@
+<macros>
+  <token name="@VERSION@">0.2</token>
+
+  <xml name="python_requirements">
+    <requirements>
+        <requirement type="binary">@EXECUTABLE@</requirement>
```
bgruening: I don't think this is needed anymore.

pkohvaei:
Those are not complete yet. But I will add the yml file.

tools/statistics/macros.xml
pkohvaei: It is still using the python script. Next step that would be removed.

PR #353, opened 2016-05-31
Barcode collapse files

thaze:


tools/barcode_collapse/.shed.yml
```diff
@@ -0,0 +1,21 @@
...
+homepage_url: https://github.com/YeoLab/gscripts/releases/tag/1.0
+long_description: This tool is used in the eCLIP-seq Processing	Pipeline. Read in a '.bam' file where the first 9 nucleotides of the read name are the barcode and merge reads mapped to the same position that have the same barcode.
+Assumes paired end reads are adjacent to each other in the output file (i.e. only provide unsorted bams). It also assumes that there are no multimappers in the bam file, if there are multimappers behavior is undefined.
+    
+remote_repository_url: https://github.com/bgruening/galaxytools/tree/master/tools/barcode_collapse
+type: unrestricted
+categories:
+- RNA
+- Sequence Analysis
+auto_tool_repositories:
```
bgruening: If this is only one tool you don't need to define a suite and hence you can remove `auto_tool_repository` and `suite`

tools/barcode_collapse/.shed.yml
```diff
@@ -0,0 +1,21 @@
+name: barcode_collapse
+owner: rnateam
+description: "Paired End randomer aware duplicate removal algorithm"
+homepage_url: https://github.com/YeoLab/gscripts/releases/tag/1.0
```
bgruening: https://github.com/YeoLab/gscripts

tools/barcode_collapse/BarcodeCollapse_pe.xml
```diff
@@ -0,0 +1,48 @@
+<tool id="barcode_collapse" name="barcode_collapse" version="0.1.0.0">
+    <description>Paired End randomer aware duplicate removal algorithm !</description>
+    <requirements>
+        <requirement type="package" version="0.1.0.0">barcode_collapse</requirement>
```
bgruening: If it's just one python file that you need you can include it next to your xml file and in this PR, given that the licence allows it.

bgruening: I have seen that this tool depends on pysam, so you need to include: 
`<requirement type="package" version="0.8.4">pysam</requirement>`

tools/barcode_collapse/BarcodeCollapse_pe.xml
```diff
@@ -0,0 +1,48 @@
...
+        <exit_code range=":-1" />
+        <!-- In case the return code has not been set propery check stderr too -->
+        <regex match="Error:" />
+        <regex match="Exception:" />
+    </stdio>
+    <command>
+<![CDATA[
+        ln -s "${input_file}" infile.bam &&
+        python
+        barcode_collapse_pe.py
```
bgruening: If you include the python file next to your xml file, use $__tool_directory__

PR #357, opened 2016-06-03
Scipy toolkit

pkohvaei:
Sklearn preprocessing tool added.
Extended help will be added in the future.


tools/sklearn/pre_process.xml
```diff
@@ -0,0 +1,223 @@
+<tool id="sklearn_data_preprocess" name="Preprocess" version="@VERSION@">
+    <description>raw feature vectors into standardized datasets</description>
```
bgruening: Oh I like this tool!

bgruening:
Nice, thanks!

PR #365, opened 2016-06-19
WIP: T-Coffee

anatskiy:


tools/tcoffee/tcoffee.xml
```diff
@@ -0,0 +1,186 @@
...
+            <option value="muscle_msa">Muscle</option>
+            <option value="t_coffee_msa">T-Coffee</option>
+            <option value="pcma_msa">pcma</option>
+            <option value="kalign_msa">kalign</option>
+            <option value="amap_msa">amap</option>
+            <option value="proda_msa">proda</option>
+            <option value="prank_msa">prank</option>
+        </param>
+
+        <section name="out" title="Output Format" expanded="True">
```
bgruening: You can use here a `type="select" multiple="true"` like here: https://github.com/bgruening/galaxytools/blob/master/tools/antismash/antismash.xml#L180

tools/tcoffee/tcoffee.xml
```diff
@@ -0,0 +1,186 @@
+<tool id="tcoffee" name="T-Coffee" version="0.1">
```
bgruening: id="t_coffee"

tools/tcoffee/tcoffee.xml
```diff
@@ -0,0 +1,186 @@
...
+            $out.format_phylip
+            $out.format_pir_seq
+            $out.format_fasta_seq
+            $out.format_score_ascii
+            $out.format_score_html
+        #end if
+]]>
+    </command>
+    <inputs>
+        <param name="input" type="data" format="fasta" label="Source file" />
```
bgruening: label="sequences in FASTA format"

tools/tcoffee/tcoffee.xml
```diff
@@ -0,0 +1,186 @@
...
+
+        <param name="method1" type="select" label="Pairwise Sequence Alignment Method">
```
bgruening: Do you know what default is? You can use a `type="select" optional="True"` and remove the Default option.

tools/tcoffee/.shed.yml
```diff
@@ -0,0 +1,12 @@
+name: tcoffee
```
bgruening: t_coffee

tools/tcoffee/.shed.yml
```diff
@@ -0,0 +1,12 @@
...
+owner: bgruening
+description: A collection of tools for Computing, Evaluating and Manipulating Multiple Alignments of DNA, RNA, Protein Sequences and Structures
+homepage_url: http://www.tcoffee.org/Projects/tcoffee/index.html
+long_description: T-Coffee is a multiple sequence alignment package. You can use T-Coffee to align sequences or to combine the output of your favorite alignment methods (Clustal, Mafft, Probcons, Muscle...) into one unique alignment.
+remote_repository_url: https://github.com/bgruening/galaxytools/tree/master/tools/tcoffee
+type: unrestricted
+categories:
+- Sequence Alignment
+- DNA
+- RNA
```
bgruening: You can only use categories from https://toolshed.g2.bx.psu.edu
* Sequence Analysis

nsoranzo:
Are you aware of https://toolshed.g2.bx.psu.edu/view/ayllon/tcoffee/  ?

anatskiy:
@nsoranzo, sure, I'm basing on the mentioned tool and enhancing it a bit.

nsoranzo:
@anatskiy Have you checked with the original author if he is still maintaining it? If it's abandoned, maybe the https://github.com/galaxyproject/tools-iuc repository would be a better place for it, since IUC is usually willing to adopt abandoned tools. We are using it at TGAC, so I'd be also interested for example.

bgruening:
@nsoranzo we have contacted the authors and never got a reply.

nsoranzo:
@bgruening tools-iuc then?

bgruening:
Yes, this is an incubator :)

tools/tcoffee/tcoffee.xml
```diff
@@ -0,0 +1,174 @@
+<tool id="t_coffee" name="T-Coffee" version="0.1">
```
bgruening: Can you please add ` profile="16.04"` to the tool tag?
This will hopefully fix the stderr error.

nsoranzo:
@bgruening I actually completely forgot this PR in the rush to finish our work for http://biorxiv.org/content/early/2016/12/23/096529 , sorry about that. Our new wrapper started from https://toolshed.g2.bx.psu.edu/view/ayllon/tcoffee/ with the following modifications:
- Remove the `tcoffee.pl` Perl script
- Add CIGAR as possible output format using a Python script to convert `fasta_aln` to CIGAR
- Add requirement
- Add `<version_command>`
- Quote datasets in `<command>`
- Add multithreading support with `-n_core` option
- Use a multiselect for the output formats
- Add tests
- Fix indentation

We also added the possible to specify a second input file with a list of sequence IDs, which essentially integrates `filter_by_fasta_ids` into the `t_coffee` wrapper. I don't really like this, but we need it to reduce the number of steps in our workflow, which run on very large collections, otherwise the huge number of jobs kills our Galaxy job handlers.

To conclude, I think that we re-did most of things in this PR, except we are still using the t_coffee methods from the original wrapper, while here there seems to be more. If you want to contribute anything to our wrapper, PRs are obvoiusly welcome.

bgruening:
Crashing the job handlers? How big are your worklfows, we are running here workflows with multiple 100 steps, nearly 1000?

So this means we will close this. @anatskiy sorry!
Will this wrapper enter IUC at some point?

nsoranzo:
> - Add CIGAR as possible output format using a Python script to convert fasta_aln to CIGAR

FYI this has been removed from the wrapper, so it is now more similar to the original tool.

PR #367, opened 2016-06-23
new travis, file based

bgruening:


filter_directories.py
```diff
@@ -13,7 +19,10 @@
     while directory:
         if os.path.exists( os.path.join(directory, '.shed.yml') ) or os.path.exists( os.path.join(directory, '.shed.yaml') ):
             if directory not in bl:
-                print(directory)
+                for filename in os.listdir(directory):
```
bgruening: fyi: @nsoranzo this will enhance the built-time for large PR like mothur. We will now split by xml file not per directory.
Need to get this into IUC as well.

filter_directories.py
```diff
@@ -2,6 +2,12 @@
 
 import os
 import sys
+import lxml.etree as etree
+
+def is_tool(path):
+    xml = etree.parse(path)
+    return xml.getroot().tag == 'tool'
```
jmchilton: galaxy-lib and planemo have abstractions for this stuff - we should get this all in planemo at some point.

bgruening: Yes, ideally this should all be covered by some library so we can share this setup with multiple repositories. I'm fine if this is planemo. Are you joining the IUC meeting? This fits very much into my evil tools plans :)

filter_directories.py
nsoranzo: I suspect this may break the automatic push to TTS if there is more than one tool in a directory.

PR #375, opened 2016-07-21
should solve the downcast issue

gianlucacorrado:


bgruening:
Thanks, will test it later after finishing the grant.

PR #378, opened 2016-07-22
Migrate ImageJ2 tools to use conda fiji dependency

gregvonkuster:
Also eliminate all references to the 3rd party javabridge and bioformats packages as they were used only by a single ImageJ2 tool which has been eliminated.


packages/package_fiji_20141125/.shed.yml
```diff
@@ -1,13 +0,0 @@
-name: package_fiji_20141125
-owner: iuc
-description: Contains a tool dependency definition that downloads and extracts version
-  20141125 of Fiji.
```
bgruening: @gregvonkuster no need to remove the packages. We need to maintain them ... or wait ... they never hit the MTS, right? In this case let's kill them :)

bgruening:
@gregvonkuster the travis error indicates that the program hang for a long time. My guess would be that it's not running headless?

bgruening:
@gregvonkuster thanks for working on it!

gregvonkuster:
@bgruening I'm not sure what is causing this, but I don't think it's a headless issue.  This line https://github.com/bgruening/galaxytools/blob/master/tools/image_processing/imagej2/imagej2_base_utils.py#L35 is setting up ImageJ to run in headless mode.  Is there some information I can take a look at?  I've tried planemo test and it hangs too.  I thought this might be due to the local job runner, but wasn't sure.  Is this the runner that planemo and travis use?  Remember that the local job runner blocks when calling ImageJ.

bgruening:
Oh true there was this issue with the local job runner. Mh, this was never fixed, wasn't it?
Is it passing tests for you locally?

We will soon be able to use the Docker container with SLURM to test tools.

gregvonkuster:
Yes, the preexec in this line is the problem - it causes the job to block.
```
            proc = subprocess.Popen( args=command_line,
                                     shell=True,
                                     cwd=job_wrapper.working_directory,
                                     stdout=stdout_file,
                                     stderr=stderr_file,
                                     env=self._environ,
                                     preexec_fn=os.setpgrp )
```
Should I close this PR until we can use slurm in docker?

bgruening:
@gregvonkuster no this is great, we just need to remember this.
Thanks for you hard work here. I will update the Imaging Docker Image and we can test there for the time being and releasing it to MTS later on. ok?

gregvonkuster:
Sounds great!  Thanks @bgruening Let me know if there is anything you need from me throughout this process.

PR #383, opened 2016-07-29
updated cofold version number

HassanAmr:


HassanAmr:
I hope what I'm doing is correct.

tools/rna_tools/cofold/cofold.xml
```diff
@@ -1,7 +1,7 @@
 <tool id="cofold" name="Cofold" version="0.0.1">
```
bgruening: The tool version needs also to be bumped as soon as you change the tool. I prefer 2.0.4.0. Last digit is if you change the tool without changing the requirement.

tools/rna_tools/cofold/cofold.xml
```diff
@@ -1,7 +1,7 @@
...
     <description>An RNA secondary structure prediction method that takes co-transcriptional folding into account</description>
     <requirements>
-        <requirement type="package" version="0.0.1">cofold</requirement>
+        <requirement type="package" version="2.0.4">cofold</requirement>
     </requirements>
     <version_command>CoFold --version</version_command>
```
bgruening: Because CoFold has this annoying version error, you can also include something like `<version_command>echo "2.0.4"</version_command>`

bgruening:
@HassanAmr nice!
Next thing ... please have a look at the travis testing output. A few things got tested an it is green, this is great. But we can test even more if you remove this line from the blacklist: https://github.com/bgruening/galaxytools/blob/master/.tt_blacklist#L31

We have this blacklist because we only can test tools with a conda package. As we have now a conda package we can test it. Try it :)

bgruening:
Look at Travis! Beautiful! Everything works and is tested!

HassanAmr:
I was preparing some more test cases, or it isn't necessary anymore?

bgruening:
We can never have enough tests ;)

PR #390, opened 2016-08-20
strip-it

HassanAmr:
removed strip-it from the blacklist

### DIFF
--- a/.tt_blacklist
+++ b/.tt_blacklist
@@ -1,5 +1,4 @@
 chemicaltoolbox/osra
-chemicaltoolbox/silicos-it/shape-it
 chemicaltoolbox/silicos-it/strip-it
 tools/gotohscan
 tools/proteomics/improviser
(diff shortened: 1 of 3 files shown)


bgruening:
I think I know what it is: strip-it was compiled again openbabel-2.3 but now `openbabel-2.3.90dev7d621d9` gets installed. We need to fix the conda package and strictly depend on openbabel-2.3

HassanAmr:
In bioconda? am I allowed to just change someone else's package there?

bgruening:
Sure ;)

HassanAmr:
The package strictly depends on openbabel-2.3.2 (openbabel ==2.3.2). Why does it install 2.3.90?

bgruening:
It turns out that there is a newer version of strip_it that is not installable with the old version of conda `3.19.3`. If you use a newer version of conda you will get the latest, fixed version.

To use a newer conda version this needs to be fixed: https://github.com/conda/conda/issues/2035

PR #391, opened 2016-08-20
rnaz

HassanAmr:
removed rnaz from the blacklist


HassanAmr:
Should I add a test case here?

bgruening:
That would be great!

PR #393, opened 2016-08-20
rnacode

HassanAmr:
removed rnacode from the blacklist


HassanAmr:
Ok, now I see the problem in the other branches :)

HassanAmr:
One more question. should I change my master back to without removing shape-it, and have shape-it with its own branch, and if yes, how can I manage that?

bgruening:
This one looks good! Thanks.

PR #394, opened 2016-08-23
shape-it

HassanAmr:
removed shape-it from the blacklist

### DIFF
--- a/.tt_blacklist
+++ b/.tt_blacklist
@@ -1,5 +1,4 @@
 chemicaltoolbox/osra
-chemicaltoolbox/silicos-it/shape-it
 chemicaltoolbox/silicos-it/strip-it
 tools/gotohscan
 tools/proteomics/improviser
--- a/chemicaltoolbox/silicos-it/shape-it/shape-it.xml
+++ b/chemicaltoolbox/silicos-it/shape-it/shape-it.xml
@@ -1,5 +1,5 @@
 <tool id="ctb_shapeit" name="Shape alignment" version="0.1">
-    <description>Against a database of molecules (shape-it).</description>
+    <description>Against a database of molecules (Shape-it).</description>
     <requirements>
         <requirement type="package" version="1.0.1">shape_it</requirement>
     </requirements>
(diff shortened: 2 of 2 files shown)


HassanAmr:
Hi @bgruening ,
I think this has the same problem as targetfinder.
```
Fatal error: Exit code 127 ()
/tmp/tmpO3jvBp/job_working_directory/000/3/tool_script.sh: line 9: shape-it: command not found
```

bgruening:
I fixed the error. Can you run it locally and see if this example takes more than 10min?

HassanAmr:
The error above is from my local. The branch is up-to-date with master. I just retried it now, and it is the same.

bgruening:
The error is fixed, now it is timeout from travis. So I'm wondering if this test really takes 10 and more minutes.

HassanAmr:
Locally, it takes about a minute, but it comes back with an error.

HassanAmr:
Hi @bgruening,

I added the parameter --scoreOnly to skip performing the test. It exits with code 0, and it still processes all the parameters given in the test. But of course, the test is not meaningful in any way.
I also kept the old test output to revisit it later and check why it gets stuck while performing the test.

HassanAmr:
Hi @bgruening,

Did I miss something in this branch?

HassanAmr:
I don't get it. It was fine before the rebase

bgruening:
@HassanAmr this was a hack right? The binary is still crashing?

HassanAmr:
@bgruening, it doesn't crash, it keeps running forever. So, I changed it to scoreOnly. Should I leave this open till I figure out the problem with this package?

bgruening:
So essentially scoreOnly keeps looping? Is this reproducibly with other files?

HassanAmr:
scoreOnly works (or it used to before I did the rebase). without scoreOnly it doesn't finish. I also tried CID_2244.sdf (from openbabel), and it also didn't finish.

bgruening:
Included here: https://github.com/bgruening/galaxytools/pull/576/commits/c659c1553be79ff3590e4543af8ed009d0d5f481

PR #395, opened 2016-08-23
mcl

HassanAmr:
removed mcl from the blacklist


tools/mcl/mcl.xml
```diff
@@ -1,5 +1,5 @@
 <tool id="mcl_clustering" name="MCL"  version="14.137">
-    <description>(Markov Cluster Algorithm)</description>
+    <description>A Markov Cluster Algorithm.</description>
```
bgruening: Please remove the leading dot.

PR #397, opened 2016-08-23
rnaz

HassanAmr:
removed rnaz from the blacklist and removed period from mcl's description


bgruening:
Can you add tests for rnaz?

HassanAmr:
I believe the problem here is in the input parameters tags, am I correct? Can I change that?

tools/rna_tools/rnaz/rnaz.xml
```diff
@@ -40,6 +40,12 @@
     <outputs>
         <data name="outfile" format="fasta" />
     </outputs>
+    <tests>
+        <test>
+            <param name="input1" value="test-data/tRNA.aln"/>
```
bgruening: You don't need to specity test-data, only the name is important ... also you can add parameters here if you like and if you want to test them (you need to change the test data accordingly)

tools/mcl/mcl.xml
```diff
@@ -1,5 +1,5 @@
 <tool id="mcl_clustering" name="MCL"  version="14.137">
-    <description>A Markov Cluster Algorithm.</description>
+    <description>A Markov Cluster Algorithm</description>
```
bgruening: Why do you touch mcl in this PR. I see travis is failing because of these tests. MCL does not have yet a conda recipe, at least not in this version?

HassanAmr: Yeah, I saw that, and thought the tool depends on it. But I haven't added anything for it to go for mcl.

HassanAmr: Should I remove the mcl requirement?

bgruening: What requirement? This PR is about RNAz, isn't it Idon't understand why this change is needed in this PR :)

HassanAmr:
Hi @bgruening, I'll delete the branch, and then create a new one.

PR #398, opened 2016-08-23
segemehl

HassanAmr:
removed segemehl from the blacklist


bgruening:
@HassanAmr this needs some test cases, sorry.

HassanAmr:
Yeah, I will do some test cases for it and rnaz :)

HassanAmr:
I tried running segemehl after I installed it on my machine, but got this error:

segemehl.x: error while loading shared libraries: libncurses.so.5: cannot open shared object file: No such file or directory

After searching for possible causes for the error, it seems that I need the 32-bit library as per the following links:

http://stackoverflow.com/questions/17005654/error-while-loading-shared-libraries-libncurses-so-5
http://askubuntu.com/questions/369205/missing-libncurses-so-5

If that is really the case, how can I fix that in bioconda?

bgruening:
Ping @swill

swill:
I delivery this was the wrong mention?

bgruening:
@swill sorry!!!
I meant @s-will.

swill:
No worries. I was a bit confused for a minute. :)

bgruening:
Can you add test-data here?

HassanAmr:
Ok, I'm on it :)

HassanAmr:
@bgruening,

How can I overcome the above mentioned error?
I can't run the package to test it.

bgruening:
This should be fixed already. Do you still see the error?

HassanAmr:
Yes.
Maybe the problem is Arch linux? Because I even tried to manually install `ncurses5-compat-libs` (which according to some internet findings should fix the problem), and it seems that it cannot automatically be installed unless I invoke some configurations/modifications to the package when building.
Should I create a new virtual machine image other than Manjaro/Arch Linux (what distribution do you recommend?) to test on it?

HassanAmr:
What is `Failed linting`?

bgruening:
linting means that the strucutre of the XML file or the shed.yml file is not correct. In this case the .shed.yml file is missing the homepage_url tag:
`WARNING: Missing shed metadata field [homepage_url] for repository`

And maybe other errors, these are not strictly functional errors but styling and best-practice errors.

HassanAmr:
`RunToolException: Error creating a job for these tool inputs - Parameter index requires a value, but has no legal values defined.` How can I fix that?

jfallmann:
@bgruening guess this PR can be closed now

bgruening:
Yeah! @HassanAmr thanks for your conda package! @jfallmann fixed the rest and this is now upstream!
Thanks a bunch!

HassanAmr:
Hi @bgruening,

Sorry for not following through with this tool. 
I'm glad it is fixed now :)

PR #403, opened 2016-09-03
miRDeep2

HassanAmr:
removed miRDeep2 from the blacklist


HassanAmr:
Here, I should resolve all the dependencies (requirements) in galaxy first, before I can continue, right?

PR #404, opened 2016-09-03
samtools_sort

HassanAmr:
removed samtools_sort from the blacklist


HassanAmr:
@bgruening, the error suggests that there is something wrong with the test case. But I have the same problem as with segemehl (error while loading shared libraries: libncurses.so.5), so I cannot run it locally.

TorHou:
maybe related to this: https://github.com/rtfd/readthedocs.org/issues/1902 ?

HassanAmr:
Sorry, I broke it :(
I will do a fresh branch with the updated files.

PR #405, opened 2016-09-03
bamhash

HassanAmr:
removed bamhash from the blacklist


bgruening:
Only bamhash 1.0 seems to be available: https://github.com/bioconda/bioconda-recipes/blob/master/recipes/bamhash/meta.yaml

HassanAmr:
can I just edit it?

HassanAmr:
I think that might be the problem with many of the packages I just released.... is it possible to manually just edit the version in the xml file?

bgruening:
A new conda package would be nice

HassanAmr:
Ok, for this one I will make a new conda package, but what about the ones that have version behind the conda version? can I just edit the version number?

bgruening:
Yes. Upgrading the package is OK. Remember if you do so increase the tool version number and remove the tool_dependencies.xml file.

HassanAmr:
any convention I should follow when increasing the tool version?

bgruening:
A version close or equal to the requirement version would be nice.
On Sep 3, 2016 3:41 AM, Hassan Gamaleldin <notifications@github.com> wrote:any convention I should follow when increasing the tool version?

HassanAmr:
The bioconda package is already 1.0. Am I missing something?

bgruening:
Yeah, but the tool requirement is 1.1 :)
I will change this!

bgruening:
Hi @HassanAmr,

turns out the 1.1 version was correct: https://github.com/DecodeGenetics/BamHash/blob/dd15dee202b4d873768017b8959e8000d62ccf67/bamhash_checksum_common.h

Can you create a new conda package for the latest commit or maybe we will get soon a new release: https://github.com/DecodeGenetics/BamHash/issues/7

HassanAmr:
Hi @bgruening,

So, this should work now?

bgruening:
I get crazy: |  /tmp/tmp5O2hMi/job_working_directory/000/10/tool_script.sh: line 9: bamhash_checksum_fastq: command not found

HassanAmr:
Sorry, this one was obviously failing because I forgot to remove the dependecies xml file

bgruening:
The tool_dependencies.xml has nothing to do with this, still not clear for me what happend, but everything is green so let's merge!
Thanks!

HassanAmr:
My pleasure :)

PR #407, opened 2016-09-12
Update to 0.1.13, add the -d, --OT, --OB, --CTOT, and --CTOB options.…

dpryan79:
… Also, all mbias images are put in the history.

This should make the wrapper a bit more functional. Each of the strand images is now viewable in the history and you can now specify inclusion bounds during extraction.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,7 @@ env:
 
 before_install:
   - export GALAXY_REPO=https://github.com/galaxyproject/galaxy
-  - export GALAXY_RELEASE=release_16.04
+  - export GALAXY_RELEASE=release_16.07
   - export CONDA_PREFIX="$HOME/conda"
 
 install:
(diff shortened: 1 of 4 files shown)


bgruening:
Mh, the binary can not be found. Is the conda package broken?

dpryan79:
Odd, `conda install -c bioconda pileometh` works and PileOMeth is then in the appropriate place, so I'm not sure why that isn't working.

dpryan79:
I wonder if the issue is with however conda is setup in the test. It installs PileOMeth but the Galaxy dependency resolver then can't find that, so the environment isn't activated (if I'm reading it right): https://travis-ci.org/bgruening/galaxytools/jobs/159312659#L5032-L5034

dpryan79:
I found why it couldn't find the binary, the current version of planemo doesn't work with the 16.04 Galaxy release. Updating to 16.07 causes the test to start passing (I made one error in the wrapper that's now corrected).

bgruening:
Oh great thanks. Time to update the build system.

PR #410, opened 2016-09-27
Nucleosome prediction

ramezrawas:
I hope my first tool will be as good as it should be :)


bgruening:
The failure is not on your side, but the very old binary that is shipped with the tool is not working. I contacted the author, the wrapper looks great.

PR #416, opened 2016-10-05
ids updated

zwanli:


bgruening:
Will fix: https://github.com/bgruening/galaxytools/issues/269

PR #419, opened 2016-10-19
RNAz: update arguments, add test data 

zwanli:


tools/rna_tools/rnaz/rnaz.xml
```diff
@@ -1,45 +1,51 @@
-<tool id="rnaz" name="RNAz" version="0.1.0">
```
bgruening: Maybe version 2.1.0?

TorHou:
@bgruening delete the tool_dependencies.xml? it is now in bioconda

bgruening:
@TorHou yes, removing tool_dependencies seems to be correct!

PR #423, opened 2016-10-26
add needed zlib package dependency

martenson:
should solve https://github.com/galaxyproject/galaxy/issues/2318 for Main and similar

### DIFF
--- a/tools/diffbind/diffbind.xml
+++ b/tools/diffbind/diffbind.xml
@@ -3,6 +3,7 @@
     <requirements>
         <requirement type="binary">Rscript</requirement>
         <requirement type="R-module">DESeq2</requirement>
+        <requirement type="package" version="1.2.8">zlib</requirement>
         <requirement type="package" version="3.0.3">R</requirement>
         <requirement type="package" version="1.2.10">deseq2</requirement>
         <requirement type="package" version="1.8.3">diffbind</requirement>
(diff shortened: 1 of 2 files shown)


tools/diffbind/diffbind.xml
```diff
@@ -3,6 +3,7 @@
     <requirements>
         <requirement type="binary">Rscript</requirement>
         <requirement type="R-module">DESeq2</requirement>
+        <requirement type="package" version="1.2.8">zlib</requirement>
```
nsoranzo: Is this really needed? Maybe just adding:
```
    <repository name="package_zlib_1_2_8" owner="iuc">
        <package name="zlib" version="1.2.8" />
    </repository>
```
to `tool_dependencies.xml` (as for libxml2 in https://github.com/galaxyproject/tools-iuc/commit/8ed92ab064fd3d7e19530b5ff0bd6da0bcef8ab0 ) would be better.

martenson: I mean at times other than installation.

martenson: @nsoranzo Do you still thing your suggestion is the best way to resolve this?

nsoranzo: @martenson Yes, this can create problems in other resolvers (e.g. Conda), where diffbind may require a different zlib version.

nsoranzo: BTW, if you want to try the Conda dep, it's called `bioconductor-diffbind`.

martenson: @nsoranzo I don't want to try Conda. I want to make this work on Main.

PR #424, opened 2016-10-28
First pass at upgrading ImageJ2 tools

gregvonkuster:
to use Fiji lifeline version 20151222.  The remaining bUnwarpJ based tools will be upgraded in a separate PR (coming soon).  When both PRs are merged, I'll update the Tools Sheds with these new versions.

Current known issues:
- planemo lint imagej2_enhance_contrast.xml produces this error:

```
.. ERROR: Invalid tmpXFSCEO found. Errors [/tmp/tmpXFSCEO:30:0:ERROR:SCHEMASV:SCHEMAV_CVC_DATATYPE_VALID_1_2_1: Element 'validator', attribute 'exclude_min': 'true' is not a valid value of the atomic type 'xs:decimal'.]
```

I'm not quite sure how to resolve this...
- If executing these tools with the local Galaxy job runner, the following patch is needed.

```
index d5baa40..ce27d91 100644
--- a/lib/galaxy/jobs/runners/local.py
+++ b/lib/galaxy/jobs/runners/local.py
@@ -94,7 +94,8 @@ class LocalJobRunner( BaseJobRunner ):
                                      stdout=stdout_file,
                                      stderr=stderr_file,
                                      env=self._environ,
-                                     preexec_fn=os.setpgrp )
+                                     #preexec_fn=os.setpgrp )
+                                     )
             job_wrapper.set_job_destination(job_wrapper.job_destination, proc.pid)
             job_wrapper.change_state( model.Job.states.RUNNING 
```


bgruening:
For the fist issue, planemo is now linting XML with this XSD file: https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/tools/xsd/galaxy.xsd and it seems the XML has an invalid attribute or the XSD in incomplete.

Can you double check if 'validator', attribute 'exclude_min' and true can work together?

For the latter problem, we will not get this patch in aren't we? 
If so can you try to run the tests with 

```
--engine galaxy     Select an engine to run tools and workflows
                              using, defaults to Galaxy.
```

This will hopefully run planemo against the galaxy-stable Docker image with a non-local job runner.
We can use this for travis as well.

gregvonkuster:
@bgruening for the first issue, it looks to me like the XSD is incorrect since it requires the value of exclude_min to be an integer, but I believe it should be a boolean.  Here is an example of the attribute - see https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/tools/parameters/validation.py#L110
```
    ... <param name="blah" type="integer" size="10" value="10">
    ...     <validator type="in_range" message="Not gonna happen" min="10" exclude_min="true" max="20"/>
    ... </param>
```
If I am correct, the problem lies both in Galaxy: https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/tools/xsd/galaxy.xsd#L3008
and planemo: https://github.com/galaxyproject/planemo/blob/master/planemo/xml/xsd/tool/galaxy.xsd#L2970
Please let me know if I am correct.  If so, PRs to correct each application should be opened, correct?

bgruening:
I think you are correct! Thanks for looking into it. A PR to galaxy should be enough, planemo will use this XSD ... but we probably need to have a new release at some point.

gregvonkuster:
First issue fixed here: https://github.com/galaxyproject/galaxy/pull/3108

bgruening:
@gregvonkuster great thanks.

Can you try `planemo test --engine`.
In general this is sufficient, but in a long run we should get these tools under testing on travis as well and the --engine option is made for this. Here we can test against different job_runners etc ...

gregvonkuster:
Ah, I didn't see the --engine option because I didn't know it was associated with the test option:
```
  --engine [galaxy|docker_galaxy|cwltool]
                                  Select an engine to run or test aritfacts
                                  such as tools and workflows. Defaults to a
                                  local Galaxy, but running Galaxy within a
                                  Docker container or the CWL reference
                                  implementation 'cwltool' and be selected.
```
Is there a specific existing docker container that uses another job runner (e.g., https://github.com/bgruening/docker-galaxy-stable) that I should use?

bgruening:
The main Galaxy Docker image is using SLURM so this should be enough. For your usecase anything that is not `local` should work :)

gregvonkuster:
@bgruening Giving this a try, but not able to get it to work.  Can you see if there is something wrong with my command?
```
$ planemo test --docker_sudo --conda_dependency_resolution --engine docker_galaxy --docker_galaxy_image 1d8b498de930 imagej2_adjust_threshold_binary.xml
Enable beta testing mode to test artifact that isn't a Galaxy tool.
  File "/usr/bin/planemo", line 11, in <module>
    sys.exit(planemo())
  File "/usr/lib/python2.7/site-packages/click/core.py", line 716, in __call__
    return self.main(*args, **kwargs)
  File "/usr/lib/python2.7/site-packages/click/core.py", line 696, in main
    rv = self.invoke(ctx)
  File "/usr/lib/python2.7/site-packages/click/core.py", line 1060, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/usr/lib/python2.7/site-packages/click/core.py", line 889, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/usr/lib/python2.7/site-packages/click/core.py", line 534, in invoke
    return callback(*args, **kwargs)
  File "/usr/lib/python2.7/site-packages/click/decorators.py", line 64, in new_func
    return ctx.invoke(f, obj, *args[1:], **kwargs)
  File "/usr/lib/python2.7/site-packages/click/core.py", line 534, in invoke
    return callback(*args, **kwargs)
  File "/usr/lib/python2.7/site-packages/planemo/cli.py", line 174, in handle_profile_options
    return f(*args, **kwds)
  File "/usr/lib/python2.7/site-packages/planemo/commands/cmd_test.py", line 72, in cli
    with engine_context(ctx, **kwds) as engine:
  File "/usr/lib64/python2.7/contextlib.py", line 17, in __enter__
    return self.gen.next()
  File "/usr/lib/python2.7/site-packages/planemo/engine/factory.py", line 41, in engine_context
    engine = build_engine(ctx, **kwds)
  File "/usr/lib/python2.7/site-packages/planemo/engine/factory.py", line 28, in build_engine
    raise Exception(UNKNOWN_ENGINE_TYPE_MESSAGE % engine_type_str)
Exception: Unknown engine type specified [docker_galaxy].
```
I also noticed this in the help text:
```
  --docker_galaxy_image TEXT      Docker image identifier for docker-galaxy-
                                  flavor used if engine type is specified as
                                  ``docker-galaxy``. Defaults to to bgruening
                                  /galaxy-stable.
```
But docker-galaxy doesn't work either:
```
$ planemo test --docker_sudo --conda_dependency_resolution --engine docker-galaxy --docker_galaxy_image 1d8b498de930 imagej2_adjust_threshold_binary.xml
Usage: planemo test [OPTIONS] TOOL_PATH

Error: Invalid value for "--engine": invalid choice: docker-galaxy. (choose from galaxy, docker_galaxy, cwltool)
```

bgruening:
@gregvonkuster this does work for you locally?

bgruening:
Thanks @gregvonkuster!

gregvonkuster:
@bgruening Yes, all of these tests pass for me locally with planemo using a Galaxy environment with the small patch applied to the local job runner.

bgruening:
Awesome!

PR #437, opened 2016-11-22
Pipmir

dyusuf:


tools/rna_tools/pipmir/pipmir.xml
```diff
@@ -0,0 +1,230 @@
...
+
+    samtools sort
+
+    ## output in BAM format
+    -O BAM
+
+    ## output file name
+    -o test_sorted.bam
+
+    "$input_bam"
```
bgruening: single quotes are generally more recommended

tools/rna_tools/pipmir/pipmir.xml
```diff
@@ -0,0 +1,230 @@
...
+
+    &&
+    samtools index test_sorted.bam test_sorted.bai
+
+    ## the tool requires the location of RNAfold binary
+    &&
+    RNAfold_location=\$(which RNAfold)
```
bgruening: Oo! ;)

tools/rna_tools/pipmir/pipmir.xml
```diff
@@ -0,0 +1,230 @@
...
+                the reads were aligned.">
+                <option value="2bit" selected="True">
+                    Use a built-in 2bit genome</option>
+                <option value="history">
+                    Use a 2bit genome from my current history</option>
+            </param>
+            <when value="2bit">
+            <param name="builtin" type="select"
+                label="Select a reference genome">
+                <options from_data_table="2bit_genome">
```
bgruening: Can you use this one: https://github.com/galaxyproject/tools-iuc/blob/596a23abc0c657eff218d16df685b427ae7cdd0a/tools/extract_genomic_dna/extract_genomic_dna.xml#L56

With the corresponding *.loc file and the data_table ... this way both tools and many more are sharing the twobit indices.

tools/rna_tools/pipmir/pipmir.xml
```diff
@@ -0,0 +1,230 @@
...
+                    <filter type="sort_by" column="2"/>
+                    <validator type="no_options"
+                    message="No 2bit genomes are available for
+                    the selected input dataset"/>
+                </options>
+            </param>
+            </when>
+            <when value="history">
+                <param name="ownFile" type="data" format="twobit"
```
bgruening: I would prefer a fasta file here, which in converted to a twobit file. 

See here: https://github.com/galaxyproject/tools-iuc/blob/596a23abc0c657eff218d16df685b427ae7cdd0a/tools/extract_genomic_dna/extract_genomic_dna.xml#L63

tools/rna_tools/pipmir/pipmir.xml
```diff
@@ -0,0 +1,230 @@
...
+                label="Optional parameters"
+                help="You can use the default settings or
+                set custom values for any of pipmir's parameters.">
+              <option value="default">Use defaults</option>
+              <option value="custom">Full parameter list</option>
+            </param>
+            <when value="default" />
+            <!-- Full/advanced params. -->
+            <when value="custom">
+                <param name="min_precursor" type="integer"
```
bgruening: Do you have some min/max values of these integers? I guess 0 is a min value for all?

tools/rna_tools/pipmir/pipmir.xml
```diff
@@ -0,0 +1,230 @@
...
+        label="${tool.name} on ${on_string}: predicted miRNAs"/>
+    </outputs>
+    <tests>
+        <test>
+            <param name="input_bam" value="Aligned.out.sam" ftype="sam" />
+            <param name="genomeSource" value="history" />
+            <param name="ownFile" value="test_seq.2bit" />
+            <param name="settingsType" value="custom" />
+            <param name="step_size" value="10" />
+            <output name="predicted_miRNA" file="test_predicted_miRNAs.txt"
```
bgruening: Where are the two other output files?

tools/rna_tools/pipmir/pipmir.xml
```diff
@@ -0,0 +1,230 @@
...
+A text file of predicted precursors
+ * Form: Chromosome,Strand,Precursor_Start,Precursor_End,Sequence,Fold_Structure,Normalized_Minimum_Free_Energy,Mature_miRNA_Location
+
+ * Chromosome = chromosome that the precursor is on
+
+ * Strand = strand that the precursor is on
+
+ * Precursor_Start = start nucleotide (1 based) of the precursor
+
+ * Precursor_End = last nucleotide (1 based, inclusive) of the precursor
```
bgruening: I like this help!!!

tools/rna_tools/pipmir/tool_data_table_conf.xml.sample
```diff
@@ -0,0 +1,7 @@
+<!-- Use the file tool_data_table_conf.xml.oldlocstyle if you don't want to update your loc files as changed in revision 4550:535d276c92bc-->
```
bgruening: https://github.com/galaxyproject/tools-iuc/blob/596a23abc0c657eff218d16df685b427ae7cdd0a/tools/extract_genomic_dna/tool_data_table_conf.xml.sample

tools/rna_tools/pipmir/tool-data/2bit_genome.loc.sample
```diff
@@ -0,0 +1,10 @@
+#This is a sample file distributed with Galaxy that enables tools
+#to use a directory of 2bit genome files. You will
+#need to create these data files and then create a 2bit_genome.loc
```
bgruening: https://github.com/galaxyproject/tools-iuc/blob/596a23abc0c657eff218d16df685b427ae7cdd0a/tools/extract_genomic_dna/twobit.loc.sample

tools/rna_tools/pipmir/pipmir.xml
dyusuf: is the conversion automatic? I see fatotwobit is the dependency. but I don't find when it is used.

tools/rna_tools/pipmir/pipmir.xml
dyusuf: it was explicitly specified by the tool. but, it should be the usual length of plant miRNA precursor. ranging from ~50 to a couple of hundred.

tools/rna_tools/pipmir/pipmir.xml
dyusuf: The output depends on the other two which are the upstream outputs. It means, if this one is identical, the other two have to be identical as well.

tools/rna_tools/pipmir/tool_data_table_conf.xml.sample
dyusuf: ok

tools/rna_tools/pipmir/tool-data/2bit_genome.loc.sample
dyusuf: thanks, yes, this is more relevant.

tools/rna_tools/pipmir/pipmir.xml
dyusuf: but it was not practiced here,  https://github.com/galaxyproject/tools-iuc/blob/596a23abc0c657eff218d16df685b427ae7cdd0a/tools/extract_genomic_dna/extract_genomic_dna.xml#L15

but changed to signle quote any way.

tools/rna_tools/pipmir/tool_data_table_conf.xml.sample
dyusuf: <columns>dbkey, value</columns> vs <columns>value, dbkey, name, path</columns>

which is one is correct? the latter one is in the current version of pipmir. the former one is in extract_genomic_dna.

tools/rna_tools/pipmir/pipmir.xml
dyusuf: Now I see, it is not galaxy that does the conversion. it is the tool (the python script) itself does it.  

https://github.com/galaxyproject/tools-iuc/blob/596a23abc0c657eff218d16df685b427ae7cdd0a/tools/extract_genomic_dna/extract_genomic_dna.py#L36

PIPmiR itself does not have the corresponding conversion function. if fasta is preferred, the only solutions is to add conversion in the command section of the galaxy tool.

dyusuf: thanks for the update. so may be, for now, best to add a conversion in the command section?

tools/rna_tools/pipmir/pipmir.xml
dyusuf: I use min="1" for all integers.

tools/rna_tools/pipmir/pipmir.xml
bgruening: Yeah, we need to change these tools as well :(

tools/rna_tools/pipmir/pipmir.xml
bgruening: Yes, please.

tools/rna_tools/pipmir/pipmir.xml
bgruening: Can you include the other as well, if they are small ... I have some evil plans how we can make use of these test-data as part of trainings and Galaxy tours.

tools/rna_tools/pipmir/tool_data_table_conf.xml.sample
dyusuf: This two vs four columns seems to be matter. so far. I am not successful with the two columns.

dyusuf: I think I stick to the four-column version for now since it works on our server.

tools/rna_tools/pipmir/tool-data/twobit_update.loc.sample
```diff
@@ -0,0 +1,23 @@
+#This is a sample file distributed with Galaxy that is used by some
```
bgruening: Any reason you call it twobit_update and not twobit ... it should be exactly the same as the other tools that are using twobit files. Only then you can all use the same data_manager.

bgruening:
I think this is ready to go with when you adopt the twobit.loc changes.

tools/rna_tools/pipmir/tool-data/twobit_update.loc.sample
dyusuf: twobit.loc is from 2007, two column format, r u sure it is still valid format. I tested on our sytem, it did not work

dyusuf: now I update twobit.loc with 4 columns so I think it may be good to use a new name to differentiate, or...?

bgruening: But you are right, there is something fishy ... I'm pretty sure this is working, so maybe you want to copy this: https://github.com/fidelram/deepTools/blob/fd4e8dc896c2ac6f838e35e8160046d6ac862692/galaxy/wrapper/deepTools_macros.xml#L530

dyusuf: thanks. lastz_seqs.loc.sample has three columns, though. I just wonder what the standards applied when these location files were prepared. for 2bit, there are already 2-column versus 3-cloumn format. 

2-column: <Build>	<FullPathToFile>
3-column: <unique_build_id>	<display_name>		<file_path>

I made a 4-column version which worked on our server: <unique_build_id>   <dbkey>   <display_name>   <file_path>

it seems to me there is no rule to govern the preparation of loc file.

dyusuf:
@bgruening the new update uses lastz_seqs. it now works.

bgruening:
Awesome stuff!!!

dyusuf:
@bgruening thanks a lot for the compliments. :) but feel a bit overated.

bgruening:
No way ... I really like where all this is going for RNA-bioinformatics!!!

PR #438, opened 2016-11-23
WIP testing new tool with Travis

ramezrawas:
Need to know why the test is failing, "Failed to  find identifier ....."


tools/unique_regions/unique.xml
```diff
@@ -0,0 +1,46 @@
...
+        -i $infile
+        -o output_dir
+    ]]></command>
+
+    <inputs>
+        <param type="data" name="infile" format="bed" />
+    </inputs>
+
+    <outputs>
+        <!-- <data name="bed_file" type="bed">
```
bgruening: This can be removed

tools/unique_regions/unique.xml
```diff
@@ -0,0 +1,46 @@
...
+                <element name="unique_regions_1" file="out/unique_regions_1.bed" ftype="bed" />
+                <element name="unique_regions_2" file="out/unique_regions_2.bed" ftype="bed" />
+                <element name="unique_regions_3" file="out/unique_regions_3.bed" ftype="bed" />
+                <element name="unique_regions_4" file="out/unique_regions_4.bed" ftype="bed" />
+            </output_collection>
+        </test>
+    </tests>
+
+    <help><![CDATA[
+**Unique Reads**
```
bgruening: Can you write a little bit more about your tool here? What it does, what are the inputs, what it creates ... BED files with non-overlapping regions.

bgruening:
You need to update the test data as well.

tools/unique_regions/unique_regions.py
```diff
@@ -0,0 +1,85 @@
...
+def overlap(infile):
+    overlapping = ""
+    unique = ""
+    prev_chromo = ""
+    prev_start = 0
+    prev_end = 0
+
+    #iterating the bed file, line by line
+    for line in open(infile, 'r'):
+        split_line = line[:-1].split("\t")
```
bgruening: This should not be needed `line[:-1]` just use `line.split()`

tools/unique_regions/unique_regions.py
```diff
@@ -0,0 +1,85 @@
...
+        chromo = str(split_line[0])
+        start = int(split_line[1])
+        end = int(split_line[2])
+
+        #verifying whether the two successive regions are on the same chromosome,
+        #if not, the comparison will be skipped
+        if chromo != prev_chromo:
+            prev_chromo = chromo
```
bgruening: These blocks are the same/similar than the blocks below. Maybe you can put this in a small helper function to reduce code and DRY.

tools/unique_regions/unique_regions.py
```diff
@@ -0,0 +1,85 @@
...
+    unique_regions = "%s\t%i\t%i\n" % (chromo, start, end)
+    unique += unique_regions
+    unique = remove_first_line(unique)
+    return unique, overlapping
+
+def unique_regions(args):
+    l = 0
+    #retrieving the two output groups (unique and overlapping regions)
+    unique, overlapping = overlap(args.input)
+    unover_file_name = "unique_regions_"+str(l)+".bed"
```
bgruening: use %s or other string substitutions ....

tools/unique_regions/unique_regions.py
```diff
@@ -0,0 +1,85 @@
...
+        l += 1
+        args.input = "temp_overlapping_regions.bed"
+        unique, overlapping = overlap(args.input)
+
+        unover_file_name = "unique_regions_"+str(l)+".bed"
+        unover_file_name = os.path.join(args.output, unover_file_name)
+        with open(unover_file_name, 'w') as f1:
+            f1.write(unique)
+        with open("temp_overlapping_regions.bed", 'w') as f2:
+            f2.write(overlapping)
```
bgruening: So when overlapping is full you write something and than the while loop will continue, correct?
If `overlapping` is empty the while loop will terminate ... you could simply ask `if not overlapping` and either set something to False or terminate the while loop directly without the need to `.st_size`

ramezrawas: Done :)

tools/unique_regions/.shed.yml
```diff
@@ -0,0 +1,17 @@
+categories:
+- BedTools
+- RNA
```
bgruening: It should be "Text Manipulation", remove the others please.

tools/unique_regions/test-data/out/unique_regions_2.bed
```diff
@@ -0,0 +1,2 @@
+
```
bgruening: Here is something wrong.

tools/unique_regions/unique.xml
```diff
@@ -0,0 +1,46 @@
...
+    </requirements>
+
+    <stdio>
+        <exit_code range="1:" />
+    </stdio>
+
+    <command><![CDATA[
+      mkdir output_dir
+      &&
+      python $__tool_directory__/unique_regions.py
```
bgruening: ping

tools/unique_regions/unique_regions.py
```diff
@@ -0,0 +1,80 @@
+#!/usr/bin/env python
+
+import os
+import argparse
+
+def remove_first_line(s):
```
bgruening: This function is not used.

tools/unique_regions/unique_regions.py
bgruening: Cool!

tools/unique_regions/unique_regions.py
ramezrawas: it is used in line 43:
unique = remove_first_line(unique)

tools/unique_regions/unique_regions.py
```diff
@@ -0,0 +1,80 @@
...
+
+    #writing the last region
+    unique += build_cont(prev_chromo, prev_start, prev_end)
+    unique = remove_first_line(unique)
+    return unique, overlapping
+
+def unique_regions(args):
+    l = 0
+    #retrieving the two output groups (unique and overlapping regions)
+    unique, overlapping = overlap(args.input)
```
ramezrawas: it can be integrated in the while loop as follows:
    while l == 0 or overlapping != "":
        if l == 0:
            args.input = XXXX
        else:
            args.input = "temp_overlapping_regions.bed"
How would I replace XXXX to read from the original file??

PR #446, opened 2016-11-28
osra

HassanAmr:
fixed linting for osra


bgruening:
Can you please add to those tools a real citation, where it makes sense?

PR #450, opened 2016-11-28
strip-it

HassanAmr:
fixed linting errors for strip-it


HassanAmr:
Hi @bgruening,
I'm a bit confused, why would some tools fail linting when the citation tag is missing, and some like this one and shape-it succeed while missing the citation tag?

PR #457, opened 2016-12-04
add tool mlocarna

s-will:
* most of the core functionality of mlocarna based on locarna,
  locarna-p and sparse
* input: fasta or 'text' (i.e. 'locarna-extended fasta')
* output: alignments in clustal and stockholm format
* supports global and local alignment, as well as probabilistic
  alignment
* known bugs: stockholm output fails in probabilistic mode with
  locarna 1.8.11 (this will be fixed not before locarna 1.8.13)


s-will:
Hi Bjoern, this is -finally- a hand-written wrapper for mlocarna (locarna, locarna-p, sparse). Most functionality is there, only some more exotic stuff is missing. I hope this works as a start :).
Definitely this needs an requirement update to locarna-1.8.12 or better -13 (soon to come)

bgruening:
Amazing! This is great and much needed, thanks a lot!
I will review tomorrow. Please include a `.shed.yml` file, this will trigger a test with planemo on travis and enable an automatic TS push.

Thanks again!
Bjoern

tools/rna_tools/mlocarna/mlocarna.xml
```diff
@@ -0,0 +1,167 @@
...
+    <expand macro="requirements" />
+
+    <expand macro="stdio" />
+    
+    <expand macro="version" />
+    
+    <command><![CDATA[
+    mlocarna
+    
+    $input_data
```
bgruening: All type=data and type=text should be single quoted.

tools/rna_tools/mlocarna/mlocarna.xml
```diff
@@ -0,0 +1,167 @@
...
+    --stockholm
+    --tgtdir outdir
```
bgruening: single quoted

tools/rna_tools/mlocarna/macros.xml
```diff
@@ -0,0 +1,102 @@
...
+        <param name="indel_opening" label="Indel Opening Score" type="integer"
+               value="-500" help="--indel-opening" />
+        <param name="indel" label="Indel score" type="integer" value="-350" help="--indel" />
+        <param name="tau" type="integer" value="50" 
+               label="Sequence contribution at structure match in percent"
+               help="--tau factor"/> 
+
+        <conditional name="sequence_score">
+            <param name="sequence_score_selector" type="select" label="Type of sequence score contribution">
+                <option value="ribofit">Use ribofit</option>
```
bgruening: You could use as value="--ribofit true" and save a little bit of code above in the command section, not sure it's worth the effort and maybe it makes the code not so readable.

tools/rna_tools/mlocarna/macros.xml
```diff
@@ -0,0 +1,102 @@
...
+        
+        <param name="rnafold_temperature" type="float" value="37.0"
+               label="Temperature for RNAfold (RNAfold's -T option)"
+               help="--rnafold-temperature=temp" />
+    </xml>
+
+
+
+    <xml name="common_heuristic_parameters">
+        <param  name="min_prob" type="float" value="0.0005" label="Minimal probability" help="-p cutoff" />
```
bgruening: Do you have some `min` `max` recommendations? Including min, max will render a nice slider in the Galaxy UI.

tools/rna_tools/mlocarna/test-data/Makefile
```diff
@@ -0,0 +1,28 @@
+ALL_TESTS=archaea-default archaea-probabilistic haca.snoRNA-default
```
bgruening: Is this Makefile needed in test-data? :)

tools/rna_tools/mlocarna/mlocarna.xml
```diff
@@ -0,0 +1,167 @@
...
+    --max-diff-at-am $heuristics.max_diff_at_am
+    --max-bps-length-ratio $heuristics.max_bps_length_ratio
+    
+    $heuristics.alifold_consensus_dp
+
+    ## -------------------- other parameters
+    
+    $other.lonely_pairs
+    
+    > $stdout
```
bgruening: quotes

tools/rna_tools/mlocarna/mlocarna.xml
```diff
@@ -0,0 +1,167 @@
+<tool id="mlocarna" name="LocARNA" version="1.8.11">
```
bgruening: Should the version be a macro, so it can be reused in the requirement tag?

tools/rna_tools/mlocarna/mlocarna.xml
```diff
@@ -0,0 +1,167 @@
...
+        <test>
+            <param name="input_data" value="archaea.fa" />
+            <param name="alignment_mode_selector" value="probabilistic" />
+            <output name="stdout" file="archaea-probabilistic.stdout" />
+        </test>
+    </tests>
+
+    <help><![CDATA[ **Multiple alignment of RNAs**
+    
+For more information, see
```
bgruening: I would prefer some more help directly in the tool, for people without internet access.
But we can do this later.

tools/rna_tools/mlocarna/macros.xml
```diff
@@ -0,0 +1,102 @@
...
+    </xml>
+    
+    <xml name="common_scoring_parameters">
+        <param name="struct_weight" label="Structure Weight" type="integer" 
+               value="200" help="--struct-weight" />
+        <param name="indel_opening" label="Indel Opening Score" type="integer"
+               value="-500" help="--indel-opening" />
+        <param name="indel" label="Indel score" type="integer" value="-350" help="--indel" />
+        <param name="tau" type="integer" value="50" 
+               label="Sequence contribution at structure match in percent"
```
s-will: this is certainly very useful, thanks:)

tools/rna_tools/mlocarna/test-data/Makefile
s-will: is there a standard way to generate the test-output automatically? What if default parameter values change?
Currently, I generate the test output with this Makefile -- so, for me it is at least useful :)

bgruening: `planemo test --conda_dependency_resolution --update_test_data  ./path/to/tool/` generates you test data depending on your test-cases. It is very recommended to check the output carefully afterwards :)

tools/rna_tools/mlocarna/macros.xml
```diff
@@ -4,6 +4,9 @@
             <requirement type="package" version="1.8.11">locarna</requirement>
```
bgruening: You can use @VERSION@ here as well.

bgruening:
If you add .shed.yml we will see if this is working here :)
Thanks!

bgruening:
last test failed: https://travis-ci.org/bgruening/galaxytools/jobs/181278888#L2577
Other looks good!

bgruening:
Seems that the last test does not produce the exact same output?
https://travis-ci.org/bgruening/galaxytools/jobs/181347155#L2544

s-will:
Yup, the difference is quite small -- in practice negligible, nevertheless this is disconcerning given in both cases the exact same binaries. 
I am still puzzled -- maybe you have ideas for debugging?:
Is there a chance to get (intermediary) result files from the travis run? If I have time, I plan to run tests in a Precise environment. Btw,  how does planemo test select the binaries; would this take them from the path or does it use conda (and thus ensure the correct versions itself)?

bgruening:
If you use `planemo test --conda_dependency_resolution` it will take the conda package. This is what travis is doing. On travis there is currently no way to get the intermediate results, we need to change the travis setup to make this happen.

But I guess you can easily reproduce this error on Jörgs computer, isn't it? It seems to be a problem with some randomness?

Nevertheless, if this is expected you can use `lines_diff` and say both files needs to be identical, but 6 lines ...

bgruening:
Yeah!!!!
Danke!

s-will:
Hi @bgruening, I am still puzzled what causes the small differences. I could reproduce my results in a Precise-VM and on a different computer running a recent Fedora; i.e. I never observed the deviations seen on Travis. Thus, I decided to postpone the problem and just modified the 3rd check to compare only the alignments. Let's see what happens when this is running on a Galaxy server -- maybe the problems will magically vanish :).

Next, I will upgrade locarna and then write a new version of the wrapper. (This is needed to fix a bug concerning the combination of stockholm output and probabilistic alignment.)

Then, I want to add a few more related tools....

bgruening:
Sounds great! I will work on the workbench this week and get all tools installed on the FR server.

PR #459, opened 2016-12-09
Remove sklearn

eteriSokhoyan:
tool for GraphClust, without sklearn, to test if GraphClust tools are passing.


eteriSokhoyan:
this one can be ignored. it was just for test.

PR #466, opened 2016-12-10
Uniq testing

bgruening:
A simple test case is needed.

ping @HassanAmr


HassanAmr:
I'm on it

nsoranzo:
Isn't this equivalent to https://github.com/bgruening/galaxytools/blob/master/tools/text_processing/text_processing/unsorted_uniq.xml ?

bgruening:
Yes, just the very old version. I guess after this conda migration, I will deprecate this. Or do you think its ok to deprecate it now and remove it from the github.

HassanAmr:
I'm also working on fixing the linting errors in text-processing. Should I wait on this one until you decide?

bgruening:
It has for sure very low priority. I'm fine with deprecating it now, just wanted to be a good maintainer - at least one time :)

HassanAmr:
How do I push to this pull request?

nsoranzo:
If it has no future, I'd the deprecate it now. The sooner the better, in my opinion, no point in releasing a new version and then deprecate it. But it's your package, so your choice of course!

HassanAmr:
linting awk.xml locally doesn't produce any errors, can we retry the test please?

HassanAmr:
I commented in the wrong thread, sorry.

PR #469, opened 2016-12-11
Text processing

HassanAmr:


HassanAmr:
linting awk.xml locally doesn't produce any errors, can we retry the test please?

bgruening:
Seems to be a legit failure on the cut tool.

```
```
ERROR: Invalid tmpZ_eDCM found. Errors [/tmp/tmpZ_eDCM:72:0:ERROR:SCHEMASV:SCHEMAV_ELEMENT_CONTENT: Element 'conditional': This element is not expected. Expected is ( action ).]
```

HassanAmr:
Yes, cut.xml has a problem. I mean awk.xml

bgruening:
Oh I see, it has a URL that is linted, so planemo is checking if all URLs are valid and it seems this line : http://www.ibm.com/developerworks/linux/library/l-awk1.html is not valid :)

bgruening:
Can you try this one: http://www.ibm.com/developerworks/linux/library/l-awk1/index.html

HassanAmr:
It worked, thanks :)

PR #474, opened 2016-12-15
Segemehl Tests added

jfallmann:
Added segemehl tests, Travis builds, local planemo tests not, probably due to some misconfiguration.
Fixed typo in tool options setting hardclip with split option


bgruening:
@jfallmann thanks! There is one lint'ing error: https://travis-ci.org/bgruening/galaxytools/jobs/184282892#L807

jfallmann:
Hmm, but this is also used in 
https://github.com/galaxyproject/tools-devteam/blob/master/tools/bowtie2/bowtie2_wrapper.xml
I was just adding test data and the test section, but if this is a problem that should be addressed,
what's the right way to do it?

bgruening:
@jfallmann do you understand the latest error? Do segemehl need a full path to the index_dir?

jfallmann:
@bgruening I think it's necessary to name the idx,

file: libs/sufarray/sufarray.c, line: 220: Couldn't open file ./temp_index/. Exit forced

This seems as if segemehl is using the $temp_idx which is set to './temp_index/' as name for the index and not generating an index file in the dir $temp_idx, maybe we need $temp_idx_dir and $temp_idx?

jfallmann:
@bgruening  If I try this 

mkdir tmp_idx
export tmp='./tmp_idx/'
segemehl.x -x $tmp -d chr1.fa

on my workstation, I get

file: libs/sufarray/sufarray.c, line: 220: Couldn't open file ./tmp_idx/. Exit forced

if I do
export tmp='./tmp_idx/tmp.idx'
and run the same line, I get 

writing suffix array './tmp_idx/tmp.idx' to disk.
[SEGEMEHL] Fri Dec 16 10:44:51 2016: 
Goodbye.

tools/segemehl/segemehl.xml
```diff
@@ -14,7 +14,7 @@
         ## prepare segemehl index if no reference genome is supplied
         #if $refGenomeSource.genomeSource == "history":
             mkdir ./temp_index/ &&
-	    #set $temp_index = './temp_index/'
+	    #set $temp_index = './temp_index/$refGenomeSource.own_reference_genome'+'.idx'
```
bgruening: I think we can just use some temp name like `./temp_index/tmp.idx` or is the name really interesting. `$refGenomeSource.own_reference_genome` does probably not work, as this is the entire path to a file.

bgruening:
Yeah!!!

bgruening:
You don't managed to get planemo running locally? If you did and have still some energy a second test for non-single-end libraries would be great.

jfallmann:
@bgruening Travis says
All 1 test(s) executed passed.
segemehl[0]: passed

Steve promised to have the new version soonish, I can make the test for paired end but will have to update everything anyway, if I'm able to run planemo I'll give it another try

PR #478, opened 2016-12-22
GraphClust wrapper fix if orig-head not available

mmiladi:


mmiladi:
@bgruening  There was a bug in reading fasta format. This PR should have fixed it.

PR #479, opened 2016-12-23
Graphclust update

mmiladi:
This PR adds feature to report clustering quality metrics if input data is labeled.


tools/GraphClust/CollectResults/evaluation.py
```diff
@@ -46,3 +47,31 @@ def sh(script):
     toWrite += listOfClasses[i] + "\t" + listOfClusters[i] + '\n'
 with open("RESULTS/fullTab.tabular", "w") as full:
     full.write(toWrite)
+
+
+listOfClasses = []
+listOfClusters = []
+pattern = re.compile("^RF.*$")
+
+with open("RESULTS/fullTab.tabular", 'r') as f:
```
bgruening: Why are you opening the same file twice? Could you not put the if condition into with and check the first line with your regex?

mmiladi: @eteriSokhoyan It seems one can check the pattern on first element of listOfClasses and not open the file second time. Do yo agree?

eteriSokhoyan: Yes sure, it's possible

mmiladi: @bgruening @eteriSokhoyan  Below commit should fix this.

bgruening:
Thanks!

PR #481, opened 2016-12-23
Change to imgteam

bgruening:
ping @gregvonkuster @ThomasWollmann 

This will kickstart a Galaxy-Imaging community :)

### DIFF
--- a/tools/image_processing/imagej2/.shed.yml
+++ b/tools/image_processing/imagej2/.shed.yml
@@ -1,5 +1,5 @@
 name: imagej2
-owner: iuc
+owner: imgteam
 description: |
   ImageJ2 is a new version of ImageJ for the next generation of multidimensional
   image data, with a focus on scientific imaging.


bgruening:
@gregvonkuster we have a new team dedicated from image processing in the TS.
Do you agree to use this from now on?
I would also like to push these tools to the MTS under the new namespace.

gregvonkuster:
Sure - this is great!  There is a slight enhancement needed for these tools that I have been trying to find time to do, but haven't yet.  :(

This file: https://github.com/bgruening/galaxytools/blob/master/tools/image_processing/imagej2/imagej2_base_utils.py
included the following:
```
FIJI_JAR_DIR = os.environ.get('FIJI_JAR_DIR', None)
FIJI_OSX_JAVA3D_DIR = os.environ.get('FIJI_OSX_JAVA3D_DIR', None)
FIJI_PLUGIN_DIR = os.environ.get('FIJI_PLUGIN_DIR', None)
FIJI_ROOT_DIR = os.environ.get('FIJI_ROOT_DIR', None)
```
This needs to be fixed so that the environment has these values at tool execution time.  In the past, we set these in an external env.sh file that was sourced by the job.  But I think we have better approaches for this now.  This should be fixed before we upload these latest version to the TS.

I will fix this, but want to make sure of the best approach.  Should I do something similar to this: https://github.com/bgruening/galaxytools/pull/490/commits/227686754daa2c515679ebc3e8a29418b5b7f073

bgruening:
hehe, you have seen this nice trick ;) What do you think about it?
I like it and we can do this as well here, yes!
Thanks!

gregvonkuster:
@bgruening It is not readily clear to me how this https://github.com/bgruening/galaxytools/pull/490/commits/227686754daa2c515679ebc3e8a29418b5b7f073 works.  I cannot find an example tool that uses this trick to set environment variables like I need.  Can you clarify how this works for me?

gregvonkuster:
@bgruening it turns out the fix for this is much easier than I expected.  Making the installation via conda (which I did a while back for version 20151222) completely eliminates the need for these environment variables (I believe) since they are now defined here:  https://github.com/gregvonkuster/bioconda-recipes/blob/master/recipes/fiji/20151222/bunwarpj.sh.

I'll submit a PR eliminating the use of these environment variables.

bgruening:
As travis can not test these wrappers ... can I merge, have you tested them? :) Thanks.

PR #482, opened 2016-12-24
Samtools sort

HassanAmr:
- Applied modifications from @bgruening 
- Removed tool from blacklist
- Added citation


HassanAmr:
Hi @bgruening,

There is another citation I should add as per this line from the old xml tool:
`If you use this tool in Galaxy, please cite Blankenberg D, et al. *In preparation.*`

I tried searching for it (since it has been 2 years), but I can't figure out which article is should be.

bgruening:
thats fine, this was never published.

PR #484, opened 2016-12-26
mlocarna

s-will:
update to LocARNA 1.9.0
add reliability-profile script


s-will:
sorry, I pushed this too early; the new tool reliability-profile needs some fixes

s-will:
Hi @bgruening; the last one was hard to find, since archive files are generally ignored in galaxytools (.gitignore). Are you sure, this is a good idea? Is this necessary -- maybe for such purposes, rather .git/info/exclude should be used.

bgruening:
Let's remove this. It has mostly historical reasons ... At this time planemo was not available and we created tarballs to upload tools to the toolshed. We can remove it today I think.
Sorry to have wasted your time.

s-will:
no worries

tools/rna_tools/locarna/locarna.xml
```diff
@@ -1,92 +0,0 @@
-<tool id="locarna" name="locarna" version="0.1.0">
-    <description></description>
-    <requirements>
-    <requirement type="package" version="x.y.z">locarna</requirement>
```
s-will: Hi @bgruening, I don't think that this was ever working properly -- also I don't find it in the toolshed. The old wrapper would be completely replaced by the new locarna_multiple.

I don't completely see the consequences of removing the old locarna. If you think we should rather not remove "locarna", I am fine with deprecating it (than we can add a reference to the new wrapper).
(Interestingly, there is a mlocarna wrapper in graphclust as well -- I didn't realize this so far.)

s-will:
Hi @bgruening,

grade hab ich noch ein paarweises locarna alignment tool geaddet -- damit wird vielleicht klarer, warum es mir sinnvoll scheint, diese tools mit locarna_ prefix zu benennen. Es wird vermutlich noch ein locarna_pairwise_p und ein exparna_p dazukommen (Namen so oder so aehnlich, evtl 'locarna_exparna_p')

Denkst du das passt so oder waers besser das zu aendern? [Geht das zusammen mit der Idee das alte "locarna" beizubehalten und zu deprecaten? oder brauchen wir das vielleicht eh nicht...]

tools/rna_tools/locarna/locarna.xml
bgruening: regarding the mlocarna wrapper of graphclust ... can you please talk to @mmiladi about this.

tools/rna_tools/locarna/locarna_pairwise.xml
```diff
@@ -0,0 +1,284 @@
...
+    #if str($stdout_verbosity) != "--quiet":
+        > '$stdout'
+    #end if
+
+    #if 'clustal' in str($outputs).split(",") and 'clustal_strict' in str($outputs).split(",")
+        && cp $clustal $clustal_strict
+    #end if
+
+    #if 'clustal_strict' in str($outputs).split(",")
+        && sed -i '/\^#/d' '$clustal_strict'
```
bgruening: This will not work on OSX. `-i` needs to be replaces by `-i.bak` or you can strictly depend on `sed`

tools/rna_tools/locarna/locarna_reliability_profile.xml
```diff
@@ -0,0 +1,196 @@
...
+    #else if $output_format == 'png'
+         --out '$png_output_plot'
+    #end if
+
+    #if str($title) != ""
+        --title '$title'
+    #end if
+
+    #if str($seqname) != ""
+        --seqname $seqname
```
bgruening: quotes

bgruening:
I'm all in favor of a suite, especially if we can replace the graphclust one as well.

bgruening:
Will merge now, the comments can be fixed in a follow up PR. Thanks.

tools/rna_tools/locarna/locarna.xml
mmiladi: Hi @s-will, historically I am not precisely knowledgeable about the details of *special* way graphclust invokes locarna. The locarna_graphclust wrapper is a wrapper that invokes locarna reliability script and further post-processes the mlocarna tree to get best subtree and because of that I asked @eteriSokhoyan to append the name graphclust to avoid confusion. Do you think it would be possible to merge locarna_graphclust   with locarna_multiple wrapper?

PR #485, opened 2016-12-31
Add CD-HIT plus fix evaluation

mmiladi:


tools/GraphClust/CollectResults/addCdhitseqs.py
```diff
@@ -0,0 +1,57 @@
...
+
+#clusterFiles = clusters.split(',')
+repSeqRedSeqdict = {}
+repLine = ""
+count = 0
+first = False
+
+with open(cdhitcluster, 'r+') as f:
+    lines = f.readlines()
+    for i in range(0, len(lines)):
```
bgruening: This is very un-pythonic.
You usually use:
```
for i,line in enumerate(lines):
```

tools/GraphClust/CollectResults/addCdhitseqs.py
```diff
@@ -0,0 +1,57 @@
...
+            #lineArr.append(add_short_id)
+            #repSeqRedSeqdict[rep_short_id] = lineArr
+
+toWrite = ""
+for singleFile in sorted(cluster_seqs_stats_files):
+    for key, val in repSeqRedSeqdict.items():
+        if key in open(singleFile).read():
+            line = open(singleFile).readline()
+            for i in val:
+                toWrite = line.split()[0] + "  " + line.split()[1] + "  " + line.split()[2] + "  " + " - " + "   " + "CD-Hit" + "    " + line.split()[5] + "  " + "ORIGID" + "  "  + str(i) + "\n"
```
bgruening: why space separated and not tab saparated?

tools/GraphClust/CollectResults/addCdhitseqs.py
mmiladi: Agree

mmiladi: @eteriSokhoyan Hint for future projects ;-)

tools/GraphClust/CollectResults/addCdhitseqs.py
```diff
@@ -0,0 +1,57 @@
...
+            #add_short_id = re.findall("_".join(["[^_]+"] * 2), add_FullId)[0]
+            lineArr.append(add_FullId)
+            repSeqRedSeqdict[rep_FullId] = lineArr
+            #lineArr.append(add_short_id)
+            #repSeqRedSeqdict[rep_short_id] = lineArr
+
+toWrite = ""
+for singleFile in sorted(cluster_seqs_stats_files):
+    for key, val in repSeqRedSeqdict.items():
+        if key in open(singleFile).read():
```
bgruening: This will access the file everytime in the loop again. Just read once, and store it for later. Try to avoid disc IO.

tools/GraphClust/CollectResults/addCdhitseqs.py
mmiladi: Hmm, for historical reasons.

tools/GraphClust/CollectResults/addCdhitseqs.py
```diff
@@ -0,0 +1,57 @@
...
+            line = open(singleFile).readline()
```
bgruening: And here you access it again ... just to get the first element of the already processed/read file from above?

tools/GraphClust/CollectResults/addCdhitseqs.py
mmiladi: I agree, but since this is done only once at the end it should be ok for now..
@eteriSokhoyan Please consider this as improvement update for later.

tools/GraphClust/CollectResults/addCdhitseqs.py
mmiladi: @eteriSokhoyan Can you please look at the write loop? Do you really want to read twice and write once in the same file each iteration?
Also I guess the check for `if key in open(singleFile).read():` should be on specific column of the file.

tools/GraphClust/CollectResults/addCdhitseqs.py
bgruening: It's your decision, but it's done for every element in `repSeqRedSeqdic` which corresponds to all lines in the CDHit results, more or less ...

mmiladi: I missed the bigger loop, agree with you. This part should be re-factored, see comment below.

mmiladi:
@bgruening  Pls check Eteri's improved code

mmiladi:
Thanks @bgruening !

mmiladi:
@eteriSokhoyan There is an error with cd-hit, after loading the workflow it immediately complains.
@bgruening The latest docker tag is failing, could you please have a look to this screen shot. CD-hit argument supposed to be optionally provided.
image

bgruening:
I think this is ok. Just open the workflow, save it, export it and update the github repo with the new one. It means that this parameter is new and was not used before.

mmiladi:
Alright :-) I downloaded the workflow, is it save&export you mean?

bgruening:
Yes

bgruening:
The diff should not be that different I hope.

mmiladi:
This is the diff: https://github.com/BackofenLab/docker-galaxy-graphclust/pull/23/commits/9e24e5d24b5c3eeb081c66e48fd15801619adda5

PR #489, opened 2017-01-11
update locarna suite; add wrappers

s-will:


s-will:
Hey @bgruening,
This rather looks like some internal problem; nothing due to my wrappers, right? Could you please have a look.

bgruening:
I hope not, internal problem are bad :)

bgruening:
Indeed an internal github error as it seems. Runs now.

bgruening:
Nice! Macros for the win!

s-will:
thanks! that was quick :). Good that the internal error was not so bad after all.

PR #490, opened 2017-01-12
Use msgf_plus package from conda and make MSGFPlusAdapter work with it.

blankclemens:
Yes, a bit hacky. But: #workswithconda :dark_sunglasses: 

.. and yes, this only works when installing msgf_plus with conda, but in other cases one would have to manually edit this file anyways to set the place for MSGFPlus.jar.


bgruening:
Ha, cool! Do you remember the conda pyhton wrapper for Java stuff. What about enhancing this wrapper with an argument to return the JAR file path. This way we could save this hack here and everyone else can also get the JAR file path using conda.

blankclemens:
Good point. Fun fact: I spent quite some time rewriting some of the msgf Java wrapper in the Cheetah code, just to realize that it won't work as intended since the cheetah code is run in the main Conda environment. Well, guess I'll update the package to have it clean.

blankclemens:
Should be fixed with https://github.com/bioconda/bioconda-recipes/pull/3455.

bgruening:
Great stuff thanks!

tools/openms/MSGFPlusAdapter.xml
```diff
@@ -10,11 +10,31 @@
...
   <expand macro="stdio"/>
   <expand macro="requirements"/>
-  <command>MSGFPlusAdapter
--executable msgfplus.jar
+  <command>
+<![CDATA[
+
+## check input file type
+#if $param_in.is_of_type('mzml'):
+  #set $in_type = 'mzml'
```
bgruening: I think you can drop all of this and just use `$in_type.ext`

blankclemens: Is this implemented? Couldn't find it in wrappers.py.

bgruening: Yes :)

blankclemens: Done.

tools/openms/MSGFPlusAdapter.xml
```diff
@@ -10,11 +10,26 @@
   <expand macro="references"/>
   <expand macro="stdio"/>
   <expand macro="requirements"/>
-  <command>MSGFPlusAdapter
--executable msgfplus.jar
+  <command><![CDATA[
 
+## check input file type
+#if $param_in.is_of_type('mzml'):
+  #set $in_type = 'mzml'
```
bgruening: I still think $param_in.ext should work.

blankclemens: Oh, seems like I forgot to push, that was done already ;)

PR #492, opened 2017-01-13
Eliminate unnecessary env vars

gregvonkuster:
No longer necessary since the introduction of the 20151222 conda install recipe for Fiji here https://github.com/bioconda/bioconda-recipes/blob/master/recipes/fiji/20151222/bunwarpj.sh

### DIFF
--- a/tools/image_processing/imagej2/imagej2_base_utils.py
+++ b/tools/image_processing/imagej2/imagej2_base_utils.py
@@ -3,11 +3,6 @@ import shutil
 import sys
 import tempfile
 
-FIJI_JAR_DIR = os.environ.get('FIJI_JAR_DIR', None)
-FIJI_OSX_JAVA3D_DIR = os.environ.get('FIJI_OSX_JAVA3D_DIR', None)
-FIJI_PLUGIN_DIR = os.environ.get('FIJI_PLUGIN_DIR', None)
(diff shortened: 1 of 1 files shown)


tools/image_processing/imagej2/imagej2_base_utils.py
```diff
@@ -3,11 +3,6 @@
 import sys
 import tempfile
 
-FIJI_JAR_DIR = os.environ.get('FIJI_JAR_DIR', None)
-FIJI_OSX_JAVA3D_DIR = os.environ.get('FIJI_OSX_JAVA3D_DIR', None)
-FIJI_PLUGIN_DIR = os.environ.get('FIJI_PLUGIN_DIR', None)
```
bgruening: FIJI_OSX_JAVA3D_DIR this was never used?

gregvonkuster: No, it was never used.  I believe it was just left over from my early days writing these tools.  The Java 3D library stuff threw all kinds of exceptions on Linux, so I never used it.  It isn't easy (or maybe not even possible) in the Galaxy UI anyway.  Since the Java 3D libraries never posed problems on OS X, I think I just forgot to eliminate this unused env var.

PR #496, opened 2017-01-17
Proper handling of boolean values that are not a flag

blankclemens:
..which have "true" as their default value and require a text parameter.

### DIFF
--- a/tools/openms/CompNovo.xml
+++ b/tools/openms/CompNovo.xml
@@ -41,11 +41,6 @@
 #end if
 #if $param_algorithm_estimate_precursor_mz:
   -algorithm:estimate_precursor_mz
-  #if " " in str($param_algorithm_estimate_precursor_mz):
-    "$param_algorithm_estimate_precursor_mz"
-  #else
-    $param_algorithm_estimate_precursor_mz
-  #end if
 #end if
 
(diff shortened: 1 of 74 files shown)


tools/openms/CompNovo.xml
```diff
@@ -41,6 +41,11 @@
 #end if
 #if $param_algorithm_estimate_precursor_mz:
   -algorithm:estimate_precursor_mz
+  #if " " in str($param_algorithm_estimate_precursor_mz):
```
bgruening: The only difference between if/else are the "quotes"?
The if condition can also be never true anymore? Is this all due to the converter?
Without looking much further I think you can remove the entire if and write `-algorithm:estimate_precursor_mz $param_algorithm_estimate_precursor_mz`

blankclemens: You're right, I have this on my list.

blankclemens: Fixed.

blankclemens: Eh nope, forgot the command :)

blankclemens: Now.

bgruening:
Nice! 500 lines less code :)

PR #497, opened 2017-01-19
Fix for PeakPickerHiRes

blankclemens:
Fix for PeakPickerHiRes. Check OpenMS/OpenMS#2358 for more details.

### DIFF
--- a/tools/openms/PeakPickerHiRes.xml
+++ b/tools/openms/PeakPickerHiRes.xml
@@ -36,7 +36,7 @@
   -algorithm:report_FWHM
 #end if
 #if $param_algorithm_report_FWHM_unit:
-  -algorithm:report_FWHM_unit $param_algorithm_report_FWHM_unit
+  -algorithm:report_FWHM_unit "$param_algorithm_report_FWHM_unit"
 #end if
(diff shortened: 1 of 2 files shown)


bgruening:
Is this because of the brackets?

blankclemens:
Yes. To make it an automatic solution, what are your thoughts about quoting all parameters in general?

bgruening:
In general all type=data and type=text should be quoted, but with single-quotes.
We can not quote all parameters, this will probably fail on booleans and integers, isn't it?

blankclemens:
Single-quotes would be fine as well of course. So yes, that would be an easy solution, I just thought there might be a reason why CTD2Galaxy is not quoting these parameters at the moment - well, why quote something which is automatically defined anyways.

But since numbers should be flagged as such, quoting all text and data parameters might just do it without side effects.

bgruening:
If we really think it's worth to spend more time in this we should not create the command section at all and use <configfiles>.

blankclemens:
What exactly do you mean with "use ."?  Create an INI file for every job and load this one?

bgruening:
Galaxy has a configfile option that we can use to create this ini file very easily. But I think this is too much work.

PR #500, opened 2017-01-21
Graphclust all-in-one results file

mmiladi:


tools/GraphClust/LocARNAGraphClust/locarna_best_subtree.xml
```diff
@@ -160,23 +221,23 @@ This saves the calculation of pairwise all-vs-all similarities and construction
                   characterized.}
 }
 
-					</citation>
-					<citation type="bibtex">@Article{Smith:Heyne:Richter:Freib_RNA_Tools:NAR2010,
-  author =	 {Smith, Cameron and Heyne, Steffen and Richter, Andreas S.
+          </citation>
+          <citation type="bibtex">@Article{Smith:Heyne:Richter:Freib_RNA_Tools:NAR2010,
```
bgruening: I saw this to late, but it is way easier to just specify.
```
<citation type="doi">10.1093/nar/gkq316</citation>
```

tools/GraphClust/NSPDK/NSPDK_candidateClusters.xml
```diff
@@ -1,6 +1,6 @@
 <tool id="NSPDK_candidateClust" name="NSPDK_candidateClusters" version="9.2">
 	<requirements>
-		<requirement type="package" version="0.1">graphclust-wrappers</requirement>
```
bgruening: This could actually become a macro at some point.

tools/GraphClust/LocARNAGraphClust/locarna_best_subtree.xml
mmiladi: True, it would have been easier. For future citations we take care of using doi.

tools/GraphClust/NSPDK/NSPDK_candidateClusters.xml
mmiladi: In some cases, though rare, we may need to keep some wrappers with older version..

bgruening: Why is this so? This seems strange. But nevertheless, you can do this with `yield` in a macro.

mmiladi: What yield is used for? Thanks for merging.

PR #501, opened 2017-01-23
OpenMS 2.1 - Fix for FileConverter

blankclemens:


tools/openms/FileConverter.xml
```diff
@@ -75,7 +75,7 @@
     </expand>
   </inputs>
   <outputs>
-    <data name="param_out" metadata_source="param_in" format="input"/>
+    <data name="param_out" metadata_source="param_in" auto_format="true"/>
```
bgruening: explicit is better then implicit ;)
But let's try this!

PR #504, opened 2017-01-23
Glimmer3

HassanAmr:
Fixed linting and added test cases for glimmer2gff and gbk2orf


bgruening:
Thanks! To really test this you need to take it off the tt_blacklist.

bgruening:
These tools are now in IUC.

PR #506, opened 2017-01-27
fixed bug when using cached transcriptome

jvolkening:
Location of an '&&' in the <command> block followed by an optional indexing command causes an error if the indexing command is not called (using cached transcriptome) because two '&&' end up adjacent. Moving the first one inside the if block fixes this.


nsoranzo:
@bgruening Travis automatic TS upload failed.

PR #512, opened 2017-02-06
fixing linting issue for cofold

HassanAmr:


HassanAmr:
Hi @bgruening,

The error is related to python 3, right?
should I skip py3k in the bioconda recipe in that case?

PR #516, opened 2017-02-07
perf

HassanAmr:
added citation tag for linting


HassanAmr:
@bgruening,
This doesn't really have a citation, right?

bgruening:
I don't think so.

PR #519, opened 2017-02-13
Add Methyldackel

dpryan79:
This adds MethylDackel (formerly PileOMeth) version 0.2.1 and updates the wrapper with more (probably all, I haven't checked) of the command line options.


dpryan79:
Ping @joachimwolff

bgruening:
@dpryan79 what is the update strategy? Or shouldn't there be one? pileometh should be removed from Galaxy and deprecated from the TS? If not, you could simply rename the pileometh folder and change everything, besides the tool-id ... if the tool-id keeps the same, methyldackel becomes just a newer version of pileometh.

dpryan79:
I'll remove the current PileOMeth folder and change the MethylDackel toolid to be the same. If that's all that's needed for this to appear as an update then that's the easiest way to go.

dpryan79:
The rebase got a bit messy, but this seems to essentially have resulted in git catching most things as a folder rename.

bgruening:
Great thanks, I will play around with this tm a little bit and update it on our instance.

PR #520, opened 2017-02-15
IntaRNA 2.0

zwanli:
A new wrapper for IntaRNA 2.0


zwanli:
@martin-mann here is the PR for the wrapper.

bgruening:
@zwanli can you please add a shed.yml file, so that the tests are triggered?

martin-mann:
@zwanli please ensure to pin the version to 2.0.0, since the interface is going to undergo minor changes with the next release.

Thanks, Martin

tools/rna_tools/intarna/test-data/intarna_result.csv
```diff
@@ -0,0 +1,5 @@
+id1;start1;end1;id2;start2;end2;subseqDP;hybridDP;E
```
bgruening: @zwanli can we change the output to a tab separated file? Simply add a `sed` after the intarna command and change this file? CSV files are not really useful in the Galaxy context.

zwanli: I'm not quit sure if this would work.
@martin-mann , will it cause a problem if I change the format?

martin-mann:
@zwanli since this is a pure output file, there is no problem with intarna.. the only problem might be due to wrong FASTA input headers including tab-whitespaces (that are not pruned by intarna) that might appear as sequence ids in the output and cause problems. but that's nothing we have to care about..

or you can do by putting two consecutive piping `sed` calls: the first converting tabs to space and the second converting semicolon to tab. that way the output should be always correct. :)

thanks,
Martin

tools/rna_tools/intarna/test-data/intarna_result.csv
martin-mann: @zwanli since this is a pure output file, there is no problem with intarna.. the only problem might be due to wrong FASTA input headers including tab-whitespaces (that are not pruned by intarna) that might appear as sequence ids in the output and cause problems. but that's nothing we have to care about..

or you can do by putting two consecutive piping sed calls: the first converting tabs to space and the second converting semicolon to tab. that way the output should be always correct. :)

thanks,
Martin

bgruening:
Thanks @zwanli!!!!

martin-mann:
thanks to both of you!

PR #523, opened 2017-02-16
condafi diffbind, update requirements

martenson:
Diffbind wrapper has no tests, but it is broken anyway, xref. galaxyproject/galaxy#2318 .

Alternative to bgruening#423 .

PR #524, opened 2017-02-16
Properly Condafy diffbind

martenson:
Way better than that crappy #523 by that git @martenson

No idea if this works because there are no tests, but it is broken
anyway, xref. https://github.com/galaxyproject/galaxy/issues/2318 .

Alternative to https://github.com/bgruening/galaxytools/pull/423 .


bgruening:
I try to work on tests ....
Thanks!

PR #525, opened 2017-02-17
Replace column

joachimwolff:
A wrapper which can be used to replace for a specific column all entires by another entry. These entries are defined in a two column file. This is can be useful if someone wants to convert e.g. an ensembl based chromosome naming to a UCSC based.


tools/juicer/juicer_apa.xml
```diff
@@ -0,0 +1,93 @@
+<tool id="Juicer_apa" name="HiC" version="0.1.0">
+
+    <description>Conducting aggregate peak analysis</description>
+
+    <requirements>
+    </requirements>
+    <stdio>
+        <exit_code range="1:" />
+    </stdio>
+    <command>
```
bgruening: Here is something wired in this PR ;)

tools/replaceColumn/replaceColumn.xml
```diff
@@ -0,0 +1,91 @@
...
+            <![CDATA[
+import sys
+
+replace_file = "$replace_information"
+original_file = "$original_file"
+column = int("$column_replace") - 1 ## assume that user thinks column 0 is column 1
+ignore_start_lines = int("$ignore_lines")
+
+if column < 0:
+    print 'No negative column id possible! Columns have to start with \'1\''
```
bgruening: This should be a function, to be python3 compatible...

bgruening: Is this really needed, if the script is only working with the Galaxy wrapper?

tools/juicer/juicer_apa.xml
joachimwolff: Not again... :) I really need to clean up my galaxytools fork.

tools/replaceColumn/replaceColumn.xml
```diff
@@ -0,0 +1,91 @@
...
+    sys.exit(1)
+
+conv = {}
+
+
+fp = open(replace_file, 'r')
```
bgruening: with open() as ...:

tools/juicer/juicer_apa.xml
bgruening: Yeah, you do :)

tools/replaceColumn/replaceColumn.xml
```diff
@@ -0,0 +1,99 @@
...
+            conversion[conv_key_value[0].strip('\n')] = ''
+                    
+
+## read file line by line, search for column entry if it can be replaced. Otherwise it will be skipped.
+with open("output_file", 'w') as output:
+    count = 0
+
+    with open(original_file) as original:
+
+        for line in original:
```
bgruening: enumerate(original) instead of count

tools/replaceColumn/replaceColumn.xml
```diff
@@ -0,0 +1,99 @@
...
+            write = True
+            if count < ignore_start_lines:
+                count += 1 
+                continue
+            line_content = line.split()
+            string_out = ""
+            for i in xrange(len(line_content)):
```
bgruening: for i in line_content?

tools/replaceColumn/replaceColumn.xml
```diff
@@ -0,0 +1,99 @@
...
+                if i == column:
+                    if conversion[line_content[i]] != '':
+                        string_out += conversion[line_content[i]]
+                        string_out += '    '
+                    else:
+                        write = False
+                        break
+                else:
+                    string_out += line_content[i]
+                    string_out += '    '
```
bgruening: '\t'?

tools/replaceColumn/replaceColumn.xml
```diff
@@ -0,0 +1,99 @@
...
+            
+            string_out = string_out.strip()
```
bgruening: is this strip needed, you might remove whitespaces that are needed for some reason in the last column ... rstrip('\n') could be an alternative.

tools/replaceColumn/replaceColumn.xml
```diff
@@ -0,0 +1,99 @@
...
+
+## read file line by line, search for column entry if it can be replaced. Otherwise it will be skipped.
+with open("output_file", 'w') as output:
+    count = 0
+
+    with open(original_file) as original:
+
+        for line in original:
+            write = True
+            if count < ignore_start_lines:
```
bgruening: These lines should also end up in the output, just unmodified, isn't it?

tools/replaceColumn/replaceColumn.xml
```diff
@@ -0,0 +1,99 @@
...
+                        write = False
+                        break
+                else:
+                    string_out += line_content[i]
+                    string_out += '    '
+            
+            string_out = string_out.strip()
+            string_out += "\n"
+
+            if write:
```
bgruening: In which case we do not want to write it out?

tools/replaceColumn/replaceColumn.xml
joachimwolff: The idea is to remove starting lines. Can be useful e.g. the result of PileOMeth is having a first line with some information which need to be excluded if you want to convert the bedgraph to a bigwig file. 
I can include both: Ignore first n lines but keep them or remove them.

tools/replaceColumn/replaceColumn.xml
joachimwolff: I had some troubles with '\t'. The test case was failing because of it, I cannot explain why this happens, the output which is used to compare to was having '\t' too. Probably some troubles with the good old game 'how many blank characters are a tab'. :( If you know here a solution please let me know. Maybe I need to change something in the test case definition.

tools/replaceColumn/replaceColumn.xml
joachimwolff: It can happen that there is no replacement for a value. If you have a look at Devon's chromosome mapping lists e.g. https://github.com/dpryan79/ChromosomeMappings/blob/master/GRCh38_ensembl2UCSC.txt
in line 23 following there is no value for this in UCSC. In this case the line need to be removed.

tools/replaceColumn/replaceColumn.xml
bgruening: Oh, not to skip them from modification. We have a tool for this already ... so I guess this is not needed.

tools/replaceColumn/replaceColumn.xml
bgruening: I have no clue why this should not work ... maybe because of the cheetah stuff ... so because it is in the XML file ... Maybe \\t works?

tools/replaceColumn/replaceColumn.xml
joachimwolff: I keep the lines but let them unmodified. I think it is user-friendly to say 'Ignore first n lines here' in this tool instead of choosing another tool just to remove these lines.

tools/replaceColumn/replaceColumn.xml
joachimwolff: Solved. For reasons I don't know it makes a difference if a string is created in this way: 

```output_string = foo + "\t" + foo +  "\t"``` 

compared to: 

``` output_string = ""
for i in xrange(2): 
      output_string += foo
      output_string += "\t"
```

bgruening:
```
In [2]: foo = 'foo'

In [3]: output_string = foo + "\t" + foo + "\t"

In [4]: print output_string
foo	foo	

In [5]: output_string = ''

In [6]: for i in xrange(2): 
   ...:       output_string += foo
   ...:       output_string += "\t"
   ...: 
   ...: 

In [7]: print output_string
foo	foo	

```

tools/replaceColumn/replaceColumn.xml
```diff
@@ -0,0 +1,109 @@
...
+## read conversion information to index 
+conversion = {}
+
+with open(replace_file, 'r') as conversion_file:
+    for line in conversion_file:
+        conv_key_value = line.strip().split()
+        if len(conv_key_value) == 2:
+            conversion[conv_key_value[0]] = conv_key_value[1]
+        else:
+            conversion[conv_key_value[0]] = ''
```
bgruening: Does this mean that the conversion will create a column with an empty value? So the chromosome column is empty after conversion?

joachimwolff: It means that this line will be removed.

bgruening:
Can you please look at my latest changes? I guess we are ready to merge this, but I would like to make your test maybe a little bit more complicated. can you make up an example, so we have lines in it without a mapping and more complicated mappings and so on....

tools/replaceColumn/replaceColumn.xml
```diff
@@ -46,7 +43,7 @@ with open("output_file", 'w') as output:
             string_out = ""
             for j, line_content_column in enumerate(line_content):
                 if j == column:
-                    if conversion[line_content_column] != '':
+                    if line_content_column in conversion:
```
joachimwolff: Is this an alias? ```foo[bar]``` vs ```bar in foo```? If not, what is the complexity of ```bar in foo```?

bgruening: This is just checking if a key in a dictionary exists.

bgruening:
Thanks @joachimwolff!

jennaj:
Great tool and test cases. xref with https://github.com/galaxyproject/galaxy/issues/860

Future enhancements (aka scope creep) could include:
  1. Tool knows how to replace content in more complex datasets. Eg: fasta, vcf, bam/sam, gtf, gff3. This includes replacing content not only in the data lines, but the header content as appropriate (example: @ lines).
  2. Pre-index very common mappings (these won't change over time) and include on tool form filtered based on the input dataset's assigned database metadata. Eg. The dataset is gtf assigned to hg19, so all mappings _from UCSC hg19 to XXX_ are offered in a pull-down menu. User override with custom input ("from the history") the same as a custom ref genome can be used vs a built-in index.
 3. Consider a Data Manager for the pre-index in item 2. Make it simple - get them all by default. Maybe a list of available files with checkboxes and select all/none that admin user can adjust.
 4. Think about how this could be incorporated into A) any appropriate tool (standardized option on forms) and/or B) workflows (post-job action?). Eg. Input data "screened" to see what it matches now and/or use the assigned database, then available alternative mappings are pre-indexed selectables (I made that word up :1st_place_medal: ).
 5. Guessing the database in general for datasets based on the chromosomes is a more meta enhancement - but would certainly be useful and is likely possible for many (name + length == pretty good way to identify source dbkey/build/version). The utility of this for users - new or experienced - is big, imo, across tools and functions. Could be used to catch input errors at time of job submission and reduce weird tool failures (ideally, tool form would not even submit if there is a mismatch between inputs, but a failure with an informative error message pointing out how to fix [gen|transcript]ome mismatches would still be a win).

Thanks - looking forward to having this tool - just as it is - everywhere! So so useful.

bgruening:
@joachimwolff can you assign yourself and put that in a separate issue?

PR #534, opened 2017-03-03
Hicup

joachimwolff:
Wrapper for hicup.


tools/hicup/.shed.yml
```diff
@@ -0,0 +1,10 @@
+name: hicup
+owner: bgruening
+description: The HiCUP-Pipeline from the Bioinformatics Babraham Institute.
+
+long_description: A galaxy wrapper implementation of the HiCUP-Pipeline from the Bioinformatics Babraham Institute. See http://www.bioinformatics.babraham.ac.uk/projects/hicup/.
+
```
bgruening: Can you change this to be a tool suite, rather than one repo. For example like the openms tools?

tools/hicup/hicup_filter.xml
```diff
@@ -1,4 +1,4 @@
-<tool id="hicup_filter" name="Filter" version="0.1.0">
+<tool id="hicup_filter" name="Hicup_Filter" version="0.1.0">
```
bgruening: This was not needed as the ID will be picked up as repository name.
If you want to include Hicup in the name, please use a whitespace. This is directly exposed to the user.
Btw. why is the "description" tag missing in the tools.

tools/hicup/.shed.yml
```diff
@@ -2,9 +2,14 @@ name: hicup
...
-  
+long_description: A galaxy wrapper implementation of the HiCUP-Pipeline from the Bioinformatics Babraham Institute. See http://www.bioinformatics.babraham.ac.uk/projects/hicup/ 
+
 
 categories:
   - Epigenetics
-remote_repository_url: https://github.com/joachimwolff/galaxytools/tree/hicup/tools/hicup
\ No newline at end of file
+remote_repository_url: https://github.com/joachimwolff/galaxytools/tree/hicup/tools/hicup
+
```
bgruening: Here is something missing to instruct planemo to do the demultiplexing isn't it? Please compare with other examples.

bgruening:
:tada:

PR #535, opened 2017-03-03
Extended REST API for accessibility info from AREsite

jfallmann:
Minor update, adding motif accessibilty info to output for workflow integration


bgruening:
Thanks!

PR #539, opened 2017-03-10
Selectsequences

eggzilla:
Depends on these 2 bioconda-recipe merge requests for:
RNAlien 1.3.6
https://github.com/bioconda/bioconda-recipes/pull/4103
and SelectSequences 1.0.1
https://github.com/bioconda/bioconda-recipes/pull/4100


tools/rna_tools/rnalien/.shed.yml
```diff
@@ -1,7 +1,6 @@
 name: rnalien
 owner: rnateam
 description: RNAlien unsupervized RNA family model construction
-homepage_url: http://rna.tbi.univie.ac.at/AREsite
```
bgruening: You can include a link to your website if you want.

eggzilla: There is a second homepage_url in the code below, that does that :-)

tools/rna_tools/selectsequences/selectsequences.xml
```diff
@@ -0,0 +1,64 @@
+<tool id="SelectSequences" name="SelectSequences" version="1.0.0">
+  <description>
+    Tool to select representative sequences from a multiple sequence alignment.
```
bgruening: Galaxy will create from the name __and__ the description the text in the left-hand-side toolbox. I would remove the `useful` section from the description. Maybe name="Select sequences" description="from representatives of a multiple sequence alignment"?

tools/rna_tools/selectsequences/selectsequences.xml
```diff
@@ -0,0 +1,64 @@
...
+    Useful before running RNAz, RNAcode, RNAalifold on alignments with many entries.
+  </description>
+  <requirements>
+    <requirement type="package" version="1.0.0">selectsequences</requirement>
+  </requirements>
+  <command detect_errors="aggressive"><![CDATA[
+    SelectSequences 
+      -c '$input_clustal'
+      #if $x
```
bgruening: You can replace the entire `if` with just `$x`. You are using falsevalue and truevalue in your param, so in case the option is not active it is an empty string, in case of active it will be `-x`

Issue #540, opened 2017-03-12
methyldackel tool dep builds a file called "bin" instead of "MethylDackel"


bwlang:
ls -l /mnt/galaxy/data/galaxy/tool-deps/methyldackel/0.2.1/bgruening/pileometh/f112bf3dd5ff/bin
-rwxr-xr-x 1 galaxy galaxy 1639993 Mar 11 12:58 /mnt/galaxy/data/galaxy/tool-deps/methyldackel/0.2.1/bgruening/pileometh/f112bf3dd5ff/bin

I fixed the problem by renaming the bin file, creating a bin directory and moving MethylDackel inside it.

bgruening:
@bwlang does this help? https://github.com/bgruening/galaxytools/commit/2953c88908606f4f22b3987bc4ba29bdb4c30479

bwlang:
sorry - nope... 
maybe this is a makefile bug?

Sorry to bug you with this...
I don't know how to make changes to tool dep scripts and test without actually uninstalling and reinstalling from the toolshed. Let me know if there is a better way.

From INSTALLATION_LOG
> 
> make install prefix=/mnt/galaxy/data/galaxy/tool-deps/methyldackel/0.2.1/bgruening/pileometh/f112bf3dd5ff/bin/
> STDERR
> install: target `/mnt/galaxy/data/galaxy/tool-deps/methyldackel/0.2.1/bgruening/pileometh/f112bf3dd5ff/bin/' is not a directory: No such file or directory

bwlang:
tested and working in my environment - thanks!

PR #541, opened 2017-03-12
maybe explicitly create a bin directory?

bwlang:
re #540


bgruening:
Thanks!
Looking at the conda recipe, which is the preferred way now, it looks like this is the correct way. https://github.com/bioconda/bioconda-recipes/blob/master/recipes/methyldackel/build.sh

(conda is creating the bin dir automatically)

PR #546, opened 2017-04-05
WIP: added music to galaxy tools

HassanAmr:


tools/rna_tools/music/music.xml
```diff
@@ -0,0 +1,68 @@
+<tool id="music" name="MUSIC" version="0.1.0">
+    <description>An algorithm for identification of enriched regions at multiple scales in the read depth signals from ChIP-Seq experiments</description>
+    <requirements>
+        <requirement type="package" version="1.0.0">music</requirement>
+    </requirements>
+    <stdio>
+        <exit_code range="1:" />
+    </stdio>
+    <command><![CDATA[
+        cat "$input1" | MUSIC -preprocess SAM stdin .
```
bgruening: single quotes

HassanAmr: Ok, I will fix it in the next push (1 hour hopefully).
Should I follow antarna.xml as an example to cover the various operation options?

bgruening: One of them is probably a better choice: https://github.com/galaxyproject/tools-iuc/

HassanAmr:
@bgruening,

I'm not finished yet. I still have to look over how the output is produced, and finalize some of the input parameters. But I wanted to know if everything is OK so far, otherwise I should repeat it.

HassanAmr:
@bgruening,
the number of output files is dependent on how the program runs. How can I handle that inside the xml?

bgruening:
Same output type? https://planemo.readthedocs.io/en/latest/writing_advanced.html#multiple-output-files

HassanAmr:
Yes. They are all .bed format. Thanks for the link.

HassanAmr:
@bgruening, I am not sure of my regex in the dynamic output part. Other than that, I only need to add conditions for 3 input parameters (since they don't have the default values), and then hopefully the tool will be ready for testing.

tools/rna_tools/music/music.xml
```diff
@@ -88,12 +103,19 @@
     <outputs>
         <data name="output1" format="txt" from_work_dir="chr_ids.txt" />
         <data name="output2" format="txt" from_work_dir="*_mapped_reads.txt" />
+		<data format="bed" name="SSERs_report">
+			<discover_datasets pattern="SSERs_*_designation_and_ext__" directory="." visible="true" />
```
bgruening: You could also `mv SSERs*` into it's own subdir and collect all from this directory with the standard regex.

HassanAmr:
Is it OK to only test the -preprocess option? Because I don't have any adequate test data that will guarantee a non-empty output.

tools/rna_tools/music/music.xml
```diff
@@ -0,0 +1,168 @@
...
+	mkdir SSERs
+	mkdir ERs
+	mv SSERs* SSERs/
+	mv ERs* ERs/ 
+    
+    ]]></command>
+    <inputs>
+
+		<param name="options" type="select" label="Options">
+			<option value="-preprocess">Preprocess</option>
```
bgruening: indentation

tools/rna_tools/music/music.xml
```diff
@@ -0,0 +1,168 @@
...
+        <param name="l_win_min" type="integer" value="100" help="Minimum p-val window length"/>
+        <param name="l_win_max" type="integer" value="5000" help="Maximum p-val window length"/>
+        <param name="l_win_step" type="integer" value="250" help="p-val window length step"/>
+
+    </inputs>
+    <outputs>
+        <data name="output1" format="txt" from_work_dir="chr_ids.txt" />
+        <data name="output2" format="txt" from_work_dir="*_mapped_reads.txt" />
+		<data format="bed" name="SSERs_report">
+			<discover_datasets pattern="__designation_and_ext__" directory="SSERs" visible="true" />
```
bgruening: Indentation, 4 spaces instead of tab.

tools/rna_tools/music/music.xml
```diff
@@ -0,0 +1,168 @@
...
+        <test>
+            <param name="input1" value="input.sam"/>
+            <output name="output1" file="chr_ids.txt"/>
+            <output name="output2" file="*_mapped_reads.txt"/>
+        </test>
+    </tests>
+    <help><![CDATA[
+        USAGE: MUSIC [options] [arguments]
+Options:
+Read Preprocessing:
```
bgruening: here a nice text would be nice.

bgruening:
Test data can be found here: https://github.com/fidelram/deepTools/tree/master/deeptools/test/test_data

HassanAmr:
@bgruening, I think this will also fail, because I still have a parsing error with planemo, but I can't figure out what is wrong.

HassanAmr:
@bgruening, sorry for the stupid mistakes that I didn't notice.
Is it possible to merge this one so that the tool is in Galaxy, and do extra testing on a new branch?

tools/rna_tools/music/music.xml
```diff
@@ -0,0 +1,180 @@
...
+		<data format="bed" name="SSERs_report">
+			<discover_datasets pattern="__designation_and_ext__" directory="SSERs" visible="true" />
+		</data>
+		<data format="bed" name="ERs_report">
+			<discover_datasets pattern="__designation_and_ext__" directory="ERs" visible="true" />
+		</data>
+	</outputs>
+	<tests>
+		<test>
+			<param name="input1" value="input.sam"/>
```
bgruening: indentation

tools/rna_tools/music/music.xml
```diff
@@ -0,0 +1,180 @@
...
+		mkdir SSERs
+		mkdir ERs
+		mv SSERs* SSERs/
+		mv ERs* ERs/
+
+	]]></command>
+    <inputs>
+    	<param name="options" type="select" label="Options">
+			<option value="-preprocess">Preprocess</option>
+			<option value="-sort_reads">Sort Reads</option>
```
bgruening: indentation, you should adopt your editor to 4 spaces instead of tab

tools/rna_tools/music/music.xml
```diff
@@ -0,0 +1,180 @@
...
+	<description>An algorithm for identification of enriched regions at multiple scales in the read depth signals from ChIP-Seq experiments</description>
+	<requirements>
+		<requirement type="package" version="1.0.0">music</requirement>
+	</requirements>
+	<stdio>
+		<exit_code range="1:" />
+	</stdio>
+	<command><![CDATA[
+		#if str($options) == "-preprocess"
+			#if str($input_dir) == "stdin"
```
bgruening: indentation :(

bgruening:
@HassanAmr zero tests where run. There is something wrong here and the test do not super solid.

tools/rna_tools/music/music.xml
HassanAmr: Oh, it should be 4 spaces and not a tab character? Ok, I got it the other way around. I'll fix it.

HassanAmr:
@bgruening, this is just a preprocess command. The actual test is not implemented yet. Or do you mean something else?

bgruening:
No that's what I mean and if you look into travis, there is no test actually running. So planemo is not testing your tool, for unknown reason.

jmchilton:
No tests because the tool does not load into Galaxy:

```
2017-04-07 17:27:18,610 ERROR [galaxy.tools.toolbox.base] Failed to load potential tool /home/travis/build/bgruening/galaxytools/tools/rna_tools/music/music.xml.
  File "/tmp/tmpPkdWhX/galaxy-dev/lib/galaxy/tools/toolbox/base.py", line 701, in quick_load
    tool = self.load_tool( tool_file )
  File "/tmp/tmpPkdWhX/galaxy-dev/lib/galaxy/tools/toolbox/base.py", line 736, in load_tool
    tool = self.create_tool( config_file=config_file, repository_id=repository_id, guid=guid, **kwds )
  File "/tmp/tmpPkdWhX/galaxy-dev/lib/galaxy/tools/__init__.py", line 243, in create_tool
    tool = ToolClass( config_file, tool_source, self.app, guid=guid, repository_id=repository_id, **kwds )
  File "/tmp/tmpPkdWhX/galaxy-dev/lib/galaxy/tools/__init__.py", line 419, in __init__
    raise e
ValueError: The settings for the field named 'no_max_duplicates' require a 'value' setting and optionally a default value which must be an integer
```

nsoranzo:
@jmchilton @guerler I saw this problem a few times previously, it's bit annoying that for an integer/float `param` you have to define `value=""` even if there's no default value, can we remove this restriction?

jmchilton:
@nsoranzo I'm fine with the way it is currently I think - if it is not ``optional="true"`` you need to specify something to stick in the form initially. I'm not sure why ``value=""`` would work in this case - it seems like a bug if it does.

nsoranzo:
Not having `optional="true"` just means that the **user** must fill this value to be able to run the tool, not that there's a default value known to the tool developer. `value=""` is absolutely fine and is used in various tools (e.g. https://github.com/galaxyproject/tools-iuc/blob/master/tools/macs2/macs2_macros.xml#L30 )

jmchilton:
@nsoranzo Huh - okay then. I think it is a bit odd not specify a default - most parameters probably should have defaults IMO. So I'm okay with the developer having to go out of her way to have to annotate that I think.

tools/rna_tools/music/music.xml
```diff
@@ -79,22 +79,22 @@
         <param name="reads_dir" type="text" value="." help="Reads directory"/>
         <param name="output_dir" type="text" value="." help="Output directory"/>
         <param name="sorted_reads_dir" type="text" value="." help="Sorted reads directory"/>
-        <param name="no_max_duplicates" type="integer" help="Max # of duplicates per position"/>
+        <param name="no_max_duplicates" type="integer" value="0" help="Max # of duplicates per position"/>
```
HassanAmr: I think the problem was here, since it was empty and not optional.

bgruening: Is 0 the default value of this parameter?

bgruening:
`NotFound: cannot find 'input_dir'`

guerler:
@nsoranzo I can relax the condition for integer/floats requiring default values. I ll open a PR soon and then we can continue the discussion about that there.

nsoranzo:
@guerler Awesome, thanks!

bgruening:
Please leave in the tests as you are running them locally, even if they are too big.

tools/rna_tools/music/music.xml
```diff
@@ -0,0 +1,163 @@
...
+        #end for
+        #if $control
+            mkdir control &&
+            mkdir control/preprocessed control/sorted control/pruned &&
+            run_MUSIC.csh -preprocess $control_infile control/preprocessed &&
+            run_MUSIC.csh -remove_duplicates control/preprocessed control/sorted control/pruned &&
+        #end if
+
+        mkdir chip &&
+        mkdir chip/preprocessed chip/sorted chip/pruned &&
```
yhoogstrate:
```
 +        mkdir chip &&
 +        mkdir chip/preprocessed chip/sorted chip/pruned &&
```
->
```
 +        mkdir -p chip/preprocessed chip/sorted chip/pruned &&
```

PR #548, opened 2017-04-10
Condafy uniprot_rest_interface

nsoranzo:
Also:
- Fix typos in two `<test>`s.
- Use 'detect_errors' instead of `<stdio>`.
- Update test-data.


bgruening:
Thanks Nicola! It's seems 2 tests are failing - is this some random behavior?

nsoranzo:
Yes, it seems the order of the entries in the FASTA and GFF files are random, I can use `sim_size` or do you have other suggestions?

bgruening:
sime_size should work or include, I would favor simsize.

bgruening:
Thanks a lot!

PR #549, opened 2017-04-13
Compressed input for sailfish salmon

dpryan79:
Xref: https://github.com/galaxyproject/tools-iuc/issues/1128


tools/sailfish/sailfish.xml
```diff
@@ -56,10 +67,24 @@
             --index $index_path
             #if $single_or_paired.single_or_paired_opts == 'single':
                 --libType ${single_or_paired.strandedness}
-                --unmatedReads ./single.$ext
+                #if $compressed == 'GZ':
+                    --unmatedReads <(zcat ./single.$ext)
+                #else if $compressed == 'BZ2':
+                    --unmatedReads <(bzcat ./single.$ext)
```
bgruening: I guess we need to add this as dependency as well, like ` bzip2`.

dpryan79: Isn't it already there on all systems?

bgruening: I would not assume anything ;)
Moreover, Galaxy supports to build containers automatically from these requirements, and then the job is running in a busybox, where this is not installed.

dpryan79: Good point. Is gzip going to be in there or does that need to be added too?

dpryan79: Is there a Galaxy bzip2 package even? If not then these become conda only tools.

dpryan79: Ah, it's package_bzlib_1_0

bgruening: Conda only is ok. Remove the tool_dependencies.xml if you lke.

dpryan79: I'll just make them conda only and be done with it.

dpryan79: Hah, they were already conda only :)

dpryan79: Should I go ahead and make gzip a dependency too? I guess then it's guaranteed to work in the busybox.

bgruening: I had a brief look, but zlib does not have the zcat binary. I was not able to find a conda package and I don't want to bug you with this so much. https://github.com/bgruening/galaxytools/issues/550

bgruening:
Thanks!

PR #552, opened 2017-04-19
Compressed input into trim galore

dpryan79:
This modifies Trim Galore! to handle fastqsanger.gz input. This additionally removes the old hack to handle gzipped input. Note that gzipped input will now result in gzipped output, for consistency. Since at least the IUC tools now all (once the STAR PR is finally merged) accept fastqsanger.gz input this would seem reasonable.


tools/trim_galore/trim_galore.xml
```diff
@@ -49,7 +49,7 @@
         <requirement type="package" version="1.8">cutadapt</requirement>
     </requirements>
     <version_command>
-        perl $__tool_directory__/trim_galore --version
+        perl '$__tool_directory__/trim_galore' --version
```
dpryan79: :+1:

bgruening:
Thanks @dpryan79!

PR #556, opened 2017-05-02
Footprint

dyusuf:


tools/footprint/footprint.xml
```diff
@@ -0,0 +1,214 @@
...
+            input_dir=\$(python -c "import os; print(os.path.dirname('$bam_file'))")
+
+            &&
+            mv \$(find \$input_dir -name "*.PARAM") PARAM
+
+            &&
+            mv \$(find \$input_dir -name "*.RESULTS") RESULTS
+
+            &&
+            mv \$(find \$input_dir  -name "*.plot2.png") plot2.png
```
bgruening: I think `mv *.plot2.png plot2.png` should be enough.

bgruening:
@dyusuf very nice work! I have added one small comment.

tools/footprint/footprint.xml
dyusuf: the problem is that the output dir is not the current working dir.

dyusuf:
@bgruening thanks! :)

tools/footprint/footprint.xml
bgruening: `mv $input_dir/*.plot2.png plot2.png` or is this a wired nested folder hierarchy?

dyusuf: `$input_dir`, cool. this might work. thanks!

dyusuf: i think I misunderstood just now. I thought there is a default galaxy variable called  input_dir. I guess you mean get rid of `find`, right?

bgruening: Yes getting rid of `find` would be nice. But if these files are hidden in arbitrary nested folders, it's fine to use it.

dyusuf: sure. just dit it

tools/footprint/footprint.xml
```diff
@@ -39,16 +39,16 @@
             input_dir=\$(python -c "import os; print(os.path.dirname('$bam_file'))")
 
             &&
-            mv \$(find \$input_dir -name "*.PARAM") PARAM
+            mv \$input_dir/*.PARAM PARAM
```
bgruening: The leading `\` should be removed. You want to use the cheetah variable, not a bash-ism.

dyusuf: File "/tmp/tmp170mOf/galaxy-dev/lib/galaxy/tools/evaluation.py", line 454, in build
    raise e
NotFound: cannot find 'input_dir'

dyusuf: the above message is after removing `\`. input_var is actually a bash variable.

bgruening:
Nice! Thanks!

bgruening:
Awesome! Thanks!

PR #568, opened 2017-05-16
Use the recently built bioconda-diffbind version.

davebx:


bgruening:
You don't want to add test data to this repo, do you? ;)
Usually it has a reason why these tools are here and not on tools-iuc :)

davebx:
Ooh, test data would be a good idea.

PR #569, opened 2017-05-17
wrong format for ph parameter

bernt-matthias:
Taking this parameter as float led to an error.

### DIFF
--- a/chemicaltoolbox/openbabel/ob_convert.xml
+++ b/chemicaltoolbox/openbabel/ob_convert.xml
@@ -381,7 +381,7 @@
         <!-- Options for all formats.-->
         <param name="remove_h" type="boolean" truevalue="-d" falsevalue=""
             label="Delete hydrogen atoms, make all hydrogen implicit" help="(-d)" />
-        <param name="ph" type="float" value="-1" label="Add hydrogens appropriate for pH" help="-1 means deactivated (-p)"/>
+        <param name="ph" type="integer" value="-1" label="Add hydrogens appropriate for pH" help="-1 means deactivated (-p)"/>
         <param name="dative_bonds" type="boolean" truevalue="-b" falsevalue=""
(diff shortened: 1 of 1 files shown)


bernt-matthias:
Or should one change the parsing from int() to float()?

bgruening:
@bernt-matthias can you give me the error? This works for me `obabel -i smi foo -o smi -O foo2 -p 7.0`

bernt-matthias:
In galaxy I see the error: `invalid literal for int() with base 10: '-1.0'` when running a file format conversion from mol2 to pdp.

Here is the full stack:
```
  File "/gpfs1/data/galaxy_server/galaxy-dev/lib/galaxy/jobs/runners/__init__.py", line 170, in prepare_job
    job_wrapper.prepare()
  File "/gpfs1/data/galaxy_server/galaxy-dev/lib/galaxy/jobs/__init__.py", line 971, in prepare
    self.command_line, self.extra_filenames, self.environment_variables = tool_evaluator.build()
  File "/gpfs1/data/galaxy_server/galaxy-dev/lib/galaxy/tools/evaluation.py", line 454, in build
    raise e
ValueError: invalid literal for int() with base 10: '-1.0'
```

This seems to be before the tool is started..? Otherwise it could also be that the mol2 file is in a wrong format?

bgruening:
These wrappers are way more advanced in github than on the TS. I need to update it and will do so in the next days. Sorry. We migrated more or less to Conda dependencies and did not pushed it back to the ToolShed :(

bernt-matthias:
Wonderful. Thanks a lot. Can you close the pullequest when you have done this? Then I have a notification email to try the update :)

bgruening:
So here we go: https://github.com/bgruening/galaxytools/pull/575 
An update long overdue. Let me know what you think.

PR #572, opened 2017-05-18
perl updates to text-utils

bgruening:


bgruening:
Test pass locally and fail here because of the complicated linting error.

PR #577, opened 2017-05-21
CTB: fix silicos-it tools

bgruening:


bgruening:
ping @HassanAmr 
This one is done.

PR #581, opened 2017-05-23
graphclust fix preprocessing shape

mmiladi:


mmiladi:
looks good?

bgruening:
More tests are always good!

mmiladi:
There was also a bug, no test no check :) Can I now rebuild the docker image to have the updated one?

Issue #582, opened 2017-05-24
openbabel convert


bernt-matthias:
I'm wondering why the list of input file types is limited to SDF, MOL, MOL2, CML, InChI and SMILES? I would like to convert pdb to sdf which is possible on the commandline.

bgruening:
Most probably, I just missed it. You could add it here: https://github.com/bgruening/galaxytools/blob/master/chemicaltoolbox/openbabel/macros.xml#L16

But please note that this macro is also used in other tools and we need to make sure other tools can consume pdb as well.

bernt-matthias:
I looked around a bit: 

- ob_remSmall, ob_remDuplicates, ob_depiction_svg, ob_convert directly call babel which should be fine 
- ob_grep and multi_obgrep use obgrep:  I tested successfully: `babel test-data/8_mol.smi test-data/8_mol.pdb; obgrep "CO" -i pdb test-data/8_mol.pdb`
- ob_genProp and ob_addh use python scripts which rely on pybel. This should be fine since pdb is in pybel.informats. 
- Strange is that ob_filter uses pybel but fixed sdf as informat (a problem is that the xml seems to allow sdf,smi,mol,mol2,cml,inchi) should we restrict this in the macro? 
- change_title_to_metadata_value and ob_spectrophore_search explicetely allows only sdf 

So I think extending the informats by pdb should be fine (then the common formats are covered, the alternative to allow everything from pybel.informats would be not a good idea even if this would be handy for convert). How about ob_filter: restrict to sdf or extend the python script to allow everything?

bernt-matthias:
One more thing. If I got the xml file for convert correctly then for some output formats special options are set. Is the same done also for the infiles? Maybe I missed it. I have seen that babel can list the possible options: 

```
babel -L pdb   
One of the formats
pdb -- Protein Data Bank format
Read Options e.g. -as
  s  Output single bonds only
  b  Disable bonding entirely
  c  Ignore CONECT records
```


```
babel -L sdf
One of the formats
sdf -- MDL MOL format
Reads and writes V2000 and V3000 versions

Open Babel supports an extension to the MOL file standard
that allows cis/trans and tetrahedral stereochemistry to be
stored in 0D MOL files. The tetrahedral stereochemistry is
stored as the atom parity, while the cis/trans stereochemistry
is stored using Up and Down bonds similar to how it is
represented in a SMILES string. Use the ``S`` option
when reading or writing if you want to avoid storing
[...]

PR #583, opened 2017-05-24
add pdb to the openbabel xml macro

bernt-matthias:
See issue #582


bgruening:
Thanks @bernt-matthias!

PR #585, opened 2017-05-25
Option to skip ps drawing in rnafold

mmiladi:
the ps secondary structure drawing of the structures is not always needed. Skip option avoids large overhead of ps files generation and handling.


bgruening:
@mmiladi I think you need to `filter` the output as well, similar to https://github.com/mmiladi/galaxytools/blob/1dde76f1a9fec028e103942c1462cb111f20ee54/tools/rna_tools/vienna_rna/rnafold.xml#L295 and please increase the version number of this tool.

mmiladi:
@bgruening many thanks!

PR #587, opened 2017-05-31
Add data manager for SortMeRNA

bebatut:
Hi,

Here is a data manager for SortMeRNA. You can test it (on the TTS under bebatut).

Bérénice


bgruening:
Wonderful! Thanks a bunch!

bgruening:
Have a save trip and much fun tomorrow!

nsoranzo:
@bebatut I was going to update my SortMeRNA installation when I noticed that this is probably going to break the old installations because of the added `dbkey` column to the `rRNA_databases` tool data table. Since I don't think this `dbkey` is used at all (it's just a copy of the second column), can we get rid of it before too many people install it?

PR #591, opened 2017-06-06
Fix SortMeRNA data manager

bebatut:
Hi,

As mentioned by @nsoranzo in #587, I removed the `dbkey` column in the data table to avoid conflicts with old installations.

Bérénice


data_managers/data_manager_sortmerna_database_downloader/data_manager/data_manager_sortmerna_download.py
```diff
@@ -139,8 +139,7 @@ def move_index_files(archive_content_path, target_dir, data_tables, version):
         add_data_table_entry(
             data_tables,
             "rRNA_databases",
-            dict(
-                dbkey=db_name,
+            dict(
```
nsoranzo: There's some extra spaces at the end of this line.

nsoranzo:
Thanks @bebatut!

PR #592, opened 2017-06-10
Update methyl dackel

dpryan79:
This is version 0.3.0, which supports multiple threads and detecting/avoiding likely SNPs. This also fixes a bug with methylation extraction from the last 1-3 bases in a contig.


bgruening:
Oh this sounds wonderful! Sorry for the late merge!

PR #593, opened 2017-06-12
(Another) Fix for the SortMeRNA data manager

bebatut:
Hi,

I had some issues with SortMeRNA because of:

- Duplication of the values in the tool data table
- Incomplete copy of the indexed files

I also simplified the extraction of the path to the cached data in the SortMeRNA wrapper.

I ran some tests locally (using the data manager on the TTS). It worked...

Bérénice

### DIFF
--- a/data_managers/data_manager_sortmerna_database_downloader/data_manager/data_manager_sortmerna_download.py
+++ b/data_managers/data_manager_sortmerna_database_downloader/data_manager/data_manager_sortmerna_download.py
@@ -126,23 +126,28 @@ def move_index_files(archive_content_path, target_dir, data_tables, version):
         if not filename.endswith("fasta"):
             continue
         input_filepath = os.path.join(file_dir, filename)
-        output_filepath = os.path.join(target_dir, filename)
-        # Move file
-        os.rename(input_filepath, output_filepath)
+        # Extract the db name
+        db_name = os.path.splitext(filename)[0]
+        # Create the directory where to put the fasta files and indexed files
+        filedir = os.path.join(target_dir, db_name)
+        os.mkdir(filedir)
+        fasta_filepath = os.path.join(filedir, "%s.fasta" % db_name)
+        indexed_filepath = os.path.join(filedir, db_name)
+        # Move the fasta file
+        os.rename(input_filepath, fasta_filepath)
         # Index the file with indexdb_rna
         command = "indexdb_rna --ref %s,%s" % (
-            output_filepath,
-            os.path.splitext(output_filepath)[0])
+            fasta_filepath,
+            indexed_filepath)
         process = subprocess.call(command, shell=True )
         # Add entry in the data table
-        db_name = os.path.splitext(filename)[0]
         add_data_table_entry(
             data_tables,
             "rRNA_databases",
             dict(
                 value="%s-%s" %(version, db_name),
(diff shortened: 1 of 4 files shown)


nsoranzo:
@bebatut I'm also looking at SortMeRNA and this data manager again. It seems that with the changes in https://github.com/bgruening/galaxytools/pull/587/files#diff-d0c9f2272b42fd05a0eb52130a807695 SortMeRNA expects that the 3rd column of `rRNA_databases.loc` does not contain the `.fasta` extension any more, which also break the backward compatibility with the existing loc files.

nsoranzo:
Uh, it seems that should exactly fix that, thanks @bebatut!

data_managers/data_manager_sortmerna_database_downloader/data_manager/data_manager_sortmerna_download.py
```diff
@@ -126,23 +126,28 @@ def move_index_files(archive_content_path, target_dir, data_tables, version):
...
-        os.rename(input_filepath, output_filepath)
+        # Extract the db name
+        db_name = os.path.splitext(filename)[0]
+        # Create the directory where to put the fasta files and indexed files
+        filedir = os.path.join(target_dir, db_name)
+        os.mkdir(filedir)
+        fasta_filepath = os.path.join(filedir, "%s.fasta" % db_name)
+        indexed_filepath = os.path.join(filedir, db_name)
+        # Move the fasta file
+        os.rename(input_filepath, fasta_filepath)
```
nsoranzo: You should use [shutil.move](https://docs.python.org/2/library/shutil.html#shutil.move) because [os.rename](https://docs.python.org/2/library/os.html#os.rename) may not work across filesystems.

bgruening:
Thanks @bebatut and  @nsoranzo!

bebatut:
Thanks @bgruening for the last fix. I tested everything again this morning: it looks great!

nsoranzo:
Planemo `shed_update` failed https://travis-ci.org/bgruening/galaxytools/jobs/242311769 :
```
cd '/home/travis/build/bgruening/galaxytools/tools/rna_tools/sortmerna' && git rev-parse HEAD
cd '/home/travis/build/bgruening/galaxytools/tools/rna_tools/sortmerna' && git diff --quiet
Could not update sortmerna
Unexpected HTTP status code: 400: {"err_msg": "You do not have permission to update this repository."}
Unexpected HTTP status code: 403: {"err_msg": "You are not the owner of this repository, so you cannot administer it.", "err_code": 403005}
Failed to update repository contents and metadata for repository 'sortmerna' on the test Tool Shed.
```

bgruening:
Updated it manually, planemo has some problems with different users in shed.yaml file, especially if you create new repos. Never found time to fix this.

nsoranzo:
Thanks @bgruening! Wouldn't it be easier to give permissions to your TS user to update the rnateam repos?

nsoranzo:
Also, `data_manager_sortmerna_database_downloader` needs to be updated manually too, sorry!

bgruening:
In theory this should be true already. I should be part of the rnateam group. But even if this would work, planemo does not take the user from the shed.yaml file when a new repo needs to be created.
Uploaded the DM as well, sorry for missing this.

PR #594, opened 2017-06-12
modified crt command section

HassanAmr:


bgruening:
https://github.com/bgruening/galaxytools/blob/master/.tt_blacklist#L26 it blacklisted and not tested ...

HassanAmr:
Thanks :)
I had problems with local testing, so I was trying to fix them before testing here. But glad to see it was only locally.

PR #595, opened 2017-06-14
WIP: Imagej2

HassanAmr:
I will try to get the tools in this folder off the blacklist


bgruening:
Does this work for you locally? I think these tools can not be tested with the local Galaxy job runner, only a real one, which is not the default one in planemo - means we can not tests these reliably.

HassanAmr:
Doesn't this mean it tested ok?
https://travis-ci.org/HassanAmr/galaxytools/builds/242672706

bgruening:
It was testing something yes (no clue why), but if you look at the tools that are tested - its not imageJ.

HassanAmr:
Ok, I will revisit this maybe in the future

PR #596, opened 2017-06-14
mirdeep2

HassanAmr:
removed mirdeep2 from the blacklist


bgruening:
And now you need to change something in the wrapper to trigger a test.

HassanAmr:
Is it ok that I altered the structured text in normal text and that I delete the attribute from a param: `metadata_name="dbkey"`?

tools/rna_tools/miRDeep2/miRDeep2_quantifier/quantifier.xml
```diff
@@ -1,17 +1,20 @@
...
-]]>
-    </description>
+    <description>fast quantitation of reads mapping to known miRBase precursors</description>
     <requirements>
-        <requirement type="package" version="2.0">mirdeep2_quantifier</requirement>
-        <requirement type="package" version="0.12.7">bowtie</requirement>
-        <requirement type="package" version="5.18.1">perl</requirement>
-        <requirement type="package" version="1.8.5">vienna_rna</requirement>
-        <requirement type="package" version="2.023">pdf_api2</requirement>
+        <requirement type="package" version="2.0.0.8">mirdeep2</requirement>
```
bgruening: bowtie and everything else is already defined as dependency here: https://github.com/bioconda/bioconda-recipes/blob/53f88dc6893c74cf170024401eaf62650a92ead3/recipes/mirdeep2/2.0.0.8/meta.yaml
So you don't need to add it again here.

bgruening:
Either this is missing from the package or it can not be found: `Error:	 Rfam_for_miRDeep.fa not found in your miRDeep2 scripts directory`

HassanAmr:
Yes, but I don't know why. The file should be there (I downloaded the same package and checked inside). I will look into the code itself, maybe I need to patch it.

HassanAmr:
The following is the error I get:
```
sanity_check_mature_ref.pl mature_ref_this_species.fa

Error: problem with mature_ref_this_species.fa
```

Is it possible that the following doesn't comply with "Sequences are allowed only to comprise characters [ACGTNacgtn]" which is a condition found in `sanity_check_mature_ref.pl`:
```
>cel-miR-36
TCACCGGGTGAAAATTCGCATG
>cel-miR-37
TCACCGGGTGAACACTTGCAGT
>cel-miR-38
TCACCGGGAGAAAAACTGGAGT
>cel-miR-39
TCACCGGGTGTAAATCAGCTTG
>cel-miR-40
TCACCGGGTGTACATCAGCTAA
>cel-miR-41
TCACCGGGTGAAAAATCACCTA
>cel-miR-229
AATGACACTGGTTATCTTTTCCATCG
```

HassanAmr:
Hi @bgruening,
I figured out what the problem is. When I patched miRDeep2 the first time, I removed `#!/usr/bin/perl -W` and replaced them with `#!/usr/bin/env perl -W`. That's why the `sanity_check_mature_ref.pl` wasn't working. So I will remove the `-W`. Is it safe to remove `-W` from all files that include it?

bgruening:
Cool glad to hear.

bgruening:
Is this still WIP?

HassanAmr:
I think it is ready for merge, unless you want me to add something to the tool.

PR #604, opened 2017-07-12
Enable stdout SAM file to be written as history item

malloryfreeberg:
Updated <outputs> and <command> sections. When --writeMappings is set to Yes/True, a SAM-formatted file is written to stdout and redirected to its own history item (name=output_sam).

NOTE: This filter condition (line 342) seems to always evaluate to True, even when writeMappings is False. Needs to be changed (I'm not sure what to) so that a SAM file isn't produced as output if writeMappings is set to 'No'/False. Any ideas, @bgruening ?

### DIFF
--- a/tools/salmon/salmon.xml
+++ b/tools/salmon/salmon.xml
@@ -339,7 +339,7 @@
             <filter>geneMap</filter>
         </data>
         <data name="output_sam" format="sam" label="${tool.name} on ${on_string} (SAM format)">
-            <filter>str(adv.writeMappings)</filter> <!-- This filter condition seems to always evaluate to True, even when writeMappings is False. Needs to be changed (I'm not sure what to) so that a SAM file isn't produced as output if writeMappings is set to 'No'/False. -->
+            <filter>adv['writeMappings']</filter>
         </data>
     </outputs>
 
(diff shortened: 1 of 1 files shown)


tools/salmon/salmon.xml
```diff
@@ -336,6 +338,9 @@
         <data name="output_gene_quant" format="tabular" from_work_dir="output/quant.genes.sf" label="${tool.name} on ${on_string} (Gene Quantification)">
             <filter>geneMap</filter>
         </data>
+        <data name="output_sam" format="sam" label="${tool.name} on ${on_string} (SAM format)">
+            <filter>str(adv.writeMappings)</filter> <!-- This filter condition seems to always evaluate to True, even when writeMappings is False. Needs to be changed (I'm not sure what to) so that a SAM file isn't produced as output if writeMappings is set to 'No'/False. -->
```
bgruening: can you try `<filter>adv['writeMappings']</filter>`

malloryfreeberg: Aha! That works :) Thanks!

bgruening: Thanks to you!!!

bgruening:
ToolShed is updated, thanks a lot!

PR #609, opened 2017-07-25
Segemehl fix

jfallmann:
Fixed broken galaxy dependency, now conda only


jfallmann:
@bgruening Sorry there is some old commit that belons to the aresite2 branch, whyever it made it's way into my local master, however, segemehl was not working out of galaxy due to a broken dependency (package_segemehl), with pure conda install it works now

bgruening:
@jfallmann the commits are ok, I think but have you changed indentation of the wrapper? Can you please make sure you use 4 spaces intend? The diff is quite large and therefor hard to review.

tools/segemehl/tool_data_table_conf.xml.sample
```diff
@@ -1,8 +0,0 @@
-<!-- Use the file tool_data_table_conf.xml.oldlocstyle if you don't want to update your loc files as changed in revision 4550:535d276c92bc-->
```
bgruening: This file needs to stay ... it enables shared data resources between users.

jfallmann:
@bgruening Sorry for the intent, was in the wrong emacs mode, should be fixed and the data_table_conf is back

bgruening:
Thanks a lot!

bgruening:
If you change parameters or add dependencies please increase the build number here as well - This way we get a new version of the tool in Galaxy.

PR #610, opened 2017-07-25
Reduce genomics/proteomics clashes. Change sensitivity flags to select menu. 

caleb-easterly:
The phrase 'reference genome' was confusing when using Diamond for proteomics. Also, the choice of query translation is not relevant for proteins. 

Finally, I changed the two sensitivity flag options to a select menu, to avoid possible option clash if someone is confused about how the options 'stack' (like I was) - see <https://github.com/bbuchfink/diamond/issues/160>.

### DIFF
--- a/tools/diamond/diamond.xml
+++ b/tools/diamond/diamond.xml
@@ -38,8 +38,12 @@
         #end if
 
         --compress '0'
-        $sensitive
-        $more_sensitive
+        #if $sensitivity == "1"
+          --sensitive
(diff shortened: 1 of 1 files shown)


bgruening:
@caleb-easterly thanks a lot for your changes! I like you sense of improving UX - very much appreciated. We can do so much better by simply enhancing our help text and restructuring of parameters! This PR is a great step in this direction - keep the PRs coming :)

PR #615, opened 2017-08-03
Initialized tool collection for for identification of tRNA modificati…

jfallmann:
…ons from NGS data galaxy workflow


tools/rna_tools/trna_mod/addCCA.pl
```diff
@@ -0,0 +1,34 @@
+#!/usr/bin/perl
```
bgruening: /usr/bin/env perl

tools/rna_tools/trna_mod/clustering.xml
```diff
@@ -0,0 +1,24 @@
+<tool id="clustering" name="clustering of identicial sequences" version="0.1.0">
+  <description>Clustering of identical sequneces</description>
+  <command interpreter="perl">clustering.pl $input $output1 $output2</command>
+  <inputs>
+    <param format="fasta" name="input" type="data" label="tRNAscan_mature.fasta"/>
```
bgruening: strange label

tools/rna_tools/trna_mod/addCCA.xml
```diff
@@ -0,0 +1,22 @@
+<tool id="addCCA" name="add CCA tails to tRNA chromosomes" version="0.1.0">
+  <description>Add CCA tails to tRNA chromosomes</description>
+  <command interpreter="perl">addCCA.pl $input $output</command>
+  <inputs>
+    <param format="fasta" name="input" type="data" label="tRNAscan_fasta"/>
```
bgruening: The label could be improved.

tools/rna_tools/trna_mod/addCCA.xml
```diff
@@ -0,0 +1,22 @@
+<tool id="addCCA" name="add CCA tails to tRNA chromosomes" version="0.1.0">
```
bgruening: As a general comment to this tool ... you can archive the same with a small workflow ...

fasta-to-tabular -> add column (CCA) -> merged columns -> tabular to fasta

bgruening: If you think this tool is easier, I would suggest to make the sequence which you add optional ... so that people can use this tool in more use-cases. You could even think of adding a sequence at the beginning as an option....

tools/rna_tools/trna_mod/bbduk_wrapper.xml
```diff
@@ -0,0 +1,46 @@
+<tool id="tRNAmodbbduk" name="bbduk_for_tRNAmod" version="0.1.0">
+  <description>Preprocesses reads with bbduk according to best practice for tRNA modification analysis</description>
+  <requirements>
+	<requirement type="package">bbmap</requirement>
+  </requirements>
+  <command>
+	<![CDATA[
+			 #if $input.is_of_type("fastq.gz", "fastqsanger.gz"):
```
bgruening: The indentation looks off here.

tools/rna_tools/trna_mod/bbduk_wrapper.xml
```diff
@@ -0,0 +1,46 @@
+<tool id="tRNAmodbbduk" name="bbduk_for_tRNAmod" version="0.1.0">
```
bgruening: The name looks geeky.
Should this be a general bbduk tool?

tools/rna_tools/trna_mod/bbduk_wrapper.xml
```diff
@@ -0,0 +1,46 @@
...
+			 bbduk.sh in=${input}.fastq.gz out=${output}.fastq ref=${adapter}.fa mink=8 ktrim=r k=10 rcomp=t hdist=1 qtrim=rl trimq=25 minlength=50 maxlength=100 overwrite=t	 
+			 #else
+			 ln -sf "${input}" "${input}.fastq";
+			 ln -sf "${adapter}" "${adapter}.fa";
+			 ln -sf "${output}" "${output}.fastq";
+			 bbduk.sh in=${input}.fastq out=${output}.fastq ref=${adapter}.fa mink=8 ktrim=r k=10 rcomp=t hdist=1 qtrim=rl trimq=25 minlength=50 maxlength=100 overwrite=t
+			 #end if
+	]]>
+  </command>
+  <stdio>
```
bgruening: You can replace this by detect_errors: https://docs.galaxyproject.org/en/latest/dev/schema.html#detect-errors

tools/rna_tools/trna_mod/bbduk_wrapper.xml
```diff
@@ -0,0 +1,46 @@
...
+    <exit_code range="1:" />
+    <exit_code range=":-1" />
+    <regex match="Error:" />
+    <regex match="Exception:" />
+  </stdio>
+  <inputs>
+    <param name="input" type="data" format="fastq,fastq.gz,fastqsanger,fastqsanger.gz,fastqsanger.bz2"/>
+	<param name="adapter" type="data" format="fasta"/>
```
bgruening: indentation ...

tools/rna_tools/trna_mod/gatk_wrapper.xml
```diff
@@ -0,0 +1,90 @@
...
+  <requirements>
+	<requirement type="package" version="2.7">python</requirement>
+	<requirement type="package" version="1.4.1">samtools</requirement>
+	<requirement type="package" version="2.9.2">picard</requirement>
+	<requirement type="package" version="8.0.92">java-jdk</requirement>
+  </requirements>
+  <command>
+	<![CDATA[
+			 ##cleanup
+			 rm -f "${reference}.dict";
```
bgruening: Uha, what is this? You are removing datasets from Galaxy database, I don't think this is a good idea :)

tools/rna_tools/trna_mod/gatk_wrapper.xml
```diff
@@ -0,0 +1,90 @@
...
+			 rm -f "${reference}.fasta.fai";
+			 rm -f "${input}.bam.bai";
+			 rm -f "${input}.bai";
+			 ## prepare fasta index and dict and link files for correct file type
+             ln -sf "${input}" "${input}.bam";
```
bgruening: Single quotes

tools/rna_tools/trna_mod/gatk_wrapper.xml
```diff
@@ -0,0 +1,90 @@
...
+			 #if str($gatktool.tool) == "IndelRealigner":
+			     ln -sf "${intervals}" "${intervals}.intervals";
+			 #end if
+			 samtools faidx "${reference}.fasta" ;
+			 picard CreateSequenceDictionary R= "${reference}.fasta" O= "${reference}.dict";
+			 samtools index "${input}.bam";
+			 picard BuildBamIndex I="${input}.bam" O="${input}.bam.bai";
+			 			 
+			 ##now run gatk
+			 java -jar $__tool_directory__/GenomeAnalysisTK.jar
```
bgruening: There is now a conda package for GATK4 ... the problem with older versions and this PR is that we are not allowed to distribute the JAR file ... the license does not allow this for GATK < 4 :( sorry.

jfallmann: @bgruening Hi, thanks for checking the tools. a general comment, this is jsut a collection of tools for a particular workflow, thus the wrapper for e.g. bbduk, as parameters are different from default and should not be changed by the user.
There is no way of using gatk4 as the workflow depends on gatk3.6, the bioconda solution for that is to just download the binary and link it to the bioconda dir, I can not change that.
Version 4 is also in beta and does not offer the utilities we need (yet), maybe when they release the official build at the end of september, if so I'll adapt the tools accordingly.
However, the travis build will always fail due to the gatk dependency, I have no idea how to change that and still have proper tests.

tools/rna_tools/trna_mod/gatk_wrapper.xml
jfallmann: no, just the link to a dataset, if it exists

jfallmann:
will reopen with gatk4 if possible

PR #627, opened 2017-09-07
intarna: format text -> txt

bernt-matthias:
As far as I understood it the file format should be txt (not txt).


bgruening:
correct, the datatype needs to be one from: https://github.com/galaxyproject/galaxy/blob/dev/config/datatypes_conf.xml.sample#L298

PR #637, opened 2017-09-27
Catching warnings with regex

jfallmann:


bernt-matthias:
Good idea. 

@jfallmann can you check if this also works with https://github.com/bgruening/galaxytools/pull/697 when breakmaf.sh is called and the parts are processed separately. I'm not sure if the warnings are still printed in this case (which would be good).  

I also would like to have the parts where rnacode fails printed to stderr (I think I forgot this in my PR). Then we can redirect them to https://github.com/wash/rnacode/issues :)

jfallmann:
@bernt-matthias Ok, will check this. I guess a lot of the problems occuring with code will also be relevant for the rnaz wrapper, there is currently absolutely no output produced that should be used by anyone. Maybe it's time for a cleanup hackathon? Been thinking about that for quite some time now and I get the feeling it's the best way to get things running in an coordinated effort. @bernt-matthias @bgruening should we just do this?

bernt-matthias:
Jep. The breakmaf+processmaf scripts are essentially a workaround for the failing RNAcode. In the best case RNAcode itself would be fixed :) 

Another thing is that the while loop in processmaf could be used for parallelisation. This might be useful for RNAz as well.

PR #640, opened 2017-09-28
Working on simultaneous usage of mea and partfunc

jfallmann:
@bgruening hope this does not break anything


mmiladi:
Hi @jfallmann ,  thanks for helping to push the vienna-wrapper  ahead :-)
I can see that this PR skips MEA option selection. Do you intend to make the galaxy wrapper always generate multiple structures lines in MEA mode (MFE, MEA, ...) per input sequences instead of the single MFE structure by default?

jfallmann:
Hi @mmiladi , no this was just a workaround to get fold running for a summerschool, there is still enough to fix in the package, I intent to start this with the new year, maybe we can do this in a constructive way so we generate a list of all the building sites first and then fix them, maybe we even start a kind of hackaton for this but I'm not sure who would be interested to join, would you, maybe during the bled meeting?

mmiladi:
Hi Joerg, Thanks for the feedback. This would be a very good idea. Although I would say vrna wrappers generally are in a good state, there is always room for improvement. Maybe we can start create an issue to collect a list of things that can be fixed/improved. Cheers.

PR #641, opened 2017-09-28
Adding Galaxy wrapper for Hansel.

hellothisisMatt:
Note: I left attrs within the requirements as without an updated version of conda and galaxy, Hansel will fail travis ci testing.

### DIFF
--- a/hansel/.shed.yml
+++ b/hansel/.shed.yml
@@ -1,0 +1,7 @@
+categories: [Sequence Analysis]
+description: Heidelberg and Enteritidis SNP Elucidation
+name: bio_hansel
+owner: nml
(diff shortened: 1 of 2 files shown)


hansel/.shed.yml
```diff
@@ -0,0 +1,7 @@
+categories: [Sequence Analysis]
+description: Heidelberg and Enteritidis SNP Elucidation
+name: bio_hansel
+owner: nml
```
bgruening: Not sure if you submitted it to the wrong repo :)
IUC repo would also be a good candidate.

bgruening:
Looks pretty good to me. I would like to see one test for the collections, but this is really nice!

hellothisisMatt:
Woopsies, wrong repo! Will be submitting this recipe to IUC. Closing.

PR #642, opened 2017-09-28
add first round of pgtools

bgruening:
This will use a in-user-space postgresql version and can load schemas into it and query the db. Many more tools will be added in the future. Open for comments.


tools/pg_tools/macros.xml
```diff
@@ -0,0 +1,63 @@
...
+    </token>
+
+    <token name="@PG_STOP@"><![CDATA[
+        '$__tool_directory__/pglite' stop  -d ./postgresql &&
+        sleep 10
+    ]]>
+    </token>
+
+    <token name="@ARCHIVE_DATABASE@"><![CDATA[
+        tar -cvJf postgresql_out.tar.xz ./postgresql
```
erasche: use the html extra files path thing to avoid the overhead of tarring / un-tarring.

tools/pg_tools/macros.xml
```diff
@@ -0,0 +1,63 @@
...
+
+    <token name="@PSQL@"><![CDATA[
+        psql "\$('$__tool_directory__/pglite' url -d ./postgresql)"
+    ]]>
+    </token>
+
+
+    <token name="@PG_START@"><![CDATA[
+        '$__tool_directory__/pglite' start  -d ./postgresql &&
+        sleep 10
```
erasche: slightly worried about this. Wonder if we could switch a while loop?

tools/pg_tools/pg-query.xml
```diff
@@ -0,0 +1,48 @@
+<tool id="pg_query" name="Postgresql" version="@PG_VERSION@">
+    <description>import sql dump</description>
```
erasche: description

tools/pg_tools/pg-query.xml
```diff
@@ -0,0 +1,48 @@
...
+            @PSQL@ -c '$query' -L logfile.log > '$outfile' &&
+            @PG_STOP@ &&
+            cat logfile.log
+    ]]>
+    </command>
+    <inputs>
+        <param format="postgresql" name="infile" type="data" label="Input database" />
+        <param name="query" type="text" area="True" size="5x50" label="The query to be issued to the database">
+            <sanitizer>
+                <valid initial="string.letters,string.digits">
```
erasche: this won't be enough, we'll need `'` at least.

tools/pg_tools/pg-query.xml
```diff
@@ -0,0 +1,48 @@
...
+    <macros>
+        <import>macros.xml</import>
+    </macros>
+    <expand macro="requirements" />
+    <command detect_errors="aggressive"><![CDATA[
+            @UNTAR_INFILE@ &&
+            @PG_START@ &&
+            ls -l ./postgresql
+            &&
+            @PSQL@ -c '$query' -L logfile.log > '$outfile' &&
```
erasche: could we wrap this in a `copy ($query) to stdout with csv delimiter '	';` (i.e. raw tab char)? Then you'd get automatic tabular files. (or at least make it an option and make tab output default)

tools/pg_tools/macros.xml
erasche: Retrated: out of band discussion concluded that disk space was more important than the computational overhead.

tools/pg_tools/pg-query.xml
bgruening: fixed, other can be at any time.

tools/pg_tools/pg-query.xml
bgruening: ?

tools/pg_tools/macros.xml
bgruening: We could put the sleep in a while look and ask if the server is available? Not sure what the best way is here. Maybe we could make this start command blocking?

bgruening:
I added a new test and addressed your other concerns, besides the sleep thing.

tools/pg_tools/pg-query.xml
erasche: change description from "import sql dump" to something like "query postgres database"

tools/pg_tools/pglite
```diff
@@ -0,0 +1,211 @@
...
+function err { local x=$? ; msg "$*" ; return $(( $x == 0 ? 1 : $x )) ;}
+function out { printf '%s\n' "$*" ;}
+
+# Handles "no-match" exit code specified by POSIX for filtering tools.
+function maybe { "$@" || return $(( $? == 1 ? 0 : $? )) ;}
+
+
+######################### Delegates to subcommands or runs main, as appropriate
+
+main "$@"
```
erasche: We need to document the license / source+license for this script somewhere, i.e. https://github.com/solidsnack/pglite/blob/master/LICENSE

tools/pg_tools/macros.xml
erasche: I'd love a sleep in a while. Maybe something like the following would be good so we'd be safe from sleeping too much or too little, and we can be sure we won't sleep forever if something goes wrong.

```bash
exec_count=0
'$__tool_directory__/pglite' status | grep -F -q 'server is running'
while [ $? -ne 0 ]; do
    sleep 2
    exec_count=$((exec_count+1))
    if [[ "$exec_count" -gt 10 ]]; then exit 1; fi
    '$__tool_directory__/pglite' status | grep -F -q 'server is running'
done
```

(Completely untested.)

tools/pg_tools/pglite
bgruening: I don't think we should ship this tool at all, we should rather depend on it.

erasche: works for me!

bgruening: removed this file and we now depend on a conda package

tools/pg_tools/macros.xml
bgruening: I used, unless ... `status` still does not work reliably, but checking for the socket file in addition does the trick.

bgruening:
So tests are passing. Its a start I guess ...

tools/pg_tools/macros.xml
```diff
@@ -0,0 +1,74 @@
...
+    ]]>
+    </token>
+
+    <token name="@PSQL@"><![CDATA[
+        psql "\$(pglite url -d ./postgresql)"
+    ]]>
+    </token>
+
+    <token name="@STATUS_RUNNING@"><![CDATA[
+        timeout 30 bash -c 'until pglite status -d ./postgresql | grep -F -q "server is running"; do sleep 1; done' &&
```
erasche: just fyi, the reason for my horribly long while loop was that [timeout is unavailable on OSX](https://stackoverflow.com/questions/3504945/timeout-command-on-mac-os-x). Not sure if we care about those guys (I don't really that much.)

bgruening: I don't care that much either and doubt that anyone one is running a server under OSX.
I think this is ready to go for a first version.

bgruening:
Like the additional tool!

yvanlebras:
Ready to start things on a postgis version with @ValentinChCloud and @Alanamosse for Galaxy-E!

bgruening:
@yvanlebras awesome. Let me know if these tools work for you and what I can do to improve it.
One other requirement would be a postgresql-gis conda package :)

yvanlebras:
Have you a postgresql datatype definition already ?

erasche:
@yvanlebras  https://github.com/galaxyproject/galaxy/pull/4730

yvanlebras:
Thanks @erasche ! I missed it ! Nice end of week.

bgruening:
Ok, let's get this in. I'm currently trying to ship with a GIS and rdkit enabled PG version, but this can also be an update. 

Thanks everyone!

ping @abretaud

bgruening:
ping @annefou, this might be interesting for you as well. It can use GIS-enabled postgresql databases in user-space. You can treat a PG database like any other Galaxy database.

yvanlebras:
This is something we have to work on. Maybe it's time to go further and try to test creating a qgis 3 interactive took!

annefou:
> This is something we have to work on. Maybe it's time to go further and try to test creating a qgis 3 interactive took!

Yes. We can discuss it during the co-fest?!

bgruening:
Sure :)

yvanlebras:
Yes! Very good idea Anne. I will try to have some GIS colleagues, maybe you have this ?

annefou:
Yes we use  arcGIS a lot but more and more QGIS too. We have dockerized versions of QGIS desktop that could be added to Galaxy. I am not sure what is best so we can discuss it during the co-fest!

yvanlebras:
I was notably thinking about something like thishttps://github.com/jancelin/geo-poppy

PR #646, opened 2017-10-18
IntaRNA v2.1.0 enabled + citation added

martin-mann:


martin-mann:
@bgruening Hi Björn. the IntaRNA 2.1.0 package uses VRNA 2.4.1 but galaxy wrappes so far some older version.. is this a problem?

bgruening:
@martin-mann this is not a problem. Both tools are independent, but it would be also nice to update the vrna tools. I think @jfallmann is working on this.

bgruening:
Thanks!

PR #650, opened 2017-11-08
Osra update

bgruening:
ping @HassanAmr :)

Lets see how it works with the new version!


HassanAmr:
I think I need to debug this locally and evaluate it more carefully. Every png file I use causes a core dump.

PR #652, opened 2017-11-09
use BAM everywhere, include command_version, remove hicup_hicup

bgruening:


bgruening:
@lldelisle sorry it took so long. We were able to get a `bam_native` datatype into the next version of Galaxy. So if you like we can now go for BAM.

lldelisle:
Hi,
hicup have a major update. I think it is better to integrate it. I just made a PR to bioconda, then I will do the tests with the test dataset provided (around 100k reads).

tools/hicup/hicup_mapper.xml
```diff
@@ -4,55 +4,81 @@
...
         #end if
-        hicup_mapper
-        --index '$index_path' 
-        --bowtie2 \$BOWTIE_PATH_BASH
 
-        $input_first_sequence $input_second_sequence
+        ##Dealing with fastq and fastq.gz
+        input1='$input_first_sequence' &&
+        input2='$input_second_sequence' &&
+        trunc_result_1=\$(echo '$input_first_sequence' | rev | cut -d'/' -f1 | rev) &&
```
bgruening: This is just to create a filename, right? Does it need to be a filename? Can we use the element_identifier (esacped) https://github.com/galaxyproject/tools-iuc/blob/bbb2e6b6769b03602a8ab97001f88fbec52080a1/tool_collections/fastx_toolkit/fastq_quality_boxplot/fastq_quality_boxplot.xml#L9

This would take the name from the history.

lldelisle: The first part (definition of input1 and input2) is needed because if the input files of hicup_mapper does not ends with .gz it is not able to open it correctly.
The second part (trunc_result_1) is to get the file name in order to move the svg to dataset1.mapper_barchart.svg or dataset2.mapper_barchart.svg to know which plot correspond to which mate.
You are right, I think, we could use $input_first_sequence.element_identifier.

lldelisle: Do you want me to do a new PR trying to simplify the code?

bgruening: If you like. I don't want to stress to you anymore, you did already such a great job here ... but I guess it would be a nice enhancement.

lldelisle: In fact element_identifier give the name in the history. What I would like to get is the name of the file, for example dataset_62467.dat.
Do you know which attribute is it?

lldelisle: I found: dataset_id

bgruening: Why do you want this? If it can be an arbitrary name we can also just use `./input1.fastq` and `./input2.fastq` and hardcode the names.

lldelisle: Yes I realized, after...
I am doing it.

bgruening:
Thanks a bunch @lldelisle! That was fantastic!

bgruening:
I updated the tool on usegalaxy.eu! Enjoy!

lldelisle:
Thanks

PR #656, opened 2017-11-12
Gotohscan & compalign

bgruening:
ping @HassanAmr


bgruening:
Tests have passed but are not propagated back to github.

PR #658, opened 2017-11-14
upgrade graphclust wrapper depen

mmiladi:


bgruening:
Macros can help with such PRs a lot. I might look at this ...

mmiladi:
I have to check it locally first. Linting also fails. Which kind of macros do u mean?

mmiladi:
PR seems to be ready for merge. If all the functional test with the docker instance went fine. For the next update I can make a common macro file. For the moment having them separated would be better for testing. Thanks.

bgruening:
@mmiladi it was working before the new version comes in.
@mmiladi your dev workflow seems to be too complicated. Do you know that you can use the local conda package, which you have locally build for testing with planemo? No need to create any PR.

bgruening:
Nice! Thanks a lot!

mmiladi:
Thank you!

PR #660, opened 2017-11-17
Condafy glimmer3

abretaud:
Here's the condafication of glimmer3
I didn't unblacklisted because:
 - I didn't touch the tools in additional -> are they supposed to work? I didn't find them on the toolshed
 - there's a problem in glimmer_w_icm.xml with the icm_input type which is 'data'. Is there a valid datatype for binary data?


bgruening:
Hi @abretaud! They are supposed to work, but are really old. 2013. They are working on our instance but need some love. The repo is here: https://toolshed.g2.bx.psu.edu/view/bgruening/glimmer3/b1ad88bbc5fa (with all tools). This predates the suites.

We should define a new datatype or simply use `tar` for the time being. Do you have time to work on this to get it under testing? I'm also fine to move them to IUC if this helps.

Thanks!

abretaud:
Well my first intention was to do just what's needed to fix the install in gga, but while I'm at it I'd rather finish the work!
I should be able to do it in the coming days hopefully

bgruening:
@abretaud we can close this now right? Do you mind removing the wrappers and adding a readme file pointing to the IUC once?

abretaud:
Done :)

bgruening:
Awesome!

PR #664, opened 2017-11-24
Hicup

lldelisle:
I used the --zip option and thus removed all the samtools view from the previous version.
This tools will generate unsorted bam. I hope it is not a problem.


tools/hicup/hicup_macros.xml
```diff
@@ -6,6 +6,7 @@
             <requirement type="package" version="2.2.6">bowtie2</requirement>
             <requirement type="package" version="1.2">samtools</requirement>
             <requirement type="package" version="0.13.1">docutils</requirement>
+            <requirement type="package" version="3.1.4">R</requirement>
```
bgruening: hicup already depends on R here, so this is not necessary https://github.com/bioconda/bioconda-recipes/blob/master/recipes/hicup/meta.yaml#L15

Or do you need this old R version?

tools/hicup/hicup_hicup.xml
```diff
@@ -0,0 +1,178 @@
...
+            <param name="re1" value="A^AGCTT"/>
+            <param name="input_files_digest" value="chr1.fa"/>
+            <param name="genome" value="human"/>
+            <conditional name="reference_genome">
+                <param name="source" value="history" />
+                <param name="own_file" value="chr1.fa"/>
+            </conditional>
+
+            <!-- outputs -->
+            <output name="hicup_results" file="results.html" ftype="html" lines_diff="10000"/>
```
bgruening: If we really want to keep this tool, we should also keep the test file. This was removed in this PR.
Or do we want to remove this tool altogether?

tools/hicup/hicup_macros.xml
lldelisle: When I do not put this line, I have an error:
Error in svg(file = outputfilename) : X11 is not available
Execution halted
Could not produce hicup_mapper summary bar chart: /slipstream/galaxy/production/data/jobs/027/27529/conda-env/bin/Rscript /slipstream/galaxy/production/data/jobs/027/27529/conda-env/share/hicup-0.5.10-0/r_scripts/hicup_mapper_summary.r ./ ./hicup_mapper_summary_09-56-19_24-11-2017.txt: Inappropriate ioctl for device at /slipstream/galaxy/production/data/jobs/027/27529/conda-env/bin/hicup_mapper line 232.
Thanks to the touch lines in the wrapper, the job is not failing but I have no plot.

But probably I can use another way.

tools/hicup/hicup_hicup.xml
lldelisle: Sorry I thought I put it. I will put it back.
I think it is better to keep this tool.

tools/hicup/hicup_macros.xml
bgruening: I wonder why this is the case. `<requirement type="package" version="3.1.4">R</requirement>` is not even a conda package it should be `<requirement type="package" version="3.1.4">r</requirement>`

Either way, if you really want to include this all-in-one tool lets create a test for these outputs and remove the touch lines. I will care then about the R failure.

tools/hicup/hicup_hicup.xml
lldelisle: The result.html is in my github. Do I need to do a new pull request or how can I do?

tools/hicup/hicup_macros.xml
lldelisle: The touch lines are also needed for the hicup_mapper and hicup_truncater if the R version does not manage to plot the svg...

tools/hicup/hicup_hicup.xml
bgruening: No, perfect. If you look in a little bit down you see your new commits and the added file.

tools/hicup/hicup_macros.xml
bgruening: We should ensure that R can plot it no matter what and that we have tests that ensure it. This way we don't need the touch commands.
Can you add tests for the SVGs, only a few no need for all.

lldelisle: I do not know why but when I put 
<requirement type="package" version="3.1.4">r</requirement>
I get: 
Detecting R automatically
Found R at '/slipstream/galaxy/production/data/jobs/027/27574/conda-env/bin/R'
Error in svg(file = outputfilename) : X11 is not available
Execution halted
Could not produce hicup_truncater summary bar chart: /slipstream/galaxy/production/data/jobs/027/27574/conda-env/bin/Rscript /slipstream/galaxy/production/data/jobs/027/27574/conda-env/share/hicup-0.5.10-0/r_scripts/hicup_truncater_summary.r ./ ./hicup_truncater_summary_16-59-05_24-11-2017.txt: Inappropriate ioctl for device at /slipstream/galaxy/production/data/jobs/027/27574/conda-env/bin/hicup_truncater line 443.

lldelisle: > Can you add tests for the SVGs, only a few no need for all.
I do not see what I should do to generate these tests.

bgruening: Mh, this is so wired. I will see if I can reproduce this error on travis.
Thanks a lot. I will merge this now and proceed on the other branch!

PR #669, opened 2017-12-18
Improve  RNAfold MEA handling and more tests

mmiladi:


mmiladi:
This is a followup to PR https://github.com/bgruening/galaxytools/pull/640
The current version always produces MEA output format. However it might break depending tools expecting the default cmd binary behavior, i.e. single MFE structure by default.

tools/rna_tools/vienna_rna/rnafold.xml
```diff
@@ -31,15 +31,19 @@
     #if $layout_type ==0
         --layout-type=$general_options.layout_type
     #end if
-    #if $mea:
-        --MEA=$mea
+    
+	#if $mea:		 
+        $mea=$meagamma
```
bgruening: Why this change? The old one called $mea was for the gamma value, isn't it?

mmiladi: $mea is selective, meaning user wants to use mea mode instead of default mode. mea mode has a gamma parameter that the user can modify. The old one always invokes mea mode, which is not the default binary behavior.

mmiladi:
Work In Progress

tools/rna_tools/vienna_rna/rnafold.xml
bgruening: We could make the $mea (gamma) float just optional, so if this value is set, then the mode is choosen.

mmiladi:
In pf mode, matrix ps outputs are shown both inside the collection ans also as single output entries. I don't know what causes them appear in the root history. @bgruening do u have any idea?
image

bgruening:
Is this on dev?

mmiladi:
Nope. RnaLfold is not touched but failing...

bgruening:
Thanks @mmiladi!

PR #678, opened 2018-01-10
added ribodiff

HassanAmr:


HassanAmr:
@bgruening,

I still can't figure out why the bioconda package produces pdf files when I run it on machine, while travis and planemo test don't (problem also occurs locally). But other than the pdf problem, the tool works.

tools/ribodiff/ribodiff.xml
```diff
@@ -0,0 +1,157 @@
...
+        #if str($cutofffdr):
+            -q $cutofffdr
+        #end if
+        #if $dispdiff:
+            -d $dispdiff
+        #end if
+        #if $multitest:
+            -m $multitest_option
+        #end if
+        && mv *EmpDisp.scatter* EmpDisp.scatter.pdf
```
bgruening: It seems that these files could not be found. 
https://travis-ci.org/bgruening/galaxytools/jobs/334979039#L3381

HassanAmr: Yes, that's the problem with the pdf output currently. The tool doesn't generate any pdf files. Even when I do it with planemo. The Bioconda package works as expected. So, I can either remove the pdf feature for now and the tool will work. Or fix the pdf output problem. But I am stuck at that. Don't know what it could be.

PR #684, opened 2018-02-14
Peakachu

TorHou:
One could add more help. But I'm not an expert on these parameters, so ... it works and could be improved if anybody aside from us wants to use it.


bgruening:
.shed.yml file is missing

TorHou:
ugh.. sure

bgruening:
@TorHou how could the test pass, with this bug? Can we add it test that is catching those bugs ;)

TorHou:
It occurs when there are more than one experimental and one control library .. The test case has one e-lib and one c-lib .. that's why it worked there

bgruening:
Danke!

PR #685, opened 2018-02-14
Final touches

TorHou:


bgruening:
@TorHou ping

tools/peakachu/peakachu.xml
```diff
@@ -101,11 +103,20 @@
...
-                    <option value="manual">Manual</option>
-                    <option value="none">None</option>
-                </param>
+                <conditional name="norm_method">
+                    <param name="norm_method_selector" type="select" label="Normalisation Method.">
+                        <option value="deseq" selected="True">DESeq2</option>
+                        <option value="manual">Manual</option>
+                        <option value="none">None</option>
+                    </param>
+                    <when value="manual">
```
bgruening: was passiert bei DESeq2 and None?

tools/peakachu/peakachu.xml
```diff
@@ -114,22 +125,26 @@
...
-                </param>
+                <conditional name="norm_method">
+                    <param name="norm_method_selector" type="select" label="Normalisation Method.">
+                        <option value="tmm" selected="True">TMM</option>
+                        <option value="deseq">DESeq2</option>
+                        <option value="count">Count</option>
+                        <option value="manual">Manual</option>
+                        <option value="none">None</option>
+                    </param>
+                    <when value="manual">
```
bgruening: and here as well, what happens in the other cases?
A <when/> is at least needed

TorHou: Then everything is done internally by Peakachu, you don't need to provide your own normalization factors

TorHou: when it's `None` it won't normalize the counts ... so no additional info is needed

bgruening: Then please include the closing when.

TorHou: I'm not sure I understand ... see line 142 ?

PR #688, opened 2018-02-22
evaluation custom path args

mmiladi:


tools/GraphClust/CollectResults/evaluation.py
```diff
@@ -4,59 +4,77 @@
...
 listOfClusters = []
 listOfHeaders = []
 headersNames = set()
-cluster_seqs_stats_path = "RESULTS/*.cluster.all"
-cluster_seqs_stats_files = glob.glob(cluster_seqs_stats_path)
+idsNames = set()
 
+
+names = os.listdir(results_dir)
+cluster_seqs_stats_files = fnmatch.filter(names, '*.cluster.all')
```
bgruening: @mmiladi please have a look at the [glob](https://docs.python.org/2/library/glob.html) module. I'm not sure if can sort the values, or if you really need them sorted, but glob might be useful for you in the future :)

tools/GraphClust/CollectResults/evaluation.py
```diff
@@ -4,59 +4,77 @@
...
         if len(splits2) >= 6:
             fullHeader = splits2[5]
             headersNames.add(fullHeader)
+            fullID = splits2[3]
+            idsNames.add(fullID)
 
 blackList = []
 numberOfClusters = 0
 for singleFile in sorted(cluster_seqs_stats_files):
+    singleFile = results_dir+'/'+singleFile
```
bgruening: os.path.join()

tools/GraphClust/CollectResults/evaluation.py
```diff
@@ -72,10 +90,12 @@ def sh(script):
 
     toWrite = "completeness_score : " + str(completeness_score) + "\n" + "homogeneity_score : " + str(homogeneity_score) + "\n" + "adjusted_rand_score : " +str(adjusted_rand_score)  + "\n" + "adjusted_mutual_info_score : " + str(adjusted_mutual_info_score)+ "\n" + "v_measure_score : " + str(v_measure_score)
 
+
 else:
     toWrite = "completeness_score : NA \nhomogeneity_score : NA \nadjusted_rand_score : NA \nadjusted_mutual_info_score : NA \nv_measure_score : NA"
 
-with open("RESULTS/evaluation.txt", "w") as fOut:
+print (toWrite)
```
bgruening: should be removed I guess.

tools/GraphClust/CollectResults/evaluation.py
```diff
@@ -72,10 +90,12 @@ def sh(script):
...
+with open(results_dir+"/evaluation.txt", "w") as fOut:
```
bgruening: os.path.join()

tools/GraphClust/CollectResults/glob_report.xml
```diff
@@ -73,7 +73,7 @@
         #end if
 
         &&
-        python '$__tool_directory__/evaluation.py'
+        python '$__tool_directory__/evaluation.py' FASTA/ RESULTS/
```
bgruening: I'm wondering why you don't need to update your tests ;)
Maybe you should add a test for this?

tools/GraphClust/CollectResults/evaluation.py
mmiladi: Of course, `glob` is the first choice, here it didn't work for me. It fails if directory name contains special character. E.g. with a downloaded file from galaxy `Galaxy282-[RESULTS.zip]`, [] are special characters.

tools/GraphClust/CollectResults/glob_report.xml
mmiladi: Functionality is the same :) Before the folders were hard coded to be `FASTA/ and RESULTS/` . Now it's passed to the script, so I can use same script for local tests, with custom directory names.

mmiladi: Hmm the output is tested here, but stays the same :)

mmiladi: RESULTS/evaluation.txt

tools/GraphClust/CollectResults/evaluation.py
mmiladi: fixed

tools/GraphClust/CollectResults/evaluation.py
mmiladi: fixed

tools/GraphClust/CollectResults/evaluation.py
mmiladi: jup. done

mmiladi:
Thanks @bgruening !

PR #693, opened 2018-03-02
Added EpiCSeg

wkopp:
Hi @bgruening,

as part of de.NBI  I have established a Galaxy wrapper for [EpiCSeg](https://github.com/lamortenera/epicseg) -  a tool for conducting chromatin segmentation.

Best,
Wolfgang


bgruening:
@wkopp thanks a lot. Can you please add a .shed.yml file to add some meatadata for the repository and to trigger out CI tests. Example is here: https://github.com/bgruening/galaxytools/blob/master/tools/rna_tools/infernal/.shed.yml

Thanks.

tools/epicseg/epicseg_normalizecounts.xml
```diff
@@ -0,0 +1,66 @@
...
+            #end for#
+
+            Rscript '${__tool_directory__}/epicseg.R' normalizecounts
+                --nthreads $nthreads
+
+            #for $dataset in $datasets#
+                --counts `basename ${dataset.counts}`
+            #end for# &&
+
+            rename "s/.dat/.txt/" *_norm.dat &&
```
bgruening: You don't need to rename files. In Galaxy files do not have names and file endings.

tools/epicseg/epicseg_normalizecounts.xml
```diff
@@ -0,0 +1,66 @@
...
+            rename 's/^/norm_/' *_norm.txt;
+	    ]]></command>
+	    <inputs>
+		<repeat name="datasets" title="Count matrix">
+		    <param type="data" name="counts" label="Count matrix"
+                   format="txt"
```
bgruening: tabular?

tools/epicseg/epicseg_normalizecounts.xml
```diff
@@ -0,0 +1,66 @@
...
+
+                   help="Count matrix such as those produced by
+                         epicseg_getcounts. In case multiple count matrices
+                         are used, they must have the same marks and the same
+                         number of bins"/>
+		</repeat>
+		<param type="integer" name="nthreads"
```
bgruening: remove this one, a user does in general not know how many cores the cluster node has ...

tools/epicseg/epicseg_normalizecounts.xml
```diff
@@ -0,0 +1,66 @@
...
+                value="1" help="Number of threads to use."/>
+	    </inputs>
+	    <outputs>
+		  <data name="ncounts" format="txt">
```
bgruening: tabular?

tools/epicseg/epicseg_segment.xml
```diff
@@ -0,0 +1,127 @@
...
+        #end for#
+
+           --maxiter $maxiter
+
+        #for $annot in $annotation#
+           --annot ${annot.name}:${annot.annotation}
+        #end for# &&
+
+        mkdir -p '${ report.files_path }' &&
+        cp lmeans.txt '${ report.files_path }' &&
```
bgruening: you can use `from_work_dir` like here: https://github.com/galaxyproject/tools-iuc/blob/6236c2beed98e11da74605822fc27bd222354816/tools/roary/roary.xml#L74

This way you don't need to copy data in your tool.

tools/epicseg/epicseg_segment.xml
```diff
@@ -0,0 +1,127 @@
...
+                   help="Count matrix. If multiple count matrices are
+                         used, they must share the same number of
+                         chromatin marks and genomic regions."/>
+        </repeat>
+        <param type="integer" name="nstates" label="Number of states"
+               min="0" value="7"
+               help="Number of chromatin states."/>
+        <param type="integer" name="maxiter" value="200"
+               help="Maximum number of iterations in training mode."/>
+        <param type="integer" name="nthreads" value="1"
```
bgruening: remove this one.

wkopp:
Hi @bgruening,

thank you for your helpful comments, below you'll find how they were addressed:

- [x] The tool is now available as a single application
- [x] Requirements and version moved to macros.xml
- [x] nthreads replaced by GALAXY_SLOTS
- [x] tabular instead of txt: has become obsolete, because the count matrices are not returned to the user anymore.
- [x] reduced test dataset size

I'd like to set the tool default values for some of the options explicitly (e.g. mapq) on the command line, because there should be one instance of (e.g. --mapq <value>) per input file. While the command line tool also supports not having set these repeatable arguments at all or just setting them once for all files, 
dealing with this in the wrapper would make it, I believe unnecessarily, complicated.

If I understand correctly, 'from_work_dir' works fine to detect report.html, but the additional files which are referenced within the html are not detected if they are not moved to files_path. I have changed this part though to produce the referenced files directly into $report.files_path to avoid having to copy all referenced files explicitly.

Thank you!

tools/epicseg/epicseg.xml
```diff
@@ -0,0 +1,358 @@
...
+                    --mark '${mark.label}:${d}_${i}.bam' --mapq ${mark.mapq}
+                        ${mark.pairedend} --shift ${mark.shift}
+                #end for
+
+                &&
+        #end for
+
+        ## Normalize the datasets
+        #if len( $datasets ) == 1:
+            ln -s ${d}.tab ${d}_norm.tab
```
bgruening: @wkopp $d is not available here as you closed the for loop isn't it? So I assume it will be 0?

wkopp: $d actually still exists after the loop. So the statement indeed evaluates to `ln -s 0.tab 0_norm.tab` but I agree that this is confusing. I've changed that accordingly.

tools/epicseg/macros.xml
```diff
@@ -0,0 +1,26 @@
...
+  <xml name="requirements">
+      <requirements>
+          <requirement type="package" version="@VERSION_STRING@">epicseg</requirement>
+          <yield />
+      </requirements>
+  </xml>
+
+  <xml name="citations">
+    <citations>
+        <citation type="bibtex">
```
bgruening: You can actually give here a DOI when you use type doi.

tools/epicseg/.shed.yml
```diff
@@ -0,0 +1,14 @@
+name: epicseg
+owner: wkopp
```
bgruening: We do have a team called `rnateam` to we share the burdon of maintaining tools in de.NBI.
This owner will become a TS owner, so if you like but rnateam here :)

tools/epicseg/epicseg.R
```diff
@@ -0,0 +1,3 @@
+#!/usr/lib/R/bin/Rscript
```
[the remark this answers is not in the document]
bgruening: ping ...

wkopp:
I've fixed the remaining comments. Thank you for your help!

bgruening:
@wkopp top! It is now available under https://usegalaxy.eu.
Thanks a lot and greeting to the crowd in Berlin!

PR #694, opened 2018-03-07
Fastp - modifications

mblue9:
This PR has some suggested modifications for the fastp tool, currently in the process of being wrapped here: https://github.com/bgruening/galaxytools/pull/691. Feel free to disagree with any of them @HassanAmr, they're only pretty minor, it's great that this tool is almost wrapped!

I'll add some more modifications shortly, but so far this PR has:

- add output HTML report (currently optional but default set to True), with test
- add output JSON report
- make the Read 1 and Read 2 output names consistent
- add citation
- add version command
- fix ordering of param attributes


bgruening:
@mblue9 @HassanAmr great work. Feel free to submit it to IUC if you like.

PR #696, opened 2018-03-08
Methyldackel add tsv and bed support

bwlang:
This adds support for producing a tabular file containing the per-read methylation stats.
Also adds support for using bed files to specify regions for inclusion and fixes a few minor issues.

Maybe it's worth it to also change the name of the tool to avoid confusion?


tools/methyldackel/MethylDackel.xml
```diff
@@ -44,6 +44,9 @@
                 #if $advanced_options.mbias_regionString:
                     -r $advanced_options.mbias_regionString
                 #end if
+                #if $advanced_options.regionBed:
+                    -l $advanced_options.regionBed
```
bgruening: single quotes

bwlang:
missed a couple of test files.. whups.
i did mean the tool id... it's a bit confusing that the name is MethylDackel, but  you have to search for pileometh in the admin interfaces. I guess I see three options harmonize name and ID (breaking history) add new methyldackel tool or or live with the minor confusion for admins.

I'll leave that up to you  - it's not really very related to this PR

bgruening:
@dpryan79 any opinion on the renaming Q?

dpryan79:
We only kept pileometh in there for convenience. If it's no longer convenient then it's a good time to rename.

bgruening:
Thanks @bwlang!

PR #697, opened 2018-03-09
optionally break maf in pieces before Rnacode call

bernt-matthias:
Uses breakMAF.pl (from RNAcode package) to break the alignment into pieces. Unfortunately the perl script seems to be missing in the conda package. 

Of course we could also have a separate tool (but I don't see if splitting is interesting somewhere else).

### DIFF
--- a/tools/rna_tools/rnacode/rnacode.xml
+++ b/tools/rna_tools/rnacode/rnacode.xml
@@ -10,9 +10,9 @@
             '$__tool_directory__/breakMAF.pl'
             --maxlength $cond_breakmaf.maxlength
             --desiredLength $cond_breakmaf.desiredlength
-            < '$alignment' > aln.maf && 
+            < '$alignment' |  
         #else
-            ln -s '$alignment' aln.maf && 
+            cat '$alignment' | 
         #end if
 
         #if $cond_breakmaf.select_breakmaf == 'breakmaf' and $cond_breakmaf.processseparately == 'yes'
@@ -46,8 +46,6 @@
         #if $cond_scoringParameters.scoringParameters == 'custom'
             --pars "$pars"
         #end if
-        
-        aln.maf
         
         #if $outputFormat.value == '--tabular'
             --outfile $outFileDefault
--- a/tools/rna_tools/rnacode/rnacode.xml
+++ b/tools/rna_tools/rnacode/rnacode.xml
@@ -6,7 +6,7 @@
     <version_command>RNAcode --version</version_command>
     <command detect_errors="exit_code">
     <![CDATA[
-        #if $cond_breakmaf.select_breakmaf == 'breakmaf'
+        #if '$cond_breakmaf.select_breakmaf' == 'breakmaf'
             $__tool_directory__/breakMAF.pl
             --maxlength $cond_breakmaf.maxlength
             --desiredLength $cond_breakmaf.desiredlength
@@ -15,7 +15,7 @@
             ln -s '$alignment' aln.maf && 
         #end if
 
-        #if $cond_breakmaf.select_breakmaf == 'breakmaf' and $cond_breakmaf.processseparately == 'yes'
+        #if '$cond_breakmaf.select_breakmaf' == 'breakmaf' and $cond_breakmaf.processseparately == 'yes'
             $__tool_directory__/processMAF.sh
         #else
             RNAcode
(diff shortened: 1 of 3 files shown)


tools/rna_tools/rnacode/rnacode.xml
```diff
@@ -10,6 +10,15 @@
     <version_command>RNAcode --version</version_command>
     <command>
```
bgruening: Can we use `detect_errors` here and remove the stdio?

tools/rna_tools/rnacode/rnacode.xml
```diff
@@ -10,6 +10,15 @@
...
     <![CDATA[
+        #if $cond_breakmaf.select_breakmaf == 'breakmaf'
+            breakMAF.pl
+            --maxlength $cond_breakmaf.maxlength
+            --desiredLength $cond_breakmaf.desiredLength
+            < $alignment > aln &&
```
bgruening: alignment should be single quoted

bernt-matthias: sure

tools/rna_tools/rnacode/rnacode.xml
bernt-matthias: should be fine

bernt-matthias:
could you update the conda package

bgruening:
Yes I can, but this take some time :(

bernt-matthias:
Since breakMAF.pl did not help with the RNAcode bug (https://github.com/wash/rnacode/issues/8) I wrote a shell script that processes each part of split maf separately (which could also open some possibilities for parallel processing). I added the breakMAF.pl since it seems not to be included in the latest release of RNAcode, but is only available from github.

bernt-matthias:
I'm not so happy with creating a copy of the maf in the beginning of the script .. any thoughts on this. 

Also replacing '.' by '-' (https://github.com/wash/rnacode/issues/7) could be implemented easily.

tools/rna_tools/rnacode/rnacode.xml
```diff
@@ -1,17 +1,26 @@
...
-        <exit_code range=":-1" level="fatal" description="Error occurred. Please check Tool Standard Error"/>
-    </stdio>
     <version_command>RNAcode --version</version_command>
-    <command>
+    <command detect_errors="exit_code">
     <![CDATA[
-        RNAcode
-        
+        #if $cond_breakmaf.select_breakmaf == 'breakmaf'
+            $__tool_directory__/breakMAF.pl
```
bgruening: This one should be quoted.

tools/rna_tools/rnacode/rnacode.xml
```diff
@@ -1,17 +1,26 @@
...
+            --maxlength $cond_breakmaf.maxlength
+            --desiredLength $cond_breakmaf.desiredlength
+            < '$alignment' > aln.maf && 
+        #else
+            ln -s '$alignment' aln.maf && 
+        #end if
+
+        #if $cond_breakmaf.select_breakmaf == 'breakmaf' and $cond_breakmaf.processseparately == 'yes'
+            $__tool_directory__/processMAF.sh
```
bgruening: quoted

bgruening:
Which copy? Does a symlink work?

bernt-matthias:
I meant the 'copy' created by breakMAF.pl. I was thinking about using a named pipe. But I'm unsure how this needs to be used. I guess breakMAF would need to run in the background. I also don't know if RNAcode takes input from stdin...

tools/rna_tools/rnacode/rnacode.xml
bernt-matthias: @bgruening if I quote $cond_breakmaf.select_breakmaf then it does not work (i.e. the else path is used all the time). I guess you meant to quote something else?

bernt-matthias:
@bgruening another question calling the perl and bash scripts directly caused an error because execute rights are needed for others (our galaxy runs jobs as real user). should I always use `bash SCRIPT` or `perl SCRIPT` for the calls?

tools/rna_tools/rnacode/rnacode.xml
nsoranzo: @bernt-matthias You need to quote `$__tool_directory__/breakMAF.pl`, i.e. `'$__tool_directory__/breakMAF.pl'` , since `$__tool_directory__` may contain spaces.

bgruening:
> @bgruening another question calling the perl and bash scripts directly caused an error because execute rights are needed for others

This smells like chmod +x on the conda side?
Don't worry about the copying right now. The files are not that big isn't it.

bernt-matthias:
@bgruening we will apply it on the hg38 alignments with 20 or 99 other mammals .. quite large :)

bgruening:
Seriously? Crazy :)
But  I don't have a better option at hand.

bgruening:
Available from the toolshed and usegalaxy.eu.
Thanks!

PR #699, opened 2018-03-12
Link DOIs to preferred resolver

katrinleinweber:
Hello :-)

The DOI foundation recommends [this new, secure resolver](https://www.doi.org/doi_handbook/3_Resolution.html#3.8). Accordingly, this PR updates some static DOI links that still point to the old resolver.

Cheers!

### DIFF
--- a/tools/iprscan/interproscan.xml
+++ b/tools/iprscan/interproscan.xml
@@ -220,17 +220,17 @@ References
 Zdobnov EM, Apweiler R (2001)
 InterProScan an integration platform for the signature-recognition methods in InterPro.
 Bioinformatics 17, 847-848.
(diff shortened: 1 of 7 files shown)


bgruening:
Very cool Katrin.

I changed on of the doi occurrences to the actual abstracted DOI functionality in Galaxy. The resolving will then happen inside Galaxy.

You might also be interested in https://github.com/bioconda/bioconda-recipes/blob/00ee08eb57545fa90b4a12df06ef58eeb3b01e53/recipes/abyss/1.5.2/meta.yaml#L50 were we started to annotate BioConda recipes with dois.

katrinleinweber:
You're welcome, and thanks for merging :-)

PR #702, opened 2018-03-14
add header option for tabular data input

qiagu:
User is able to choose whether the input carry header or not. Support tabular data type only.


qiagu:
Thanks.

PR #703, opened 2018-03-14
add regression metrics tool

qiagu:
1. add regression_metrics tool based on clf_metrics tool
2. add header option in clf_inputs
3. make both clf_metrics and regresstion_metrics tools have header option


tools/sklearn/clf_metrics.xml
```diff
@@ -26,19 +26,21 @@ from sklearn import metrics
 input_json_path = sys.argv[1]
 params = json.load(open(input_json_path, "r"))
 
-
+header='infer' if params["clf_metrics"].get("header1", None) else None
```
bgruening: the else is not needed, isn't it?

bgruening:
Thanks!

PR #705, opened 2018-03-22
Intarna: Requirement version bumped to 2.1.0

martin-mann:
Seems I only changed the wrapper version last time but missed to bump the intarna tool version...


martin-mann:
mhh.. we could change the tests to "base-pair-based energy model" which should not show the value fluctuations caused by changes within the vienna rna package...

bgruening:
Stable tests are always welcome! For the time being I will merge this. Thanks.

PR #706, opened 2018-03-26
bump IntaRNA 2.2.0

martin-mann:
I will add additional arguments soon. But for now just bumping the version.


martin-mann:
mhh... seems the call itself is already wrong (and most likely was also for older versions):
```[bash]
IntaRNA --query '/tmp/tmpkOaPTB/files/000/dataset_1.dat' --qAcc C --qAccW 150 --qAccL 100  --target '/tmp/tmpkOaPTB/files/000/dataset_2.dat' --tAcc C --tAccW 150 --tAccL 100  --noSeed --seedBP 7 --seedMaxUP 0  --mode H  --out ./temp_tabular_output --outMode C --outNumber 1 --outOverlap Q
```
calls IntaRNA with `--noSeed` which should be per default off and *NOT* set! (even more so, since other seed parameters are present, ie here `--seedBP --seedMaxUP`)

but anyway strange that no output at all is generated.. will have a look.

martin-mann:
@bgruening mhh.. ' --noSeed' is still in the test call, what is wrong with the XML setup? any idea? or somebody to ask (beside you ;) )

bgruening:
I think I fixed it. booleans have a truevalue and a falsevalue argument so the `--seed` is not need, it will be used if the boolean is toggeld by the user.

martin-mann:
@bgruening 
no idea why the test fails. If I install IntaRNA-2.2.0 via bioconda locally and run the test's call I correctly get
```[bash]
(IntaRNA-2.2.0) -bash-4.1$ IntaRNA --query 'intarna_query.fa' --qAcc C --qAccW 150 --qAccL 100  --target 'intarna_target.fa' --tAcc C --tAccW 150 --tAccL 100 --seedBP 7 --seedMaxUP 0  --mode H  --outMode C --outNumber 1 --outOverlap Q --out ./temp_tabular_output && sed 's/\t/     /g' ./temp_tabular_output > ./temp_output_no_tabs && sed 's/;/\t/g' ./temp_output_no_tabs > 'dataset_3.dat'

(IntaRNA-2.2.0) -bash-4.1$ cat dataset_3.dat
id1     start1  end1    id2     start2  end2    subseqDP        hybridDP        E
mRNA1   85      95      ncRNA1  21      32      GUGGUGAGGAG&CUCCUCACACAC        (((((((((((&)))))))).)))        -11.8783
mRNA1   52      58      ncRNA2  28      34      UCAGUUA&UGACUGA (((((((&))))))) -3.08648
mRNA2   86      97      ncRNA1  20      31      UGUGUGACGAGU&ACUCCUCACACA       (((((((.((((&)))).)))))))       -8.16744
```

any further idea how to find out why the travis check of the wrapper fails?

bgruening:
Yes, but the old test data expected a different result. Can you check https://github.com/bgruening/galaxytools/pull/706/commits/b6d8f704e6a2c45a92975bff7a1f59675edd607e if it makes sense?

martin-mann:
the point is: it seems there is no output at all or do I get the travis log wrong?

martin-mann:
uaks.. yes, just realized I read it wrong all the time... thanks for doing the final step! *thumbsup*

martin-mann:
think I will push another IntaRNA-v2.1.0 version to bioconda with a recent VRNA, to correct the test problems for v2.1.0

martin-mann:
uaks.. that's not easily done right? to trigger a rebuild of a bioconda package for an older version?

bgruening:
You would need to add a subdir with the old version.
Are the old results so wrong? If so would it not be better to tell people to use the latest version instead of a newly build old version?

martin-mann:
at the least the results differ significantly compared to the previous (2.0.4) or subsequent version (2.2.0)

so most likely, I will build another 2.1.0 package

thanks for the subdir hint, guessed this might be the way..

as stated on top, will soon update the wrapper with additional arguments new to 2.2.0

thanks a lot for your help!

bgruening:
@martin-mann ping

martin-mann:
@bgruening no time this week.. :/ 

do you want to merge as it is, or do you want to wait for the additional stuff? not essential anyway..

so feel free to merge right away and I will open a new one for the additional args..

thanks

bgruening:
I thought you wanted at first to release a new version of 2.1.0 before we merge this one?

bgruening:
Thanks! Both version will be on usegalaxy.eu soon.

bgruening:
deployed ...

PR #707, opened 2018-03-28
Meth extractor raise fix

bwlang:
the raise syntax is invalid... fixes that.
Not sure about the version number bump... but usually i forget that;)


bgruening:
@bwlang you want to at the output fix here as well?

bwlang:
sure...

PR #718, opened 2018-04-26
Add a tool to handle scikit-learn model validation

qiagu:
The cross_validate outputs different test/score time for each run. How can I make a test to tolerate the time discrepancy?


tools/sklearn/model_validation.xml
```diff
@@ -0,0 +1,221 @@
...
+                    <expand macro="model_validation_common_options"/>
+                    <expand macro="scoring"/>
+                    <expand macro="pre_dispatch"/>
+                </section>
+            </when>
+        </conditional>
+        <expand macro="sl_mixed_input"/>
+    </inputs>
+    <outputs>
+        <data format="txt" name="outfile"/>
```
bgruening: the format can the `tabular` I think.

tools/sklearn/model_validation.xml
```diff
@@ -0,0 +1,221 @@
...
+import json
+import pandas
+import pickle
+import numpy as np
+import sklearn.model_selection
+from sklearn import svm, linear_model, ensemble
+
+import logging
+
+logging.basicConfig(format='%(asctime)s,%(msecs)d %(levelname)-8s [%(filename)s:%(lineno)d] %(message)s',
```
bgruening: Why do you need to change the defaults? And should this go to stdout instead? So the Galaxy user can see it in the end?

qiagu: You mean the logger?

bgruening: Yes.

qiagu: That's for debugging purpose. Will remove those lines finally. You don't a tool to track your code?

tools/sklearn/model_validation.xml
qiagu: Could you please give the details? Thanks.

tools/sklearn/model_validation.xml
bgruening: I use logging, I was just asking why you do modify the defaults and why you are not printing to stdout, this way you would also see it in Galaxy.

tools/sklearn/model_validation.xml
bgruening: The output file is a tabular isn't it? Its tab saparated.

tools/sklearn/model_validation.xml
qiagu: Sorry to misunderstand you. Seems that printing to stdout is a simple and clean way. Thanks for pointing that out.

tools/sklearn/model_validation.xml
qiagu: Oh, yes. I changed. I thought the tabular datatype could solve the diff problem as well.

qiagu:
I'm still working on the other tests. Seems no uniform output style can fit them all.

bgruening:
Thanks @qiagu!

PR #720, opened 2018-05-07
Bismark datatype, fixes upto and stdout

bwlang:
These fixes keep galaxy from trying to index the unsorted bam file produced by bismark.
Also fixes a stdout capture (w mode overwrites), and --upto support


bgruening:
cool, thanks.

PR #721, opened 2018-05-10
upgrades salmon, enables collections/interleaved fastq input

bwlang:
I used https://github.com/galaxyproject/tools-iuc/blob/master/tools/bowtie2/bowtie2_wrapper.xml as a model.
also added a few more tests

zcat < file works on stock mac os and linux


bgruening:
super cool!

PR #722, opened 2018-05-18
Adding new tool to text-format bam files and adding test cases

Florian-H-Lab:
Its simply a wrapper around a pipe of samtools view and awk, thus allowing directly to change the text of a bam file (i.e., output is again a bam file), without the need to convert the bam file into a sam file and into a bam file again.

Please check the, if the description of the tool is sufficient and valid.


Florian-H-Lab:
check failed because lint throws error when evaluating the cut.xml

PR #723, opened 2018-05-22
Updates for Scikit-learn feature selection tool: (1) set output datatype and (2) u…

jgoecks:
…se dataset headers.


bgruening:
Thanks @jgoecks. Looks good to me. I guess this will create a merge conflict now with @qiagu ones. Lets see.

PR #729, opened 2018-05-25
Hicup v6.1.0 and small additional changes

lldelisle:
I hope this time it is ok.
Sorry...


tools/hicup/hicup_mapper.xml
```diff
@@ -44,11 +69,11 @@
     </outputs>
      <tests>
         <test>
-            <param name="input_first_sequence" value="dataset1.trunc.fastq" ftype="fastq"/>
-            <param name="input_second_sequence" value="dataset2.trunc.fastq" ftype="fastq"/>
+            <param name="input_first_sequence" value="dataset1.trunc.fastq.gz" ftype="fastq.gz"/>
+            <param name="input_second_sequence" value="dataset2.trunc.fastq.gz" ftype="fastq.gz"/>
```
bgruening: These are binaries and we can not diff them. If we want to keep the compressed version we should use something like `sim_size`.

lldelisle: I do not know sim_size...

lldelisle:
I see that there is a problem with the new datatype because I made a test if $file.ext == 'bam' but if it is qname_sorted.bam the extension is not bam anymore. I am changing it to $file.ext is not 'sam'.

tools/hicup/hicup_mapper.xml
bgruening: Have a look here: https://docs.galaxyproject.org/en/latest/dev/schema.html#id84

bgruening:
> I see that there is a problem with the new datatype because I made a test if $file.ext == 'bam' but if it is qname_sorted.bam the extension is not bam anymore. I am changing it to $file.ext is not 'sam'.

Can you point me to this? where is it? You could  $file.ext == 'qname_sorted.bam'

lldelisle:
I do not know what happens if the new bam datatypes are not installed...

bgruening:
> I do not know what happens if the new bam datatypes are not installed...

Thats true, lets add a profile tag to the tools as described here: https://docs.galaxyproject.org/en/latest/dev/schema.html#attributes

lldelisle:
> Thats true, lets add a profile tag to the tools as described here: https://docs.galaxyproject.org/en/latest/dev/schema.html#attributes
I just tried on my local instance which is v17.09 and you cannot insert input as the qname_sorted.bam datatype does not exists.
So I will correct the code to change to $file.ext == 'qname_sorted.bam'

lldelisle:
In fact the hicup_hicup tool is working on my local instance v17.09. Then I need to change manually the datatype of the output, but I agree, it would be more secure to add the `profile="18.01"`.

bgruening:
Lets see what the main branch is saying :)

PR #732, opened 2018-05-28
ChIPseeker: Add option to use GTF for annotation, with tests

mblue9:
Ran into UCSC vs Ensembl naming issues ("chr1" vs"1") :( so have added in an option to use a GTF file for the annotation. This tool should work for any species now with a GTF.

I could remove the txdb and org.db packages to make this wrapper smaller, the only benefit of keeping them I think, is that you get the gene name added, but it only works for human and mouse and you can get that info from the annotatemyids tool. So should I remove the txdb and org packages from the requirements?


mblue9:
Don't know why I have these conflicts here, should I just delete this branch and make a new branch?

bgruening:
@mblue9 here is something wrong. Merge conflicts?

bgruening:
maybe ... or rebase against latest master?

bgruening:
> I could remove the txdb and org.db packages to make this wrapper smaller, the only benefit of keeping them I think, is that you get the gene name added, but it only works for human and mouse and you can get that info from the annotatemyids tool. So should I remove the txdb and org packages from the requirements?

If that is possible, I'm all in :)

mblue9:
Now this one keeps failing on me :(

Not sure, but think it might be this error here: https://travis-ci.org/bgruening/galaxytools/jobs/384993627#L1618

`migrate.exceptions.VersionNotFoundError: "Database schema file with version 141 doesn't exist."`

bgruening:
A mixture of everything :(
Not our month I guess .... is it now working locally for you again?

bgruening:
Updated the version on usegalaxy.eu

mblue9:
Yep working locally, and working well :) thanks!

I've made some small adjustments in the new PR here:https://github.com/bgruening/galaxytools/pull/735

One thing that doesn't work, and I don't why, is the bar plots are not produced in the pdf file, the plotAnnoBar and plotDistToTSS. They're not in the pdf file if I test locally with planemo test or from the toolshed wrapper. But they are in the pdf if I run the code in my local RStudio R and also if I run with R in the chipseeker env. No idea what's going on there.

PR #739, opened 2018-05-31
rnacode add forgotten redirect

bernt-matthias:
do not know why the tests were successful. but it definitely did not work on our instance. 

open: 
- [x]  the breakmaf perl script can only process maf. should we just forbid clustal as input or convert clustal -> maf before the processing starts. If the latter -> how?


bgruening:
sorry, I missed that somehow.

> the breakmaf perl script can only process maf. should we just forbid  clustal as input or convert clustal -> maf before the processing  starts. If the latter -> how?

lets not allow clustal in this tool. The conversion should happen on the Galaxy datatype level if possible. So Galaxy will auto-convert it before running this tool.

bernt-matthias:
done

bernt-matthias:
also sneaked something in. I'm happy now with this PR.

bgruening:
Thanks!

PR #741, opened 2018-06-04
Nanopolish 0.9.2

mmiladi:


tools/nanopolish/nanopolish_methylation.xml
```diff
@@ -39,6 +40,8 @@
 
         <param argument="-w" type="text" optional="true"
             label="find variants in window of region chromsome:start-end" />
+        
+        <param name="batchsize" type="integer" optional="true" value="512" label="the batch size" help="(-K)"/>
```
bgruening: if you set optional=true, I would use value="" and include `#if str($batchsize)` in the command section.

tools/nanopolish/nanopolish_variants.xml
```diff
@@ -98,6 +99,7 @@
         <param name="max_haplotypes" type="integer" optional="true" value="1000" label="Consider at most N haplotype combinations" help="(-x)"/>
         <param name="max_rounds" type="integer" optional="true" value="50" label="Perform N rounds of consensus sequence improvement" help="(--max_rounds)"/>
         <param name="ploidy" type="integer" optional="true" value="-1" label="The ploidy level of the sequenced genome. -1:disabled" help="(-p)"/>
+        <param name="min_flanking_sequence" type="integer" optional="true" value="30" label="distance from alignment end to calculate variants" help="(--min-flanking-sequence)"/>
```
bgruening: same as above

tools/nanopolish/nanopolish_methylation.xml
mmiladi: So you mean, it's preferred to not pass optional params with their default values?

bgruening: It should be fine, but 1) if the user removes everything then the value is "" and the command crashes, because you do not catch it and 2) if upstream changes the default value at some point you need to update the default value in your case. If you don't hardcode it, you don't need to change it :)

tools/nanopolish/nanopolish_eventalign.xml
```diff
@@ -56,6 +57,7 @@
         <!-- optional params -->
         <param argument="-w" type="text" optional="true"
             label="find variants in window of region chromsome:start-end" />
+        <param name="min_mapping_quality" type="integer" optional="true" value="0" label="only use reads with mapping quality at least" help="(-q)"/>
```
bgruening: here as well :)

mmiladi: :))

mmiladi:
The PR should now have covered the missing features

PR #743, opened 2018-06-05
added oom detection for diamond

bernt-matthias:


bgruening:
huh? `/tmp/tmpIcHkfp:7:0:ERROR:SCHEMASV:SCHEMAV_CVC_DATATYPE_VALID_1_2_1: Element 'regex', attribute 'level': 'fatal_oom' is not a valid value of the atomic type 'LevelType'.]`

bernt-matthias:
maybe related: https://github.com/galaxyproject/planemo/issues/844

bgruening:
Mh, we are always using the latest planemo here isn't it? How can it pass locally for you? Either way I trust you :)

PR #744, opened 2018-06-05
add all-columns-but-by-index-number option to sklearn column selector

qiagu:
Imagine there are hundreds or even thousands of features in the input dataset (e.g. RNAseq data) and only several columns, either index or labels, need be excluded from the column selection. This PR works perfectly for this kind of scenario.


bgruening:
Cool, cool!

PR #747, opened 2018-06-07
Diamond update

bernt-matthias:
- [x] update to 0.9.21
- [x] added new parameters (eg taxonomic filtering)
- [x] Should we remove the `--sall*` parameters? According to the pdf manual the -sall... parameters apply for the DAA output only which is not (yet) supported by the wrapper. 
  - moved them to the output options where they seem to work
- [x] The different defaults of the gap parameters could be solved by a conditional for the matrix.
  - solution: gap parameters optional (defaults are chosen via substitution matrix)
- [x] more tests?

Note, appart from the `--min-orf` parameter the tool still ignores all advanced parameters.

The modes getseq (Retrieve sequences from a DIAMOND database file) and dbinfo (Print information about a DIAMOND database file) are also not included in the wrapper. 

Any comments? Happy to finish this soon. Will start testing on real data on our instance soon.


bgruening:
DAA output would actually really nice for the metaproteomic projects ... 
More tests are always good, but every enhancement is good, so we can leave this for later, you are probably mega busy.

bernt-matthias:
The failing check is somehow related to planemo not knowing about the `fatal_oom` level. Should work? 

DAA is good idea. Started working on it, see also https://github.com/galaxyproject/galaxy/pull/6313.

struggling with the tests of makedb. Don't understand why they are failing. Even the output of the hash seems correct (Database hash = 5d34bf6fda41b847c0cf6ff3130c1ec4) .. should we just look for this line? few more tests are coming.

bernt-matthias:
btw. can I easily test against https://github.com/galaxyproject/galaxy/pull/6313?

bernt-matthias:
Do we want to have multiple output formats at the same time: https://github.com/bbuchfink/diamond/issues/218#issuecomment-396273715?

bernt-matthias:
One more question: How (in which tools) would you like to  integrate diamonds daa output?

bgruening:
> Do we want to have multiple output formats at the same time: bbuchfink/diamond#218 (comment)?

not really needed

> One more question: How (in which tools) would you like to integrate diamonds daa output?

I think the diamond search tool.

bernt-matthias:
Just added `diamond view`. Would like to test it against https://github.com/galaxyproject/galaxy/pull/6313 before merging..

bernt-matthias:
test against https://github.com/galaxyproject/galaxy/pull/6313 was successful. remains only the problem with fatal_oom

bgruening:
This one `bg_diamond_view[0]: failed` is failing for me locally. Seems like no output is produced.

bgruening:
datatypes are not backported, but could be by admins. The tool will still work, even if it does not know the datatype afaik.

bgruening:
where was the typo?

bernt-matthias:
The OUTPUT_ARGS token had a `#els if` .. missing an `e`.. 

So it seems that planemo does not lint the command?

bgruening:
No, we should, but planemo does not currently.

bgruening:
All passed for me locally. Ready if you are ready.

bernt-matthias:
ready

bgruening:
Cool thanks!

PR #752, opened 2018-06-21
Update RCAS to current version 1.5.4. 

michauhl:
Hi,
I updated the RCAS wrapper (1.1.1) to the current version (1.5.4). This version now supports multi set analysis (in addition to single set as before), therefore the major changes in the wrapper. I think there is still an issue if you select just one set in multi set analysis, the run will fail. Maybe we can capture this (or leave it since normally you would select 2 or more anyway when selection multi set analysis).
Best,
Michael


bgruening:
Cool, thanks!

PR #754, opened 2018-07-01
add column selector options to handle header name

qiagu:
Add two options "Select columns by column header name(s)"  and "All columns but by column header name(s)". We should be able to solve the header name selection issue.


bgruening:
Magic!

PR #762, opened 2018-07-10
add header option to pre-processing and do pre-processing in cross validation

qiagu:
add header/column selector options to pre-processing tool.
mv pre-processing UI to macro.
make pre-processing possible in pipeline of model validation tool.
change test output style.


qiagu:
Thanks

Issue #756, opened 2018-07-01
Galaxy-Scikit Learn paper


jgoecks:
@bgruening @qiagu 

To publish I suggest that we perform two analyses:

1. Reproduce the work in [this paper](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC5890912/). This is a large analysis that will exercise much of the Sk-Learn functionality in Galaxy. Here are [datasets](https://github.com/EpistasisLab/penn-ml-benchmarks) and [scripts/Jupyter notebooks](https://github.com/rhiever/sklearn-benchmarks) for this work.

2. Perform an analysis that uses genomic analysis pipelines followed by Sk-Learn tools. One idea is doing drug response prediction using cancer cell lines from [DepMap](http://depmap.org/). If we were to do all of DepMep (1000+ cell lines) it would be a very large analysis; we could focus on a particular cohort (tissue), such as breast cancer cell lines. Another idea is to look at reproducing something in [Kipoi](http://kipoi.org/), which is focused on functional genomics.

bgruening:
Thanks for the summary @jgoecks!

qiagu:
The #1 is the same as the one I shared with @bgruening 2 days ago.

qiagu:
Things need to be done in my mind:
1) introduce the new dataset input system (including the column selector, header options and so on) to the old tools.
2) add pre-processing module into the pipeline of cross-validation
3) improve the UI of param_grid input box
4) make random search cross-validation possible
5) training materials

bgruening:
:+1: !!!

jgoecks:
@qiagu 

Here's my current list:

- [x] introduce the new dataset input system (including the column selector, header options and so on) to the old tools.
- [x] add pre-processing module into the pipeline of cross-validation
- [x] improve the UI of param_grid input box
- [ ] add `random_state` parameter for estimators to improve reproducibility
- [ ] make random search cross-validation possible training materials
- [ ] add reserve index option to the Load a model and predict section, maybe give a column name to prediction result as well.

qiagu:
The current tool set works mostly fine with the Penn datasets. For further improvement, below is my new work list:

- [ ] n_jobs vs. GALAXY_SLOTS
- [ ] make scoring selector, make_scorer possible (hard)
- [ ] more scipy distribution support, change symbol table (need security inspection)
- [ ] change safe_eval function to class for code efficiency
- [ ] search param text file support
- [ ] general sklearn object eval support
[...]

PR #763, opened 2018-07-12
Add search cv tool and pipeline tool

qiagu:
UI only, but please let me know whether you like this or not.  @jgoecks @bgruening

### DIFF
--- a/tools/sklearn/search_model_validation.xml
+++ b/tools/sklearn/search_model_validation.xml
@@ -1,4 +1,4 @@
-<tool id="sklearn_searchcv" name="Search for Optimal Hyperparameters" version="@VERSION@">
+<tool id="sklearn_searchcv" name="Hyperparameter Search" version="@VERSION@">
     <description>using exhausitive or randomized search</description>
     <macros>
         <import>main_macros.xml</import>
@@ -42,6 +42,9 @@ infile2 = "$input_options.infile2"
 infile_pipeline = "$search_schemes.infile_pipeline"
 outfile_result = "$outfile_result"
 outfile_estimator = "$outfile_estimator"
+#if $search_schemes.selected_search_scheme == "RandomizedSearchCV":
+np.random.seed($search_schemes.random_seed)
+#end if
 
 params_builder = params['search_schemes']['search_params_builder']
 
@@ -120,6 +123,7 @@ with open(outfile_estimator, "wb") as output_handler:
                 </section>
             </when>
             <when value="RandomizedSearchCV">
+                <param name="random_seed" type="integer" value="65535" min="0" max="65535" label="Set up random seed:"/>
                 <expand macro="search_cv_estimator"/>
                 <section name="options" title="Advanced Options for SearchCV" expanded="false">
                     <expand macro="search_cv_options"/>
@@ -208,6 +212,10 @@ with open(outfile_estimator, "wb") as output_handler:
                 <param name="search_p" value="gamma: np_random_uniform(low=0., high=1., size=2)"/>
                 <param name="selected_param_type" value="final_estimator_p"/>
             </conditional>
+            <conditional name="search_param_selector">
+                <param name="search_p" value="random_state: [324089]"/>
+                <param name="selected_param_type" value="final_estimator_p"/>
+            </conditional>
             <param name="infile1" value="regression_X.tabular" ftype="tabular"/>
             <param name="header1" value="true" />
             <param name="selected_column_selector_option" value="all_columns"/>
@@ -216,7 +224,7 @@ with open(outfile_estimator, "wb") as output_handler:
             <param name="selected_column_selector_option2" value="all_columns"/>
             <output name="outfile_result" >
                 <assert_contents>
-                    <has_n_columns n="14" />
+                    <has_n_columns n="15" />
                     <has_text text="param_estimator__max_depth"/>
                 </assert_contents>
             </output>
@@ -240,7 +248,8 @@ with open(outfile_estimator, "wb") as output_handler:
             <param name="selected_column_selector_option2" value="all_columns"/>
             <output name="outfile_result">
                 <assert_contents>
-                    <has_text_matching expression=".+0.05363984674329502[^/d]+200[^/d]" />
+                    <has_n_columns n="13"/>
+                    <has_text text="0.05363984674329502"/>
                 </assert_contents>
             </output>
         </test>
@@ -266,6 +275,10 @@ with open(outfile_estimator, "wb") as output_handler:
                 <param name="search_p" value="n_estimators: [10, 50, 200, 1000]"/>
                 <param name="selected_param_type" value="final_estimator_p"/>
             </conditional>
+            <conditional name="search_param_selector">
+                <param name="search_p" value="random_state: [324089]"/>
+                <param name="selected_param_type" value="final_estimator_p"/>
+            </conditional>
             <param name="infile1" value="regression_X.tabular" ftype="tabular"/>
             <param name="header1" value="true" />
             <param name="selected_column_selector_option" value="all_columns"/>
@@ -274,8 +287,8 @@ with open(outfile_estimator, "wb") as output_handler:
             <param name="selected_column_selector_option2" value="all_columns"/>
             <output name="outfile_result">
                 <assert_contents>
-                    <has_n_columns n="12"/>
-                    <has_text_matching expression=".+{u'estimator__n_estimators': 1000}[^/w]" />
+                    <has_n_columns n="13"/>
+                    <has_text_matching expression=".+0.7772355090078996[^/w]+1000[^/d]" />
                 </assert_contents>
             </output>
         </test>
@@ -284,6 +297,10 @@ with open(outfile_estimator, "wb") as output_handler:
             <param name="infile_pipeline" value="pipeline07"/>
             <conditional name="search_param_selector">
                 <param name="search_p" value="n_estimators: [10, 50, 100, 200]"/>
+                <param name="selected_param_type" value="final_estimator_p"/>
+            </conditional>
+            <conditional name="search_param_selector">
+                <param name="search_p" value="random_state: [324089]"/>
                 <param name="selected_param_type" value="final_estimator_p"/>
             </conditional>
             <conditional name="search_param_selector">
@@ -298,8 +315,8 @@ with open(outfile_estimator, "wb") as output_handler:
             <param name="selected_column_selector_option2" value="all_columns"/>
             <output name="outfile_result">
                 <assert_contents>
-                    <has_n_columns n="13"/>
-                    <has_text_matching expression=".+{u'preprocessing_1__gamma': 1.0, u'estimator__n_estimators': 50}[^/w]" />
+                    <has_n_columns n="14"/>
+                    <has_text_matching expression=".+0.05747126436781609[^/d]" />
                 </assert_contents>
             </output>
         </test>
@@ -308,6 +325,10 @@ with open(outfile_estimator, "wb") as output_handler:
             <param name="infile_pipeline" value="pipeline08"/>
             <conditional name="search_param_selector">
                 <param name="search_p" value="n_estimators: [10, 50, 100, 200]"/>
+                <param name="selected_param_type" value="final_estimator_p"/>
+            </conditional>
+            <conditional name="search_param_selector">
+                <param name="search_p" value="random_state: [324089]"/>
                 <param name="selected_param_type" value="final_estimator_p"/>
             </conditional>
             <conditional name="search_param_selector">


jgoecks:
Great start, +1 in general. Some specific improvements that would be nice:

- need to output model from pipeline, not tabular file
- for pipeline tool: if estimator not on the list, provide option to either enter text or select from common estimators + their options. We should use macros to make the latter easy to implement.
- for search model validation tool: in addition to CV results, best estimator/pipeline should be an output (see `refit` parameter)

qiagu:
> need to output model from pipeline, not tabular file

Good catch.


> for pipeline tool: if estimator not on the list, provide option to either enter text or select from common estimators + their options. We should use macros to make the latter easy to implement.

This should be a huge improvement. I like it a lot. Throwing all model options into macro is not easy though.

> for search model validation tool: in addition to CV results, best estimator/pipeline should be an output (see refit parameter)

Current output options:
```
<param name="return_type" type="select" label="Select a return type">
        <option value="cv_results_" selected="true">cv_results_</option>
        <option value="best_estimator_">best_estimator_</option>
        <option value="best_score_">best_score_</option>
        <option value="best_params_">best_params_</option>
        <option value="best_index_">best_index_</option>
</param>
```
`best_estimator_` is already there. Is it the one you talked about?

jgoecks:
@qiagu Sorry I missed the best estimator. Let's make sure to provide option or automatically run `refit` As we discussed, would be nice to be able to select multiple options. Often I want to see `cv_results_` and get the best estimator.

qiagu:
@jgoecks  `select multiple options` is good. Thanks for pointing that out.

tools/sklearn/main_macros.xml
```diff
@@ -104,6 +101,77 @@ def get_X_y(params, file1, file2):
...
+
+    search_params = {}
+
+    for ep in params_builder['estimator_param_set']:
+        if ep['estimator_param'].strip() == '':
+            continue
+        lst = ep['estimator_param'].split(":")
+        if len(lst) >2:
+            exec lst[2].strip()
+        literal = lst[1].strip().replace("__sq__","'")\
```
bgruening: You can avoid this replacing  by allowing the parameter to use these characters with a sanatizer.

tools/sklearn/main_macros.xml
```diff
@@ -104,6 +101,77 @@ def get_X_y(params, file1, file2):
...
+    if not in_pipeline:
+        return search_params
+    
+    for ep in params_builder['prep_param_set']:
+        if ep['prep_param'].strip() == '':
+            continue
+        lst = ep['prep_param'].split(":")
+        if len(lst) >2:
+            exec lst[2].strip()
+        literal = lst[1].strip().replace("__sq__","'")\
```
bgruening: here as well, if you don't want to use a sanatizer, this should be a python function as you use it 3 times

tools/sklearn/main_macros.xml
```diff
@@ -1049,7 +1117,7 @@ def get_X_y(params, file1, file2):
   </xml>
 
   <xml name="feature_selection_estimator">
-    <param argument="estimator" type="select" label="Select an estimator" help="The base estimator from which the transformer is built.">
+    <param argument="estimator" type="select" label="Select an estimator:" help="The base estimator from which the transformer is built.">
```
bgruening: label should not have a `:` in the end, this is added by Galaxy

tools/sklearn/main_macros.xml
```diff
@@ -1070,7 +1138,7 @@ def get_X_y(params, file1, file2):
     <when value="yes">
     </when>
     <when value="no">
-      <param name="new_estimator" type="text" value="" label="Make a new estimator" />
+      <param name="new_estimator" type="text" value="" label="Make a new estimator:" />
```
bgruening: no `:`

tools/sklearn/main_macros.xml
```diff
@@ -1105,6 +1173,47 @@ def get_X_y(params, file1, file2):
...
+                </repeat>
+            </section>
+        </when>
+    </conditional>
+  </xml>
+
+  <xml name="search_cv_options">
+      <expand macro="scoring"/>
+      <expand macro="model_validation_common_options"/>
+      <expand macro="pre_dispatch" value="2*n_jobs" help="Controls the number of jobs that get dispatched during parallel execution"/>
```
bgruening: pre_dispatch --> I'm not sure this should be exposed to the user. The user has no clue on which cluster is is scheduled how much nodes you have available and so on ...

bgruening: We should hide this parameter tpye="hidden" und just hard code it ...

tools/sklearn/pipeline.xml
```diff
@@ -0,0 +1,130 @@
+<tool id="sklearn_pipeline" name="Build a Pipeline" version="@VERSION@">
```
bgruening: should we use a more specific tool_id? Maybe something like `sklearn_model_selection` and the name could also be more meaingfiul imho.

bgruening: or sklearn_search_model_selection?

tools/sklearn/main_macros.xml
qiagu: Good point. Thanks.

tools/sklearn/main_macros.xml
qiagu: Not exactly. The first one is somewhat different.

tools/sklearn/main_macros.xml
qiagu: Interesting topic

erasche: You could pull it from the environment. Then administrators could supply a value and change it as they like.

tools/sklearn/main_macros.xml
```diff
@@ -104,11 +101,80 @@ def get_X_y(params, file1, file2):
...
+      lst = ep['estimator_param'].split(":")
+      assert (len(lst) == 2), "Error, make sure there is one and only one colon in search parameter input."
+      literal = lst[1].strip().replace("__sq__","'")\
+          .replace('__dq__','"').replace("__oc__", "{").replace("__cc__", "}")\
+          .replace("__ob__", "[").replace("__cb__", "]")
+      try:
+        ev = ast.literal_eval(literal)
+      except ValueError:
+        valid = ['np.', 'numpy.', 'scipy.']
+        for v in valid:
```
erasche: I read the docs on startswith, apparently you can just write `if literal.startswith(valid)`, if you convert `valid` into a tuple.

tools/sklearn/main_macros.xml
```diff
@@ -104,11 +101,80 @@ def get_X_y(params, file1, file2):
...
+          if literal.startswith(v):
+            ev = eval(literal)
```
erasche: ~~Due to the `split(':')` above, I cannot currently generate any exploit but it does not give me confidence. If that was a `split(':', 1)`, then we would have an issue.~~ I was just being blind, here is an exploit:

```
literal = "numpy.array(__import__('subprocess').check_output('pwd'))"
```

So checking merely the starting characters isn't enough as you can immediately import things and start executing subprocesses. I do not have a lot of confidence in finding a safe way to validate calls to numpy/scipy. I wish I had better news/advice.

erasche: Per discussion in gitter, blacklisting `'` or `"` isn't sufficient as you can construct strings from variables. e.g. `__name__[0] + __name__[3] + ...` and other variables.

qiagu: Thanks for your wonderful comment! My new thought is using regular expression to very the validity of the `literal`. Pattern could be like this: module(numpy/scipy).submodule.method(parameters). Could you please comment on this?

qiagu: The parameters part can take letter, number plus =.,()[]+'". We could block `_import_` as well.

qiagu: For GridSearchCV, I think we could potentially always use `list` datatype. `literal_eval` is enough.  For the randomized search, however, the advantage of random list generator over a fixed list seems  apparent. That's why we want the `eval` to be here. I am thinking we could further limit the generator within the `numpy.random` and `scipy.stats` two submodules. What do others think?

tools/sklearn/main_macros.xml
```diff
@@ -106,49 +106,51 @@ def get_search_params(params_builder, in_pipeline=False):
...
-        for v in valid:
-          if literal.startswith(v):
-            ev = eval(literal)
-            break
-        else:
-          sys.exit("invalid search parameter input for %s" %ep['estimator_param'])
-      search_params[lst[0].strip()] = ev
-    return search_params
+  def in_whitelist(literal):
+    patterns = ['^(np|numpy|scipy)\.(random|stats)\.\w+\(.*\)$', '^(np|numpy)\.arange\([\s\d.,]*\)$' ]
```
qiagu: I guess concerns would come from `\(.*\)$`. The inside of the parenthesis is the parameters of the numpy/scipy method. I could explicitly list the valid characters of parameters, but I didn't since I realize galaxy has already done the job.

Please refer to:
https://github.com/galaxyproject/galaxy/blob/988cab4db7ef83714cf6f0c57f73f4255e3f075d/lib/galaxy/tools/parameters/sanitize.py#L47

tools/sklearn/feature_selection.xml
```diff
@@ -19,19 +19,29 @@ import sys
 import json
 import pandas
 import pickle
+import ast
```
bgruening: is this still needed?

tools/sklearn/main_macros.xml
```diff
@@ -104,11 +99,70 @@ def get_X_y(params, file1, file2):
...
+    module = getattr(sklearn, estimator_module)
+    cls = getattr(module, estimator_cls)
+
+  estimator_params = estimator_json['estimator_params'].strip()
+  if estimator_params == "":
+    estimator = cls()
+  else:
+    estimator_params = estimator_params.replace("__sq__","'")
+    try:
+      params = ast.literal_eval('{' + estimator_params + '}')
```
bgruening: this we can also replace I think?

tools/sklearn/feature_selection.xml
qiagu: This one is used in the place you commented on blow.

tools/sklearn/main_macros.xml
bgruening: Do you have a reference for it?

bgruening: Oh sorry, I mixed it up ... narf. Too late.

qiagu: `literal_eval` should be a safe bet, as safe as python.
https://stackoverflow.com/questions/4710247/python-3-are-there-any-known-security-holes-in-ast-literal-evalnode-or-string

tools/sklearn/main_macros.xml
```diff
@@ -1119,17 +1117,30 @@ def get_estimator(estimator_json):
...
-        <param name="estimator_param" type="text" value="" optional="true" label="Estimator parameter" help="Optional. For example: C: [1, 10, 100, 1000]. See 'help' for more details"/>
+        <expand macro="search_param_input" />
       </repeat>
     </section>
   </xml>
 
+  <xml name="search_param_input" token_param_name="estimator_param" token_label="Estimator parameter" token_help="Optional. For example: C: [1, 10, 100, 1000]. See 'help' for more details">
+    <param name="@PARAM_NAME@" type="text" value="" size="100" optional="true" label="@LABEL@" help="@HELP@">
+      <sanitizer>
+        <valid initial="default">
```
bgruening: Jupp, thats what I meant on gitter :)

qiagu: 😄

qiagu:
This is a really hard PR to me. I'm pretty done now. Look forward to your final review, @bgruening @jgoecks  and maybe @erasche

bgruening:
@qiagu that is a great PR and it even removes more code that it adds! I like it and I'm glad we found a way to hopefully make it secure.

qiagu:
@bgruening Thanks.

jgoecks:
@qiagu @bgruening I have been watching this PR and agree everything great from a high level. I hope to have a chance to dive into the details next week. Feel free to merge in the meantime if others approve.

qiagu:
Can't reach this point without everyone's intelligence investment. Happy with it. 
BTW, the asteval symbol table is great. We can add new entry as needed.

tools/sklearn/main_macros.xml
```diff
@@ -104,11 +99,76 @@ def get_X_y(params, file1, file2):
   return X, y
   </token>
 
+  <token name="@GET_SEARCH_PARAMS_FUNCTION@">
+def get_search_params(params_builder):
+  search_params = {}
+
+  def safe_eval(literal):
+    #allowed symbol table. Add more if needed.
```
bgruening: @erasche we could try/except `import pytest` here and fail it we can import it.

qiagu:
@erasche a whitelist is created explicitly for both scipy.stats and numpy.random. Could this way make the code free of worry? 
Maybe listing of numpy symbols is not necessary. I do it for two purposes: uniform naming and excluding unneeded methods, like `print_function` and `test`.

erasche:
Hey @qiagu that looks great! That feels much more safe.

I tried to think of a nice way to do the replacements so users could use the `.` separated module names that they would be used to. You might be able to do something like this? But maybe the regex is incorrect and would miss things and cause problems / maybe it isn't worth doing.

```python
re.sub(r'([a-z])\.([a-z])', r'\1_\2', input_param)
```

qiagu:
@erasche Thanks for saying positive about the code and giving a nice suggestion.
The regular expression certainly can help users, but I'm not sure whether it covers all corner cases or not. Let me take `np_random_mtrand` as an example, which is actually a module rather than a function. Therefore `np_random_mtrand.randint` is a valid input. The regular expression substitution could disable the usage of the module. Right? A good news is that I don't see any use case of the `mtrand` so far. It's safe to remove it from our symbol table.
I would like to probably stay with current design for now. After seeing more use cases, we'll then decide whether to apply the RE or not.
I could be wrong. Please feel free to correct me if I were.

erasche:
@qiagu Great counter example, just what I was hoping for since I don't know what people would be inputting. In that case, nevermind, no RE needed.

jgoecks:
+1 from me from the user/functionality standpoint, this is a huge improvement and will also us to make progress on https://github.com/bgruening/galaxytools/issues/756

bgruening:
As discussed with @qiagu we will get this in as it contains so much great stuff deploy it on usegalaxy.eu and test it with real-data. The UX can be improved step-by-step but we need to get our hands dirty now.

Thanks a lot @qiagu for this great work. Thanks also to @erasche for the security audit :)

qiagu:
Nice! Thanks for quick installation. @bgruening

PR #764, opened 2018-07-18
Racon: Add PAF format for overlaps

willemdek11:
Added the option for paf overlaps.
This makes it easier to connect racon to minimap output.


bgruening:
@willemdek11 if you plan to use PAF in multiple tools we should maybe add it to Galaxy.
Thanks for you contribution!

willemdek11:
Done!

bgruening:
@willemdek11 do we want to wait for the PAF datatype in Galaxy or merge it now?

willemdek11:
@bgruening I think merging is the way to go, then I'll change it at the moment the PAF format is recognised by Galaxy. Thanks.

PR #771, opened 2018-08-05
NanoPlot tool wrapper added

egeogretmen:


tools/nanoplot/nanoplot.xml
```diff
@@ -0,0 +1,498 @@
...
+            $customization.noN50
+            $customization.N50
+            #if $customization.title
+                --title '$customization.title'
+            #end if
+            $store
+            $raw
+            -o outputs &&
+            mv outputs/*.log outputs/NanoPlot_log.txt &&
+            if [[ -e outputs/NanoPlot-data.pickle ]]; then mv outputs/NanoPlot-data.pickle outputs/NanoPlot-data-pickle.binary; fi
```
bgruening: I don't think we need to support the pickle file case.

tools/nanoplot/nanoplot.xml
```diff
@@ -0,0 +1,498 @@
...
+            $filter.barcoded
+            $customization.noN50
+            $customization.N50
+            #if $customization.title
+                --title '$customization.title'
+            #end if
+            $store
+            $raw
+            -o outputs &&
+            mv outputs/*.log outputs/NanoPlot_log.txt &&
```
bgruening: If we print the log file to stdout we do not need to put them as real output. I do not consider them as useful enough to store them as datasets.

So a simple `cat outputs/*.log` might be enough and will simplify our outputs of this tool dramatically.

tools/nanoplot/nanoplot.xml
```diff
@@ -0,0 +1,498 @@
...
+                <option value="fasta">fasta</option>
+                <option value="fastq_rich">fastq_rich</option>
+                <option value="fastq_minimal">fastq_minimal</option>
+                <option value="summary">summary</option>
+                <option value="bam">bam</option>
+                <option value="cram">cram</option>
+                <option value="pickle">pickle</option>
+            </param>
+            <when value="fastq">
+                <param
```
bgruening: Uha, now I get your gitter question.

I think a simple `<param
                    type="data"
                    name="files"
                    format="fasta,fasta.gz,fasta.bz2,cram,bam"
                    multiple="true" />`

should be enough. No need to support all file formats, just be most standard once and keep your life simple :)

egeogretmen:
@bgruening changed the test files, it should be ok now

bgruening:
merge in IUC: https://github.com/galaxyproject/tools-iuc/pull/2100

PR #772, opened 2018-08-07
add object eval for CV input and make randomness possible in cv splitting

qiagu:
The Penn paper used object as input rather than none/int.

### DIFF
--- a/tools/sklearn/main_macros.xml
+++ b/tools/sklearn/main_macros.xml
@@ -100,52 +100,54 @@ def get_X_y(params, file1, file2):
   return X, y
   </token>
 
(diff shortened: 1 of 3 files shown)


tools/sklearn/main_macros.xml
```diff
@@ -189,6 +191,20 @@ def get_estimator(estimator_json):
   return estimator
   </token>
 
+  <token name="@GET_CV_FUNCTION@">
+def get_cv(literal):
+  if literal == "":
+    return None
+  if re.match(r'^\d+$', literal):
```
bgruening: fyi: there is a python function called literal.isnumeric()

bgruening:
Wrappers are updated.

PR #774, opened 2018-08-13
add balanced_accuracy_score, move python macros to utils.py, change SafeEval and others

qiagu:
After scikit-learn version 0.20.0, balanced_accuracy_score will become built-in


tools/sklearn/main_macros.xml
```diff
@@ -212,8 +212,23 @@ def get_cv(literal):
...
+          warnings.warn('y_pred contains classes not in y_true')
+          per_class = per_class[~np.isnan(per_class)]
+      score = np.mean(per_class)
+      return score
+
   if scorer == "default":
     return None
+  if scorer == "balanced_accuracy":
+    sk_version = tuple([int(x) for x in sklearn.__version__.split('.')])
+    if sk_version &lt; (0, 20):
```
bgruening: CDATA?

qiagu: What's CDATA? 
But here needs else clause

bgruening: https://de.wikipedia.org/wiki/CDATA

qiagu: I guess you talked about the replacement of `<` and `&` signs. `planemo lint` wants those.

qiagu:
I feel hard to make balanced_accuracy_score and multi-metrics work together while keeping a simple UI here.

tools/sklearn/main_macros.xml
bgruening: mh, unlikely, at least if you use CDATA, give me a sec

bgruening: Oh you already fixed it nice!

qiagu:
Not very good. I chose to put balanced_accuracy_score here while giving up the multi-metrics.

tools/sklearn/utils.py
```diff
@@ -0,0 +1,242 @@
...
+import xgboost
+import scipy
+import sklearn
+import ast
+from asteval import Interpreter, make_symbol_table
+from sklearn import metrics, model_selection, ensemble, svm, linear_model, naive_bayes, tree, neighbors
+
+
+def read_columns(f, c=None, c_option='by_index_number', return_df=False, **args):
+  data = pandas.read_csv(f, **args)
```
bgruening: If we start using real python files can you stick to 4 spaces indentation please.

tools/sklearn/utils.py
```diff
@@ -0,0 +1,251 @@
...
+
+    else:
+        score_func = inputs["score_func"]
+        score_func = getattr(sklearn.feature_selection, score_func)
+        new_selector = selector(score_func, **options)
+
+    return new_selector
+ 
+
+def get_X_y(params, file1, file2):
```
bgruening: we should probably call this get_x_y() to follow python guidelines

tools/sklearn/search_model_validation.xml
```diff
@@ -3,7 +3,9 @@
     <macros>
         <import>main_macros.xml</import>
     </macros>
-    <expand macro="python_requirements"/>
+    <expand macro="python_requirements">
+        <requirement type="package" version="0.6">skrebate</requirement>
```
bgruening: We should add https://arxiv.org/abs/1711.08477 to the citations of the tools that use it.

tools/sklearn/utils.py
qiagu: Could you please give more details?

tools/sklearn/search_model_validation.xml
qiagu: Good comment. Thanks for catching it and merging the PR.

tools/sklearn/utils.py
bgruening: The capital letter is just confusing, I guess we should stick to lower-case then for function names.

qiagu: I thought the capital X specifically refers to training features in convention. If you like the little one, I can change it in later PR.

PR #775, opened 2018-08-14
Racon: Changed interval to tabular input format

willemdek11:
Changed the format from interval to tabular, this makes it able to use the newest minimap2 output.


bgruening:
Thanks!

PR #778, opened 2018-08-20
Bump salmon to latest available release

scottx611x:
I wasn't able to test locally w/ `planemo` due to the latest [salmon conda install not working on OSX](https://github.com/bioconda/bioconda-recipes/blob/master/recipes/salmon/meta.yaml#L6). 

@bgruening Although this is passing now, just glancing over the [salmon release notes](https://github.com/COMBINE-lab/salmon/releases) tells me there are a few new options that could be exposed to Galaxy users. I'm not able to interpret the usefulness of these options, and could use some guidance in that respect.

I also saw that [salmon v0.10.0](https://github.com/COMBINE-lab/salmon/releases/tag/v0.10.0) warrants a rebuild of existing salmon indices, which I guess could affect a galaxy user currently pointing to one outside of their history.


bgruening:
We haven't yet provided a salmon index by default so this should not be a problem.

Regarding the new options I think most of them are useful, especially the reverted --useEM option. Rob even stats which options should go into the advanced section, so we should follow that I think :)

bgruening:
Thanks @scottx611x this is very much appreciated!

scottx611x:
@bgruening I think this is getting closer. I've added/removed things based on what I've seen in the release notes. If anything looks blatantly wrong please give me a shout, otherwise I'm going to loop back to this tomorrow.

tools/salmon/salmon.xml
```diff
@@ -362,9 +380,12 @@
             <param argument="--splitSpanningSeeds" type="boolean" truevalue="--splitSpanningSeeds" falsevalue="" checked="False"
                 label="Attempt to split seeds that happen to fall on the boundary between two transcripts."
                 help="This can improve the fragment hit-rate, but is usually not necessary."/>
-            <param argument="--useVBOpt" type="boolean" truevalue="--useVBOpt" falsevalue="" checked="False"
+            <param argument="--useVBOpt" type="boolean" truevalue="--useVBOpt" falsevalue="" checked="True"
```
bgruening: I guess we should remove this one right, as it is the default already.

scottx611x:
Ping @bgruening

scottx611x:
@bgruening I think this is in a better spot now

bgruening:
Awesome. Thanks so much!

PR #779, opened 2018-08-20
make restricted cPickle

qiagu:


tools/sklearn/main_macros.xml
```diff
@@ -8,6 +8,7 @@
           <requirement type="package" version="0.22.0">pandas</requirement>
           <requirement type="package" version="0.72.1">xgboost</requirement>
           <requirement type="package" version="0.9.12">asteval</requirement>
+          <requirement type="package" version="0.6">skrebate</requirement>
```
bgruening: this is not used in every tool, isn't it?

tools/sklearn/utils.py
```diff
@@ -2,18 +2,66 @@
...
+class SafePickler(object):
+    """
+    Used to safely deserialize scikit-learn model objects serialized by cPickle.dump
+    Usage:
+        eg.: SafePickler.load(pickled_file_object)
+    """
+    @classmethod
+    def find_class(self, module, name):
+
+        bad_names = ('and', 'as', 'assert', 'break', 'class', 'continue',
```
bgruening: blacklist / whitelist?

tools/sklearn/utils.py
```diff
@@ -2,18 +2,66 @@
...
+                    '__code__', '__closure__', '__func__', '__self__', '__module__',
+                    '__dict__', '__class__', '__call__', '__get__',
+                    '__getattribute__', '__subclasshook__', '__new__',
+                    '__init__', 'func_globals', 'func_code', 'func_closure',
+                    'im_class', 'im_func', 'im_self', 'gi_code', 'gi_frame',
+                    '__asteval__', 'f_locals', '__mro__')
+        good_names = ('copy_reg._reconstructor', '__builtin__.object')
+
+        if re.match(r'^[a-zA-Z_][a-zA-Z0-9_]*$', name):
+            if  (   (   module.startswith('sklearn.')
```
bgruening: multiple if are maybe more readable ...
And you could start with `name in bad_names` and fail early.

tools/sklearn/svm.xml
```diff
@@ -55,7 +54,7 @@ classifier_object = my_class(**options)
 classifier_object.fit(X, y)
 
 with open("$outfile_fit", 'wb') as out_handler:
-    pickle.dump(classifier_object, out_handler)
+    pickle.dump(classifier_object, out_handler, pickle.HIGHEST_PROTOCOL)
```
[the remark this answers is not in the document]
qiagu: I don't know exactly, but the default( level 0?) failed my tests easily. 
https://stackoverflow.com/questions/23582489/python-pickle-protocol-choice
It sounds people recommend the highest if possible.

tools/sklearn/main_macros.xml
qiagu: Maybe at 2 or 3 places. I put it in the tool initially.  When debugging the pickler, I moved here to rule out the potential problem of this one. I can try to move back later. Be aware that will create a few extra lines.

tools/sklearn/utils.py
qiagu: I had tried to make a explicitly name list. But very hard to make one to cover all, and the list will take a code space. 400+ names from scikit-learn, 600+ from numpy

tools/sklearn/utils.py
qiagu: Good. Should be more efficient.

tools/sklearn/utils.py
bgruening: I was wondering if we should share these list also with the asteval one. The whitelist can grow dynamically I guess. If we see something is missing we add it in later versions.

tools/sklearn/svm.xml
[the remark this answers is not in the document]
bgruening: Yes, you are right.  I assumed the text format does support all the features and is just smaller and less efficient.

PR #784, opened 2018-08-23
segemehl improvements

bernt-matthias:
Actually I wanted just to change the data set label and add output of unmapped reads... ended up with also adding:

- added -f (bisulfite mapping)
- added -R clip accuracy
- added -J jump size
- made use of argument
- documented still missing parameters


tools/segemehl/segemehl.xml
```diff
@@ -137,47 +151,57 @@
           <param name="first_strand_query" type="data" format="fastqsanger,fastqillumina,fastq,fasta" label="Reads from first strand" />
           <param name="second_strand_query" type="data" format="fastqsanger,fastqillumina,fastq,fasta" label="Reads from second strand" />
         </repeat>
-        <param name="maxinsertsize" type="integer" value="5000" label="Maximum size of the inserts (paired end)" help="default: 5000 (-I)" />
+        <param name="maxinsertsize" argument="-I/--maxinsertsize" type="integer" value="5000" label="Maximum size of the inserts (paired end)" help="default: 5000" />
```
bgruening: The argument is meant to be just one param. In this case you could just use argument="--maxinsertsize" and remove the `name="maxinsertsize"` altogether.

bernt-matthias: don

bgruening:
Thanks @bernt-matthias!

jfallmann:
too late, thx matthias

PR #787, opened 2018-08-29
Use sk_whitelist json to replace .py execfile and fix failure of raising fit error

qiagu:
using plotly to produce a parallel coordinates Plot


tools/sklearn/search_plot.xml
```diff
@@ -0,0 +1,87 @@
+<tool id="sk_search_plot" name="Plot Visualization" version="@VERSION@">
+    <description>of SearchCV results</description>
+    <macros>
+        <import>main_macros.xml</import>
```
bgruening: I think this tool should have its own folder. It is independent of sklearn, isn't it? At least it should be.
This means we can also get rid of the macro imho, we do not use it to a large degree isn't it?

tools/sklearn/search_plot.xml
```diff
@@ -0,0 +1,87 @@
...
+        <configfile name="plot_script">
+<![CDATA[
+import sys
+import json
+import plotly
+import plotly.graph_objs as go
+import pandas as pd
+
+input_json_path = sys.argv[1]
+with open(input_json_path, "r") as param_handler:
```
bgruening: I don't think you need this complexity here. See below.

tools/sklearn/search_plot.xml
```diff
@@ -0,0 +1,87 @@
...
+    params = json.load(param_handler)
+
+df = pd.read_csv("$search_results", sep='\t', parse_dates=True)
+
+dimensions = []
+for index in params['col_dimensions']:
```
bgruening: Could you just use `$col_dimensions` here?

tools/sklearn/search_plot.xml
```diff
@@ -0,0 +1,87 @@
...
+    values = df[df.columns[index-1]]
+    dimensions.append(
+        dict(   range = [min(values), max(values)],
+                label = df.columns[index-1],
+                values = values
+        )
+    )
+col_score = params['col_score'][0] - 1
```
bgruening: Same here ....

tools/sklearn/search_plot.xml
```diff
@@ -0,0 +1,87 @@
...
+        </configfile>
+    </configfiles>
+    <inputs>
+        <param name="search_results" type="data" format="tabular" label="Select a result from SearchCV:"/>
+        <param name="col_dimensions" multiple="true" type="data_column" data_ref="search_results" use_header_names="true" display="checkboxes" label="Select the columns for dimentions:"/>
+        <param name="col_score" type="data_column" data_ref="search_results" use_header_names="true" label="Select a column containg the score for coloring:"/>
+    </inputs>
+
+    <outputs>
+        <data name="output" format="html" from_work_dir="output.html" label="Plot of SearchCV result on ${on_string}"/>
```
bgruening: I guess it would be nice to also offer a SVG or PNG as output.

tools/sklearn/search_plot.xml
```diff
@@ -0,0 +1,87 @@
...
+            <output name="output" file="searchcv_plot01.html" compare="sim_size"/>
+        </test>
+
+    </tests>
+    <help><![CDATA[
+**What it does**
+
+    ]]>
+    </help>
+    <expand macro="sklearn_citation"/>
```
bgruening: We do not need this citation here.

tools/sklearn/search_plot.xml
qiagu: Thanks for those great points. Unfortunately, we can't output real images using plotly in Galaxy by now. Please refer to: https://community.plot.ly/t/use-plotly-offline-to-save-chart-as-image-file/408/32
The html offers a lot of interactive functionalities. Users can right click the plot to save to whatever image format they want.

bgruening: Yeah, sorry I should reformulate this. Do you think it would be worth to include a matplotlib version of this to output an SVG directly? Not yet, but in a future version. Maybe it's not worth the effort. The only advantage would be to have SVG as a direct output of a workflow.

qiagu: I'm not sure. Will discuss with Jeremy

PR #788, opened 2018-08-31
H5 models for sklearn modules

anuprulez:
This PR creates h5 ([HDF5](https://support.hdfgroup.org/HDF5/)) models for scikit-learn modules. It uses a package ([DeepdishIO](https://deepdish.readthedocs.io/en/latest/api_io.html)) for transforming the trained models (python objects/classes) to h5. These h5 models can be shared and reused.


tools/sklearn/nn_classifier.xml
```diff
@@ -104,7 +103,14 @@ with open("$outfile_fit", 'wb') as out_handler:
         </expand>
     </inputs>
 
-    <expand macro="output"/>
+    <outputs>
+      <data format="tabular" name="outfile_predict">
```
bgruening: Should this not be in a macro as it was before?

tools/sklearn/utils.py
```diff
@@ -14,6 +14,7 @@
                     gaussian_process, kernel_approximation, linear_model, metrics,
                     model_selection, naive_bayes, neighbors, pipeline, preprocessing,
                     svm, linear_model, tree, discriminant_analysis)
+from deepdish import io
```
qiagu: I think this import and the import in each tool are duplicate.

tools/sklearn/test-data/svc_prediction_result03.tabular
```diff
@@ -33,9 +33,9 @@
 2	-99	26	-108	1
 2	-81	19	-110	0
 2	-108	21	-108	1
-2	-92	27	-106	0
+2	-92	27	-106	1
 2	-88	2	-106	3
```
qiagu: Have you noticed that the prediction results change back and forth?

tools/sklearn/search_model_validation.xml
```diff
@@ -146,20 +145,20 @@ with open(outfile_estimator, "wb") as output_handler:
...
             <param name="selected_search_scheme" value="GridSearchCV"/>
-            <param name="infile_pipeline" value="pipeline01" ftype="zip"/>
+            <param name="infile_pipeline" value="pipeline01" ftype="h5"/>
             <conditional name="search_param_selector">
                 <param name="search_p" value="C: [1, 10, 100, 1000]"/>
                 <param name="selected_param_type" value="final_estimator_p"/>
             </conditional>
             <conditional name="search_param_selector">
-                <param name="search_p" value="k: [-1, 3, 5, 7, 9]"/>
+                <param name="search_p" value="k: [1, 3, 5, 7, 9]"/>
```
qiagu: Don't worry about the `-1`.  It was there to test the `error_score` setting on False, i.e., error tolerant.

tools/sklearn/test-data/feature_selection_result12
```diff
@@ -1,11 +1,11 @@
...
+143.762620712	-0.330941870584
+-88.5787166225	1.08055532812
+-82.8452345578	0.272541389247
+72.4951388149	-0.26868660527800003
+11.805182128	1.0360467096600001
+-63.9354970901	-0.101485840571
+126.32584079600001	-0.35999834017899995
+23.0341392692	0.5185404651359999
+67.6714937696	-0.115688051547
+47.39275848810001	-0.7850965413680001
```
qiagu: It seems the a different column is selected. Do the two columns have the same importance? or other reason?

tools/sklearn/test-data/svc_prediction_result03.tabular
anuprulez: @qiagu I think we are predicting the correct class for the sample shown in ```line 36 ```
https://github.com/anuprulez/galaxytools/blob/sklearn_h5_model/tools/sklearn/test-data/train_set.tabular#L36

https://github.com/anuprulez/galaxytools/blob/sklearn_h5_model/tools/sklearn/test-data/test_set.tabular#L36

tools/sklearn/search_model_validation.xml
anuprulez: But, the test case would fail with this value. The number of features cannot be negative.

qiagu: Have you tried this test case? I don't think it could fail, with `error_score` set to `False`.

tools/sklearn/discriminant.xml
```diff
@@ -32,6 +32,9 @@ with open(input_json_path, "r") as param_handler:
 #if $selected_tasks.selected_task == "load":
 
 classifier_object = io.load("$infile_model")
+module_name = classifier_object.__module__
+class_name = classifier_object.__class__.__name__
+SafePickler.find_class(module_name, class_name)
```
qiagu: I doubt the previous method still works here. 1) the code in the model file is already executed in the `io.load`. 2) it seems that the `SafePickerler` only checks the top level of the object.

PR #789, opened 2018-08-31
diamond: added tax test and test data correction

bernt-matthias:
- prot.accession2taxid was missing the header: https://github.com/bbuchfink/diamond/issues/217
- added test for taxonomic selection by list of ids


bernt-matthias:
Travis has hickups with the good old fatal_oom level ...

PR #791, opened 2018-09-05
make parallel coordinates plot tool

qiagu:
Good for multiple dimensional plotting, for example, visualize the scikit-learn hyperparameter search results.


tools/md_plotting/paracords_plot.xml
```diff
@@ -0,0 +1,85 @@
+<tool id="par_coord_plotting" name="Parallel Coordinates Plot" version="0.1">
+    <description>of tabular data</description>
+    <requirements>
+        <requirement type="package" version="2.7">python</requirement>
```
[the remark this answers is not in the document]
qiagu: Thanks. I have an idea to add certain functionality, but haven't had chance to do it yet.

tools/md_plotting/paracords_plot.xml
```diff
@@ -0,0 +1,120 @@
...
+    </requirements>
+    <version_command>echo $version</version_command>
+    <command detect_errors="aggressive"><![CDATA[
+    python '$plot_script'
+]]>
+    </command>
+    <configfiles>
+        <configfile name="plot_script">
+<![CDATA[
+import sys
```
bgruening: Just saying, no preferences. You could put this python code into its own python file and not mess with cheetah. The reason we chose to do it inline in sklearn was that we had too many parameters to pass around - this is not the case here.

Please ping me if this is ready for review!

qiagu: If you're Okay with this, I would like to keep current style. :)
I'll try the other python style in future tools.

tools/md_plotting/paracords_plot.xml
```diff
@@ -1,108 +1,38 @@
...
-        <requirement type="package" version="3.7">python</requirement>
+        <requirement type="package" version="3.6">python</requirement>
         <requirement type="package" version="0.22.0">pandas</requirement>
         <requirement type="package" version="3.1.1">plotly</requirement>
     </requirements>
     <version_command>echo $version</version_command>
     <command detect_errors="aggressive"><![CDATA[
-    python '$plot_script'
+    python '$__tool_directory__/paracords_plot.py'
+        -i $infile
```
bgruening: please quote all data and text parameters

bgruening:
Aweomse thanks a lot!

PR #796, opened 2018-09-22
Flye

mmiladi:


tools/flye/flye.xml
```diff
@@ -0,0 +1,103 @@
...
+    <inputs>
+        <param name="inputs" type="data" format="fasta,fasta.gz,fastq,fastq.gz" multiple="true" label="Input reads" />
+        <param name="mode" type="select" label="Mode">
+            <option value="--nano-raw">Nanopore raw</option>
+            <option value="--nano-corr">Nanopore corrected</option>
+            <option value="--pacbio-raw">PacBio raw</option>
+            <option value="--pacbio-corr">PacBio corrected</option>
+            <option value="--subassemblies">high-quality contig-like input</option>
+        </param>
+        <param argument="-g" type="text" label="estimated genome size (for example, 5m or 2.6g)" />
```
bgruening: If we find the estimated genome sizes ... we could add something like:

```
    <xml name="effectiveGenomeSize">
        <conditional name="effectiveGenomeSize">
            <param name="effectiveGenomeSize_opt" type="select" label="Effective genome size"
                help="The effective genome size is the portion of the genome that is mappable. Large fractions of the genome are stretches of NNNN that should be discarded.
                    Also, if repetitive regions were not included in the mapping of reads, the effective genome size needs to be adjusted accordingly.
                    We provide a table of useful sizes here: http://deeptools.readthedocs.io/en/latest/content/feature/effectiveGenomeSize.html">
                <option value="93260000">ce10 (93260000)</option>
                <option value="130428560">dm3 (130428560)</option>
                <option value="125464728">dm6 (125464728)</option>
                <option value="2685511504" selected="true">hg19/GRCh37 (2685511504)</option>
                <option value="2701495761">GRCh38/hg38 (2701495761)</option>
                <option value="2304947926">GRCm37/mm9 (2304947926)</option>
                <option value="2308125349">GRCm38/mm10 (2308125349)</option>
                <option value="specific">user specified</option>
            </param>
            <when value="specific">
                <param argument="--effectiveGenomeSize" type="integer" value="" label="Effective genome size"
                       help="e.g. ce10: 93260000, dm3: 130428560, hg19: 2685511504, mm9: 2304947926"/>
            </when>
            <when value="93260000" />
            <when value="130428560" />
            <when value="125464728" />
            <when value="2685511504" />
            <when value="2701495761" />
            <when value="2304947926" />
            <when value="2308125349" />
        </conditional>
</xml>
```

mmiladi: It would be a convenient asset to help the user. Still I would prefer to not guide the user, since flye deals with repeated regions, the sizes likely differ from short read 'effective genome sizes'..

bgruening: flye wants to have the `genome size` isn't it? This pasted one was just an example, we should __not__ use the effective genome size, but we could search for the `genome size` of the most common genomes and offer something similar.

bgruening: As a minimum we should add a validator to the text parameter I think.

tools/flye/test-data/result1_assembly_graph.dot
```diff
@@ -0,0 +1,10 @@
+digraph {
+nodesep = 0.5;
```
mmiladi: @bgruening this is the assembly graph. Is it a familiar/known format? currently the type is set to `data`

bgruening: I don't think so: https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/datatypes/assembly.py but not sure.

mmiladi: https://www.graphviz.org/doc/info/lang.html
How about this? https://github.com/galaxyproject/galaxy/blob/ed6f5d45a9ecbf0cf8a85341e0213b82a9132a46/client/galaxy/scripts/utils/graph.js#L479

bgruening: Yeah, but I haven't seen it here: https://github.com/galaxyproject/galaxy/blob/dev/config/datatypes_conf.xml.sample

Adding a line as `graph_dot` might be a solution. I don't think we can sniff this reliably?

mmiladi: I thought, there might be already some vis. plugins for the type in Galaxy. Google brought up this `graphviz - Visualize graph data using cytoscape.js`: https://docs.galaxyproject.org/en/master/slideshow/architecture/galaxy_architecture.html

bgruening: It takes json and a node-* datatype as seen here: https://github.com/galaxyproject/galaxy/blob/dev/config/plugins/visualizations/graphviz/config/graphviz.xml

bgruening: cool, so we should really add a `graph_dot` datatype that subclasses from txt.

tools/flye/flye.xml
```diff
@@ -0,0 +1,103 @@
...
+        <param argument="-m" type="integer" optional="true" label="minimum overlap between reads (default: auto)" />
+    </inputs>
+    <outputs>
+        <data name="contigs" format="fasta" from_work_dir="out_dir/contigs.fasta" label="${tool.name} on ${on_string} (contigs)">
+        </data>
+        <data name="scaffolds" format="fasta" from_work_dir="out_dir/scaffolds.fasta" label="${tool.name} on ${on_string} (scaffolds)">
+        </data>
+        <data name="assembly_info" format="tabular" from_work_dir="out_dir/assembly_info.txt" label="${tool.name} on ${on_string} (assembly_info)">
+        </data>
+        <data name="assembly_graph" format="data" from_work_dir="out_dir/assembly_graph.dot" label="${tool.name} on ${on_string} (assembly_graph)">
```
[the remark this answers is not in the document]
bgruening: ftype="txt" or ftype="fastqsanger" .... never data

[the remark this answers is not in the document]
mmiladi: in output params, format="data" or format="txt"?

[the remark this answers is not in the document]
bgruening: format is never of `data`. `data` is only a `type`

tools/flye/flye.xml
mmiladi: Could u pls link to a  tool with the validator parser?

tools/flye/flye.xml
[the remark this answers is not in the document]
mmiladi: :+1:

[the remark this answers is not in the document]
mmiladi: Planemo errors: `JobOutputsError: Dataset metadata verification for [file_ext] failed, expected [txt] but found [data].`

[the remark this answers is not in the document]
mmiladi: Sorry my mistake

tools/flye/flye.xml
mmiladi: May we skip this for this PR? :) If input be wrong, the tool complains and fails anyway

bgruening: https://github.com/bgruening/galaxytools/blob/5c6486dead878a8c9521e1d6d50b3a537a2ec2b0/tools/trim_galore/trim_galore.xml#L20
and
https://docs.galaxyproject.org/en/latest/dev/schema.html#tool-inputs-param-validator

tools/flye/test-data/result1_assembly_graph.dot
bgruening: added a PR for it ...

mmiladi: :+1: cool

tools/flye/flye.xml
mmiladi: added

bgruening:
Its deployed!

PR #798, opened 2018-09-24
HDF5 models for machine learning (scikit-learn) tools

anuprulez:
This pull request achieves the following:

-  It creates [hdf5](https://github.com/h5py/h5py) models for machine learning ([scikit-learn](http://scikit-learn.org/stable/)) tools. These models are created after training a classifier/regressor with data. They are first converted to a python dictionary and then to hdf5 files. They are shareable over Galaxy and are persistent.

- Now, these set of tools support python 3.6.

@qiagu thanks a lot for your awesome contribution!
@bgruening thanks a lot for your review!

P.S.: One test should fail (which is expected).


bgruening:
@anuprulez if a test is expected to fail use `expected_failure`: https://docs.galaxyproject.org/en/latest/dev/schema.html#id77

The tests all need to pass before we can merge them.

tools/sklearn/utils.py
```diff
@@ -2,7 +2,7 @@
 import os
 import pandas
 import re
-import cPickle as pickle
+import pickle as pickle
```
qiagu: Just `import pickle`.

tools/sklearn/utils.py
```diff
@@ -55,6 +54,7 @@ def find_class(self, module, name):
                 # TODO: replace with a whitelist checker
                 if fullname not in sk_whitelist['SK_NAMES'] + sk_whitelist['SKR_NAMES'] + sk_whitelist['XGB_NAMES'] + sk_whitelist['NUMPY_NAMES'] + good_names:
                     print("Warning: global %s is not in pickler whitelist yet and will loss support soon. Contact tool author or leave a message at github.com" % fullname)
+                __import__(module, level=0)
```
qiagu: This could damp the whitelist checker. I would avoid this and add needed import in the header.

tools/sklearn/search_model_validation.xml
```diff
@@ -166,23 +163,16 @@ with open(outfile_estimator, "wb") as output_handler:
...
             <param name="selected_column_selector_option" value="all_columns"/>
             <param name="infile2" value="regression_y.tabular" ftype="tabular"/>
             <param name="header2" value="true" />
             <param name="selected_column_selector_option2" value="all_columns"/>
-            <output name="outfile_result" >
-                <assert_contents>
-                    <has_text_matching expression="[^/d]+0.7938837807353147[^/d]+{u'estimator__C': 1, u'preprocessing_2__k': 9}[^/d]+1" />
-                    <has_text text="0.0"/>
-                </assert_contents>
-            </output>
```
qiagu: What happen to this test? The `error_score` is a very useful feature that was tested here. If regular express can't pass, we may try just plain text.

tools/sklearn/dict_h5_pickler.py
```diff
@@ -0,0 +1,510 @@
...
+    """
+
+    def _create_dataset(self, file_obj, key, value):
+        """
+        Create dataset
+        """
+        try:
+            file_obj.create_dataset(key, data=json.dumps(value))
+        except:
+            file_obj.create_dataset(key, data=value)
```
qiagu: Could you please give some details, about when to use `data=json.dumps(value)`  and when to use `data=value` ?  Thanks.

tools/sklearn/dict_h5_pickler.py
```diff
@@ -0,0 +1,510 @@
...
+        """
+        serialised_model = ModelToDict().dump(model)
+        h5file = h5py.File(file_name, 'w')
+
+        # nested method for recursion
+        def recursive_save_model(h5file_obj, dictionary):
+            for model_key, model_value in dictionary.items():
+                type_name = type(model_value).__name__
+                try:
+                    if type_name in ['list']:
```
qiagu: How about this? 
```
type_name = type(model_value)
try:
      if type_name is list:
```

tools/sklearn/dict_h5_pickler.py
```diff
@@ -0,0 +1,510 @@
...
+                        if len(model_value) > 0:
+                            list_obj = all(isinstance(x, dict) for x in model_value)
+                            if list_obj is False:
+                                self._create_dataset(h5file_obj, model_key, model_value)
```
qiagu: It seems h5 can store nested list directly. Great! How about storing nested tuple? I guess it can handle that as well.

tools/sklearn/dict_h5_pickler.py
```diff
@@ -0,0 +1,510 @@
...
+                                for index, model_item in enumerate(model_value):
+                                    model_key_item = model_key + "/" + str(index)
+                                    if model_item is not None:
+                                        if model_key_item in h5file_obj:
+                                            recursive_save_model(model_key_item, model_item)
+                                        else:
+                                            group = h5file_obj.create_group(model_key_item)
+                                            recursive_save_model(group, model_item)
+                                    else:
+                                        self._create_dataset(h5file_obj, model_key_item, model_item)
```
qiagu: For this `else`, `model_item` is None? Sorry, I don't quite understand here.

tools/sklearn/dict_h5_pickler.py
```diff
@@ -0,0 +1,510 @@
...
+                        self._create_dataset(h5file_obj, model_key, model_value)
+                except Exception:
+                    continue
+        recursive_save_model(h5file, serialised_model)
+
+    def _restore_list_in_model(self, model_object):
+        """
+        Convert the model dictionary to list if the keys are represented as numbers
+        """
+        if isinstance(model_object, dict) is True:
```
qiagu: `if isinstance(model_object, dict):`

tools/sklearn/dict_h5_pickler.py
```diff
@@ -0,0 +1,510 @@
...
+                    all_keys_number = all(str.isnumeric(x) for x in keys)
+                    if all_keys_number is True:
+                        keys = [int(ky) for ky in keys]
+                        model_object[key] = list()
+                        for idx in range(len(keys)):
+                            model_object[key].append(value[str(idx)])
+                self._restore_list_in_model(value)
+        elif isinstance(model_object, list) is True:
+            for k, v in enumerate(model_object):
+                self._restore_list_in_model(v)
```
qiagu: It seems the `k` is never used.

tools/sklearn/dict_h5_pickler.py
```diff
@@ -0,0 +1,510 @@
...
+        if isinstance(model_object, dict) is True:
+            for key, value in model_object.items():
+                if type(value).__name__ in ['dict']:
+                    keys = value.keys()
+                    all_keys_number = all(str.isnumeric(x) for x in keys)
+                    if all_keys_number is True:
+                        keys = [int(ky) for ky in keys]
+                        model_object[key] = list()
+                        for idx in range(len(keys)):
+                            model_object[key].append(value[str(idx)])
```
qiagu: How about some keys are numeric, some are not?

tools/sklearn/search_model_validation.xml
anuprulez: In this test, `<test expect_failure="true">` and `<param name="error_score" value="false"/>` do not go together. This test passes when either `<test expect_failure="true">` is used or `<test expect_failure="true">` and `<param name="error_score" value="true"/>` are used. When we use `<test expect_failure="true">`, it is not important to compare any output. That's why, the output of this test was removed.

https://docs.galaxyproject.org/en/release_18.05/dev/schema.html#id76

tools/sklearn/dict_h5_pickler.py
anuprulez: This is particularly useful when an item is `None`. By dumping to json, it becomes `"null"` and while loading `"null"` to `None`

tools/sklearn/dict_h5_pickler.py
anuprulez: For nested lists, I created dictionary recursively with indices as the keys and while unpacking, I remove those keys to recreate the list. It also explains the why some keys are numeric, others not

tools/sklearn/dict_h5_pickler.py
anuprulez: It takes care of complex lists containing objects of different types like `None`, `int`, `str` or `dict`. When `model_item` is `None`, just create a hierarchy with `None` as its value.

tools/sklearn/dict_h5_pickler.py
anuprulez: Lists containing dicts with the same set of keys. All these dicts are stored a big dict with their indices as their keys.

tools/sklearn/dict_h5_pickler.py
anuprulez: Removed. Thanks!

PR #799, opened 2018-09-26
Flye postpone graph_dot format

mmiladi:


bgruening:
No way ;)
release will be out on our server next week or so ... and if this is too slow for you I will hand-patch the datatype for you ;)

PR #800, opened 2018-09-27
graphmap 

mmiladi:


mmiladi:
There is also graphmap overlap variants. It is very similar. Once you reviewed this one, I will clone and add the new wrapper.

tools/graphmap/graphmap_align.xml
```diff
@@ -0,0 +1,65 @@
...
+        -o output.sam 
+        #if $preset:
+            -x $preset
+        #end if
+        -t \${GALAXY_SLOTS:-4} 
+    ]]></command>
+    <inputs>
+        <param argument="-r" type="data"  format="fasta,fastq" label="reference sequence" />
+        <param argument="-d" type="data"  format="fasta,fastq" label="reads" />
+        <param argument="-u" type="boolean" truevalue="-u" falsevalue="" label="SAM in the order of input reads" />
```
bgruening: The label could be more clear. Is this an input sorted SAM file? Wouldn't it be better to output the BAM natively?

tools/graphmap/graphmap_align.xml
```diff
@@ -0,0 +1,65 @@
...
+
+        <param name="preset" type="select" optional="True" label="Pre-set parameters (-x)">
```
bgruening: A short description what they are doing ... maybe? Not required for a first version I guess.

tools/graphmap/macros.xml
```diff
@@ -0,0 +1,23 @@
+<macros>
+    <xml name="requirements">
+        <requirements>
+        <requirement type="package" version="0.5.2">graphmap</requirement>
```
bgruening: You could make the version number a macro and use it as well in the tool-version.

tools/graphmap/graphmap_align.xml
```diff
@@ -0,0 +1,65 @@
+<tool id="graphmap_align" name="Mapper" version="0.5.2">
+    <description>for long, error-prone reads, like Nanopore ONT and PacBio.</description>
```
bgruening: remove the dot at the end

tools/graphmap/graphmap_align.xml
mmiladi: It does only report sam. With this option the sam will be sorted. I will update the desc. :+1:

tools/graphmap/graphmap_align.xml
mmiladi: added

tools/graphmap/macros.xml
mmiladi: :+1:

tools/graphmap/graphmap_align.xml
mmiladi: :+1:

tools/graphmap/graphmap_align.xml
bgruening: we could easily depend on samtools and do the conversion on our own, if needed to both coordinate-sorted-bam and name-sorted bam

bgruening: Just to be clear, I'm taking about adding a `&& samtools view -b` thing at the end of the command.

mmiladi:
This should be ready hopefully.

tools/graphmap/test-data/result1_owler.mhap
```diff
@@ -0,0 +1,7 @@
+1 157 0.742839 3672 1 9 16450 22450 0 37 16202 16211
```
bgruening: Is this really tab separated? It looks space separated ...

mmiladi: true. which format do u suggest to use?

mmiladi: This looks rather like a matrix format of connectivity.. Should I use txt?

tools/graphmap/graphmap_align.xml
mmiladi: Aha this sound useful and neeed anyway. ok

mmiladi: So like this: `&&        samtools view -b > output.bam`
Also sorting?

bgruening: https://github.com/galaxyproject/tools-iuc/blob/57493fdc88e6128bed127993464f2a6480e2774d/tools/bwameth/bwameth.xml#L81

tools/graphmap/test-data/result1_owler.mhap
bgruening: yes, then go for text, or is this is proper fileform which we should support?

mmiladi: I don't think it would become a popular format.  thank.

tools/graphmap/graphmap_align.xml
```diff
@@ -0,0 +1,66 @@
...
+        #if $preset:
+            -x $preset
+        #end if
+        -t \${GALAXY_SLOTS:-4} &&
+        samtools view -b output.sam | samtools sort -@ "\${GALAXY_SLOTS:-4}" -T tmp -O bam -o output.bam -
+    ]]></command>
+    <inputs>
+        <param argument="-r" type="data"  format="fasta,fastq" label="reference sequence" />
+        <param argument="-d" type="data"  format="fasta,fastq" label="reads" />
+        <param argument="-u" type="boolean" truevalue="-u" falsevalue="" label="Output sorted as the order of input reads" />
```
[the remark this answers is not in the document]
mmiladi: It's about the output sorting, not the input. I think we can also skip supporting this option for the first version.

bgruening:
Why not using `samtools view -u ` it should be faster as its uncompressed BAM that is piped to samtools sort ... see the link from above.

Otherwise, this looks great. Thanks a lot!

mmiladi:
Fine point! Thank you! Almost all puzzle parts are getting there ;)

PR #809, opened 2018-10-10
Visualizations for confusion matrix, precision, recall, fscore and AUC

anuprulez:
This pull request adds a tool which produces 3 separate plots, one each for confusion matrix and precision, recall, f-score and AUC (area under the ROC curve). 

cnf

prf

roc


anuprulez:
I noted a couple of points for the `perf` tool:
- It is for boolean classification problem.
- I am not sure how does it take the model file to compute probabilistic scores for computing AUC.

qiagu:
Cool. I'll add the .shed.yml later

qiagu:
Oh, It's already there for the paracoords plot. Thanks. @bgruening

PR #810, opened 2018-10-15
enhance searchcv tool and upgrade of sklearn to v0.20.0

qiagu:
1) make explicit cv splitter
2) support more estimator searches, like the `base_estimator` of `AdaBoostRegressor`
3) support preprocessor swap/search
4) make preprocessor hot number/keyword


tools/sklearn/main_macros.xml
```diff
@@ -244,8 +244,8 @@
     <param argument="shuffle" type="boolean" optional="true" truevalue="booltrue" falsevalue="boolfalse" checked="@CHECKED@" label="@LABEL@" help="@HELP_TEXT@"/>
   </xml>
 
-  <xml name="random_state" token_default_value="" token_help_text="Integer number. The seed of the pseudo random number generator to use when shuffling the data. A fixed seed allows reproducible results.">
-    <param argument="random_state" type="integer" optional="true" value="@DEFAULT_VALUE@" label="Random seed number" help="@HELP_TEXT@"/>
+  <xml name="random_state" token_default_value="" token_help_text="Integer number. The seed of the pseudo random number generator to use when shuffling the data. A fixed seed allows reproducible results. default=None.">
+    <param argument="random_state" type="integer" optional="true" value="@DEFAULT_VALUE@" label="random_state" help="@HELP_TEXT@"/>
```
bgruening: Why changing the label here ... a label with underscore looks wired.

tools/sklearn/main_macros.xml
```diff
@@ -825,10 +825,114 @@
...
+      </when>
+      <when value="LeavePGroupsOut">
+        <param argument="n_groups" type="integer" value="" label="n_groups" help="Number of groups (p) to leave out in the test split." />
+        <expand macro="cv_groups"/>
+      </when>
+    </conditional>
+  </xml>
+
+  <xml name="cv_n_splits" token_value="3" token_help="Number of folds. Muast be at least 2.">
+    <param argument="n_splits" type="integer" value="@VALUE@" label="n_splits" help="@HELP@"/>
```
bgruening: When it needs to be at least 2, then remove this text and use `min="2"`

tools/sklearn/main_macros.xml
```diff
@@ -825,10 +825,114 @@
...
+  </xml>
+
+  <xml name="cv_shuffle">
+    <param argument="shuffle" type="boolean" optional="true" truevalue="booltrue" falsevalue="boolfalse" checked="false" label="shuffle" />
```
bgruening: The label can be enhanced a little bit. What is to shuffle etc ..

tools/sklearn/main_macros.xml
```diff
@@ -825,10 +825,114 @@
...
+  </xml>
+
+  <xml name="cv_test_size" token_value="0.2">
+    <param argument="test_size" type="float" value="@VALUE@" label="test_size" help="0.0-1.0, proportion of the dataset to include in the test split; >1, integer only, the absolute number of test samples "/>
```
bgruening: remove `0.0-1.0` and use `min="0.0" max="1.0"`

tools/sklearn/main_macros.xml
```diff
@@ -825,10 +825,114 @@
...
+  </xml>
+
+  <xml name="cv_groups" >
+    <param argument="groups" type="text" value="" area="true" label="groups" help="Group lables in a list. e.g., [1, 1, 2, 2, 3, 3, 3]">
```
bgruening: indentation

bgruening: Is the bracket really needed, we can add it later isn't it?

tools/sklearn/model_validation.xml
```diff
@@ -67,7 +68,16 @@ y = read_columns(
 y=y.ravel()
 
 options = params["model_validation_functions"]["options"]
-options['cv'] = get_cv( options['cv'] )
+splitter, groups = get_cv(options.pop('cv_selector'))
+if groups is not None:
+    groups = groups.strip()
+    if groups == "":
+        groups = None
```
bgruening: How does an input look like if this is true and why do we pass then None to the splitter ... is adding `  ` a actual usecase?

tools/sklearn/main_macros.xml
qiagu: wow, I think you're right. Explicit labels sound better.

tools/sklearn/main_macros.xml
qiagu: Good idea. I think we still can keep the `min="2"` in the help.

tools/sklearn/main_macros.xml
qiagu: We can add `min="0.0"`. `test_size` takes both float, 0.0-1.0, and integer, greater than 1.0. When integer, represent the number of instances in tests.

tools/sklearn/main_macros.xml
qiagu: indentation?

I have an idea. Using string.split to generate the list. I prefer to keep the `[` and `]`, but remove them ( will be __fb__ and __rb__) before calling split. What do you think?

tools/sklearn/model_validation.xml
qiagu: It's consistent to cast empty string to `None` and may be helpful in debugging.

qiagu:
I'll make changes later today. Thanks

tools/sklearn/main_macros.xml
qiagu: Recently, I began to like short labels, especially for those key words existing in scikit-learn, maybe just because I am a heavy user. I also thought simple words made the UI clean. But I'll value your opinion and other people's as well.

tools/sklearn/model_validation.xml
bgruening: Ok, if you think this is more readable. Should we use this as a macro?

qiagu: I'll look into this later, thanks.

tools/sklearn/search_model_validation.xml
```diff
@@ -551,7 +667,118 @@ The evaluation of settings supports operations in Math, list comprehension, nump
 
 - max_depth: scipy_stats_randin(1, 11)
 
-- estimator: [ensemble_ExtraTreesClassifier(n_estimators=100, random_state=324089)]
+**Estimator search/eval (additional '-'):**
+
+- base_estimator-: [sklearn_tree.DecisionTreeRegressor(), sklearn_tree.ExtraTreeRegressor()]
```
bgruening: I find this `-:` syntax very confusing. Can you explain this in more detail, please.

tools/sklearn/search_model_validation.xml
```diff
@@ -551,7 +667,118 @@ The evaluation of settings supports operations in Math, list comprehension, nump
...
+
+'decomp_all': All decomposition estimators, i.e., 15-25
+
+'k_appr_all': All kernel_approximation estimators, i.e. 26-29
+
+'reb_all': All skrebate estimators, i.e. 31-35
+
+'imb_all': All imbalanced-learn sampling methods, i.e. 36-54. CAUTION: Mix of imblearn and other preprocessors may not work.
+
+'all': All, 0-36
```
bgruening: None, is not explained.

tools/sklearn/search_model_validation.xml
```diff
@@ -551,7 +667,118 @@ The evaluation of settings supports operations in Math, list comprehension, nump
...
+
+Support mix (except imblearn methods), e.g.:
+
+- : [None, 'sk_prep_all', 22, 'k_appr_all', sklearn_feature_selection.SelectKBest(k=50)]
```
bgruening: the `- :` is part of the parameter or not? Why do we not offer these 35 options in a select field? Or in a repeat?

tools/sklearn/utils.py
```diff
@@ -222,11 +254,112 @@ def get_search_params(params_builder):
...
+                            imblearn.over_sampling.SMOTE(random_state=0, n_jobs=N_JOBS),
+                            imblearn.over_sampling.SVMSMOTE(random_state=0, n_jobs=N_JOBS),
+                            imblearn.over_sampling.BorderlineSMOTE(random_state=0, n_jobs=N_JOBS),
+                            imblearn.over_sampling.SMOTENC(categorical_features=[], random_state=0, n_jobs=N_JOBS),
+                            imblearn.combine.SMOTEENN(random_state=0), imblearn.combine.SMOTETomek(random_state=0)]
+            i = 0
+            while i < len(ev):
+                obj = ev[i]
+                if obj is None:
+                    i = i + 1
```
bgruening: i += 1

tools/sklearn/utils.py
```diff
@@ -222,11 +254,112 @@ def get_search_params(params_builder):
...
+                    if 'n_jobs' in obj.get_params():
+                        obj.set_params( n_jobs=N_JOBS )
+                if param_type == "final_estimator_p":
+                    search_params["estimator__" + param_name[:-1]] = ev
+                else:
+                    search_params["preprocessing_" + param_type[5:6] + "__" + param_name[:-1]] = ev
+        elif param_type != "final_estimator_p":
+            #TODO regular express check ?
+            ev = safe_eval_es(literal)
+            preprocessors = [preprocessing.StandardScaler(), preprocessing.Binarizer(), preprocessing.Imputer(),
```
bgruening: I find it more readable if you create a separate list for all the sub-groups to not do the slicing later preprocessors[0:8] and so on..

tools/sklearn/utils.py
```diff
@@ -222,11 +254,112 @@ def get_search_params(params_builder):
...
+                            imblearn.over_sampling.SMOTENC(categorical_features=[], random_state=0, n_jobs=N_JOBS),
+                            imblearn.combine.SMOTEENN(random_state=0), imblearn.combine.SMOTETomek(random_state=0)]
+            i = 0
+            while i < len(ev):
+                obj = ev[i]
+                if obj is None:
+                    i = i + 1
+                    continue
+                elif obj == 'all':
+                    ev[i:i+1] = preprocessors
```
bgruening: I guess I would collect the preprocessors in a separate list and do the merging and constructing of ev in the end. This is not really readable imho.

tools/sklearn/search_model_validation.xml
qiagu: No better reason, just differ from other non-estimator eval. Do you have better idea?

tools/sklearn/search_model_validation.xml
qiagu: The `-` is not part of the parameter, but cheetah markdown. The point in your second question is very good. If we do that, it will change the UI dramatically though. Let me think about it.

tools/sklearn/utils.py
qiagu: Oh, I was confused with `++` which is not supported in Python

tools/sklearn/utils.py
qiagu: Probably you're right. My solution may not be the best, or even one of those. I have no better idea on how to handle the hot numbers, and the mix at current moment. If you have one, a PR or commit is welcome

tools/sklearn/utils.py
qiagu: Oh, using a separate list by append/extent is a good idea.

qiagu: Doing this `inplace` has a little memory benefit. :) I agree that making a newlist has more advantages.

anuprulez:
That's a great work especially adding the `SMOTE` thing. I learnt about it in a different project sometime back to create additional data/oversampling using the similar distribution as the data itself. Thanks a lot @qiagu!

Do we need to recompute the results with this new version? It seems like a lot of change :)

qiagu:
Thanks. I guess most of the dada don't need re-test. There could be float precision variance. I don't think it will change our figures much. For safe, we can run re-test of couple small datasets. Besides, we still can write the paper saying we tested in v19, IMHO

qiagu:
I feel the `n_jobs =-1` helped the searchcv on EU sever before the parameter was overridden. I didn't have experience in my laptop, sorry

tools/sklearn/feature_selection.xml
```diff
@@ -103,11 +103,14 @@ res.to_csv(path_or_buf="$outfile", sep='\t', index=False)
         <expand macro="feature_selection_all">
             <expand macro="fs_selectfrommodel_prefitted"/>
         </expand>
-        <expand macro="feature_selection_output_mothods" />
+        <param name="save" type="boolean" truevalue="booltrue" falsevalue="boolflase" checked="false" label="Save the fitted selector?"/>
```
anuprulez: Again a typo: `falsevalue="boolflase"`. It should be `falsevalue="boolfalse"` :)

tools/sklearn/main_macros.xml
```diff
@@ -824,14 +833,125 @@
...
+  <xml name="cv_reduced">
+    <conditional name="cv_selector">
+      <param name="selected_cv" type="select" label="Select the cv splitter:">
+        <expand macro="cv_splitter"/>
+      </param>
+      <expand macro="cv_splitter_options"/>
+    </conditional>
+  </xml>
+
+  <xml name="cv_n_splits" token_value="3" token_help="Number of folds. Muast be at least 2.">
```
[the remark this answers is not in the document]
qiagu: Thanks @anuprulez

tools/sklearn/model_validation.xml
```diff
@@ -67,7 +69,13 @@ y = read_columns(
 y=y.ravel()
 
 options = params["model_validation_functions"]["options"]
-options['cv'] = get_cv( options['cv'] )
+splitter, groups = get_cv(options.pop('cv_selector'))
+if groups is None:
+    options['cv'] = splitter
+elif groups == "":
+    options['cv'] = list( splitter.split(X, y, groups=None) )
```
anuprulez: I see two empty spaces before and after `list`. But, in line `72` the spaces are not present. In general, we should stick to a similar standard.

qiagu:
I plan to remove the cache. It only helps small number of algorithms and there could be alternatives. What do you think? @bgruening @anuprulez

tools/sklearn/discriminant.xml
```diff
@@ -106,7 +106,7 @@ with open("$outfile_fit", 'wb') as out_handler:
             <param name="selected_algorithm" value="LinearDiscriminantAnalysis"/>
             <param name="solver" value="svd" />
             <param name="store_covariance" value="True"/>
-            <output name="outfile_fit" file="lda_model01" compare="sim_size" delta="500"/>
+            <output name="outfile_fit" file="lda_model01" compare="sim_size" delta="1"/>
```
anuprulez: The value of `delta` varies throughout. Some places it is `1`, other places `5`. Is it possible to use just one value?

anuprulez:
@qiagu @bgruening 

It is a great work. Thanks!

I think more and more people use these tools in workflows and the workflows already save the parameter values (reusable). Moreover, for standalone tool usage as well a user can rerun the job with the same parameters as the original job. The caching of these parameter values would be superfluous in my opinion if I understand the purpose of caching correctly.

qiagu:
The caching could increase the speed of RFE when tuning the `n_features_to_select`, but we can use RFECV which is more efficient.

tools/sklearn/discriminant.xml
qiagu: The file size varies a lot. For MB sized file, dalta 500 is good enough, but for a file of hundreds of Byte, delta 1 gives better accuracy.

bgruening:
@qiagu I'm with you here. It was good to explore the caching but this can wait for later and we can discuss this in more detail and maybe can come up with some infrastructure in Galaxy as well. Ideally, this should be not exposed to the user but handled in a smart way by the system.

tools/sklearn/discriminant.xml
anuprulez: I am not sure I understood it completely. Here `sim_size` is being used under tests. How it is related to the performance of an algorithm?

qiagu: Sorry. The accuracy doesn't relate to performance, but points to the right output. The delta, in my understanding, is the difference between the output size and the test data file size. If the size of output is expected to be in hundreds of Bytes, delta 500 may provides little assurance that the output from the test is the same as the test data file.

tools/sklearn/pipeline.xml
```diff
@@ -80,26 +92,52 @@ def get_component(input_json, check_none=False):
...
+            has_imblearn = True
 
 # Set up final estimator and add to pipeline.
 estimator_json = params["final_estimator"]['estimator_selector']
-estimator = get_estimator(estimator_json)
-
-pipeline_steps.append( ('estimator', estimator) )
+if estimator_json['selected_module'] == 'none':
+    if len(pipeline_steps) == 0:
+        sys.exit("No pipeline steps!")
```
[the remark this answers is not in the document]
qiagu: Thanks @bgruening

bgruening:
@qiagu I will try to release a new version of xgboost, but this can take some time. We can only increase the version number of dependencies if there is a conda package available. For xgboost 0.81 there is none at the moment. There are also a bunch of API changes as documented here: https://github.com/dmlc/xgboost/releases

bgruening:
Ok, I was able to create this conda-package ... will restart in a minute the tests.

qiagu:
Looks there are a lot of hassles. Thank you @bgruening

bgruening:
We could also go back to the older xgboost version, I think.

bgruening:
Thanks for reverting. Let's get this in.

@qiagu this is outstanding work, thanks a lot!

bgruening:
@qiagu should be updated :)

PR #814, opened 2018-10-27
rnacode: added conditional for eps creation

bernt-matthias:
For larger fasta files the number of eps files generated overloads galaxy, so I changed the default to "not generate pictures". For users that still want the pictures I added the option to generate a single multipage eps / tar.gz. I'm happy to remove this again if you think its not necessary, since it requires ghostscript as additional dependency. I guess for larger fasta files nobody will look at the pictures anyway.

In our example for a human chromosome nearly 40000 eps file = 6GB are generated. Now our Galaxy is busy integrating them:

```
Created dynamic collection dataset for path [/work/songalax/galaxy-dev/database/jobs_directory/004/4073/working/eps/hss-16840.eps] with element identifier [hss-16840] for output [hss-16840] (734.055 ms)
```

Btw. any idea how to stop this .. our Galaxy is nearly unusable meanwhile :(


bernt-matthias:
@bgruening you were right that `eps.tar.gz` is missing. I didn't check that `.gz` is in `.gitignore`. 

oddly in this new PR linting seems to work. maybe a temprary issue with ncbi ftp.

tools/rna_tools/rnacode/rnacode.xml
```diff
@@ -1,7 +1,8 @@
-<tool id="rbc_rnacode" name="RNAcode" version="0.3.1">
+<tool id="rbc_rnacode" name="RNAcode" version="0.3+galaxy2">
```
bgruening: @bernt-matthias last coment. I think we need to keep it with `0.3.2` as `0.3.1` is greater then `0.3+galaxy2` afaik.

bernt-matthias: indeed. changed it

bgruening:
Thanks!

PR #817, opened 2018-10-31
Visualizations for regression

anuprulez:
This pull request adds three plots for regression tasks: true vs predicted curves, scatter plots for true and predicted targets and a residual plot. It takes two tabular input files: true and predicted targets.

True vs predicted curves:

true_pred_curves

Scatter plot for true and predicted targets:

true_vs_pred_scatter

Residual plot:

residual_plot


anuprulez:
ping @bgruening @qiagu

qiagu:
👍 Great

tools/plotly_regression_performance_plots/plotly_regression_performance_plots.xml
```diff
@@ -0,0 +1,43 @@
...
+    <version_command>echo $version</version_command>
+    <command detect_errors="aggressive"><![CDATA[
+    python '$__tool_directory__/plot_regression_performance.py'
+        -i '$infile_input'
+        -j '$infile_output'
+]]>
+    </command>
+    <inputs>
+        <param name="infile_input" type="data" format="tabular" label="Select input data file :" help="Input data is a matrix (tabular) where each column is a feature and the last column contains the target values."/>
+        <param name="infile_output" type="data" format="tabular" label="Select predicted data file :" help="Predicted data is a matrix (tabular) where each column is a feature and the last column contains the predicted values."/>
```
qiagu: I think once the last column contains the designated values, the data can be retrieved for visualization. In other words, files contain one column which are the true labels are still valid input here.

anuprulez: Yes, just true and predicted value columns are enough for inputs. But, we just get complete data either with true values or predicted values.

qiagu: How about this?
```A tabular in which the last column (including the column if there is only one) contains the target values ```

anuprulez: Update the help tex. Thanks :)

bgruening:
Thanks both! We do have a tool installation stop until we finish migration of usegalaxy.eu to the new head node. Sorry for that.

PR #821, opened 2018-11-26
Canu: append M to Galaxy Memory MB and bump to 1.8

mvdbeek:
Without M the value is taken as gigabyte, and `stopOnReadQuality` has been removed in 1.8.

### DIFF
--- a/tools/canu/canu.xml
+++ b/tools/canu/canu.xml
@@ -46,10 +46,10 @@
         obtovlThreads=\${GALAXY_SLOTS:-4}
         utgovlThreads=\${GALAXY_SLOTS:-4}
         batThreads=\${GALAXY_SLOTS:-4}
-        batMemory=\${GALAXY_MEMORY_MB:-7}M
-        cormhapMemory=\${GALAXY_MEMORY_MB:-7}M
-        obtovlMemory=\${GALAXY_MEMORY_MB:-7}M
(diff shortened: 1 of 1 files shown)


mvdbeek:
Hmm https://travis-ci.org/bgruening/galaxytools/jobs/459813443#L2697 looks like we need GALAXY_SLOTS = 2 to get the test running

mmiladi:
Hi @mvdbeek , thanks for the PR.  Canu's internal job scheduler has a quite complicated behavior, very hard to find an optimal resource allocation :(

bgruening:
@mvdbeek I fine with adding 2 SLOTS if this solves the error.

mvdbeek:
I'm afraid GALAXY_SLOTS isn't under our control, the local job runner just always sets it to 1 (https://github.com/galaxyproject/galaxy/blob/3c97d30cd6b4f11270fc1e057f236f420542cfea/lib/galaxy/jobs/runners/util/job_script/CLUSTER_SLOTS_STATEMENT.sh#L27) if not set explicitly in the job_conf.xml file. Maybe it'd make sense for planemo to set this to `ncpu` with a certain limit ?

bgruening:
This is sitting here far too long and it has a reason that it's not in IUC ... so I will merge as it is.

Thanks @mvdbeek!

PR #822, opened 2018-11-27
Update flye to 2.3.7

mvdbeek:


mvdbeek:
I've also taken different input data (nanopore reads from a 30KB region) because 2.37 would complain about the coverage being too low. Unfortunately it looks like results are fluctuating a bit between local test and tests on travis, so I just use compare="sim_size".

bgruening:
Thanks @mvdbeek!

PR #823, opened 2018-12-02
Get GFA output and drop contigs output

mvdbeek:
It looks like in 2.3.7 there is only a `scaffolds.fa` file, and there is no `assembly_graph.dot` file.
Also adds the gfa output.


bgruening:
Wondering how the test could pass. Maybe sim_size is fault?
Thanks Marius.

mmiladi:
Hi @mvdbeek , thanka! Do you mean there is no contigs or no scaffold file anymore? Though a bit strange update :)

mvdbeek:
You can still copy the consensus from the `1-consensus` folder, but I'm not sure there is much of a point having the contigs before the repeat resolving and polishing step. I think you can still get the individual contigs from the gfa file if needed.

PR #826, opened 2018-12-17
Bugfix for Head: Write Error: Broken pipe

Florian-H-Lab:


Florian-H-Lab:
I have no fricking clue. The planemo test run works but here on Travis it doesn't.

PR #829, opened 2019-01-03
Fix up the patch from Flo

bgruening:
replaces https://github.com/bgruening/galaxytools/pull/826


Florian-H-Lab:
Should I fix the conda package for peakachu, including r-base=3.4.1?

PR #830, opened 2019-01-08
Add RMSE and R2 scores to scatter plot to plot title

anuprulez:
rmse_r2

### DIFF
--- a/tools/plotly_regression_performance_plots/plot_regression_performance.py
+++ b/tools/plotly_regression_performance_plots/plot_regression_performance.py
@@ -1,5 +1,6 @@
 import argparse
 import pandas as pd
+import numpy as np
 import plotly
 import plotly.graph_objs as go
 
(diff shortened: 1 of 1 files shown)


tools/plotly_regression_performance_plots/plot_regression_performance.py
```diff
@@ -62,7 +69,7 @@ def main(infile_input, infile_output):
     )
 
     layout_true_pred = go.Layout(
-        title='True vs predicted values',
+        title='True vs predicted values (RMSE: ' + str(np.round(rmse, 2)) + ', R2: ' + str(np.round(r2, 2)) + ')',
```
bgruening: You should use string substitutions like `"%s (RMSE: 766) %s (foo bar)" % (var1, var2)`

anuprulez: fixed!

PR #833, opened 2019-01-11
find_and_replace: add missing global flag

bernt-matthias:


tools/text_processing/text_processing/find_and_replace.xml
```diff
@@ -42,6 +43,9 @@
         <param name="is_regex" type="boolean" checked="false" truevalue="-r" falsevalue=""
             label="Find-Pattern is a regular expression" help="see help section for details." />
 
+        <param name="global" type="boolean" checked="true" truevalue="-g" falsevalue=""
```
bgruening: can we make it false to preserve the old default behavior? Otherwise, we can bump the build number.

bernt-matthias: done .. .

bgruening:
Jupp is has a too complicated output section. The linter would need to be updated.

PR #834, opened 2019-01-15
find_and_replace: column splitting bugfix

bernt-matthias:
perl split defaults to split on any space but we need to split by tab


bgruening:
Can we add a test that fails with the old behavior and passes with the new one. This way we make sure it does not break anymore - and the tool number should be bumped.

bernt-matthias:
Test has been added

bernt-matthias:
failing test seems unrelated :(

bgruening:
Yes, the linter can not lint this super complex patter of one of the very old tools ... however, it seems to work.

@bernt-matthias can you review and comment on https://github.com/bgruening/galaxytools/pull/841

I'm unsure if I should like it ... :(

I will then merge all PR together and release a new version of all these tools.

PR #835, opened 2019-01-21
Kpbioteam - Minfi 

kpbioteam:
This PR adds a minfi tool collection, tested and served on Planemo.


tools/minfi/minfi_getsnp.xml
```diff
@@ -0,0 +1,45 @@
...
+    </command>
+    <configfiles>
+        <configfile name="minfi_snp_script"><![CDATA[
+require("minfi", quietly = TRUE)
+
+GRSet <- get(load('$grset'))
+
+snps <- getSnpInfo(GRSet)
+
+write.table(snps, '$table')
```
bgruening: the \t and the quote is lost here ...

PR #837, opened 2019-01-28
wtdbg 2.3

mmiladi:


nsoranzo:
Tests are failing with `wtdbg2: invalid option -- 'Y'`

mmiladi:
Thank you @nsoranzo . Not only the --help arg is changed. I will check the release and test it locally soon.

PR #839, opened 2019-02-04
Peakachu patch for 0.1.0.2

Florian-H-Lab:
Version number was not increased because 0.1.0.2 was initially not working.


bgruening:
@Florian-H-Lab I have updated this as well.

PR #841, opened 2019-02-08
Add ability to digest multiple files

innovate-invent:
This is extremely useful to programmatically combine multiple files without creating a new tool.
The FILENAME variable is likely useless but a running counter or flag can be used to distinguish files.
ex.
```code
BEGIN {FILEINDEX=0}
{FILEINDEX++}
FILEINDEX == 0 { digest first file; next}
{ digest remaining files }
```

The use I intended was to load an array from the first file and inject values from it into the remaining files.


tools/text_processing/text_processing/awk.xml
```diff
@@ -15,15 +15,21 @@
...
         > "$outfile"
 ]]>
     </command>
     <configfiles>
         <configfile name="awk_script">$code</configfile>
     </configfiles>
     <inputs>
-        <param name="infile" format="txt" type="data" label="File to process" />
+        <repeat name="infiles" title="Input files" min="1">
+            <param name="infile" format="txt" type="data" multiple="true" label="File to process" />
```
bgruening: The repeat could be removed correct?

innovate-invent: No, the repeat guarantees a file order, while multiple=true allows consuming a collection directly. Sorry, I totally forgot about running separately for each file. Let me make a change real quick..

bgruening: Uha, I see. Can you tell us the use-case. We might want to find a less hacky way here.

innovate-invent: Have a look at my other commit.

innovate-invent:
Putting the inputs into a conditional would break backwards compatibility and limit its functionality. I am assuming that if you passed a collection to the single input, and a collection to the multi-input, that an instance would be run for every element in the single input. Each instance would share the multi-input datasets. That is how Galaxy would map that right? I'll add a example and a test that demonstrates this.

innovate-invent:
I have expanded on my changes and added a test. I can't find any reasonable documentation related to output collections in a test so the test is likely broken. The cut tool errors out the CI so I have manually run the test in my galaxy instance. There is also the issue of the awk command needing double quotes and I will be surprised if the tool that executes the test properly unescapes XML encoded double quotes. Single quotes seem to have a different meaning in gawk.

innovate-invent:
My most recent change patches a security issue where a user can read the environment variables. Your docker images use this to set the API key and other sensitive data.

I added the ability to pass workflow variables while fixing this issue. Do you know how to disable replacement of ${} from the code input box? It really only should be done in the environment variable box.

innovate-invent:
Patched to deal with https://github.com/galaxyproject/galaxy/issues/7551

jmchilton:
The conditional is best practice, whether it breaks backward compatibility or not. We version tools for a reason, I wouldn't worry too much about breaking compatibility.

tools/text_processing/text_processing/awk.xml
```diff
@@ -1,34 +1,61 @@
...
     </macros>
     <requirements>
         <requirement type="package" version="4.2.0">gawk</requirement>
+        <requirement type="package" version="8.30">coreutils</requirement>
     </requirements>
     <version_command>awk --version | head -n 1</version_command>
     <command>
 <![CDATA[
+        #set $tool_input = 1
+        env -i -S "$envs"
```
nsoranzo: If possible, can you use single quotes instead of double quotes?

tools/text_processing/text_processing/awk.xml
```diff
@@ -1,34 +1,61 @@
...
         awk
             --sandbox
             -v FS='	'
             -v OFS='	'
             --re-interval
             -f "$awk_script"
-            "$infile"
+            #if $infile and not $last
+                tool_input=0
+                "$infile"
```
nsoranzo: Single quotes

tools/text_processing/text_processing/awk.xml
```diff
@@ -1,34 +1,61 @@
...
+            #end if
         > "$outfile"
 ]]>
     </command>
     <configfiles>
         <configfile name="awk_script">$code</configfile>
     </configfiles>
     <inputs>
-        <param name="infile" format="txt" type="data" label="File to process" />
+        <param name="infile" format="txt" type="data" optional="true" label="Single file to process" help="A seperate execution will occur for each provided file"/>
```
nsoranzo:
```suggestion
        <param name="infile" format="txt" type="data" optional="true" label="Single file to process" help="A separate execution will occur for each provided file"/>
```

tools/text_processing/text_processing/awk.xml
```diff
@@ -1,34 +1,61 @@
...
-            "$infile"
+            #if $infile and not $last
+                tool_input=0
+                "$infile"
+            #end if
+            #for group in $infiles
+                tool_input=$tool_input
+                #set $tool_input+=1
+                #for file in $group.infile
+                    "$file"
```
nsoranzo: Single quotes

tools/text_processing/text_processing/awk.xml
```diff
@@ -1,34 +1,61 @@
...
+                #end for
+            #end for
+            #if $infile and $last
+                tool_input=0
+                "$infile"
```
nsoranzo: Single quotes.

tools/text_processing/text_processing/awk.xml
```diff
@@ -1,34 +1,61 @@
...
+            #end if
         > "$outfile"
```
nsoranzo: Single quotes.

tools/text_processing/text_processing/awk.xml
```diff
@@ -1,34 +1,61 @@
...
     </configfiles>
     <inputs>
-        <param name="infile" format="txt" type="data" label="File to process" />
+        <param name="infile" format="txt" type="data" optional="true" label="Single file to process" help="A seperate execution will occur for each provided file"/>
+        <param name="last" type="boolean" checked="false" label="Include the above file last" help="Include the above input after any of the below inputs" />
+        <repeat name="infiles" title="Multiple input files">
+            <param name="infile" format="txt" type="data" multiple="true" label="File to process" />
+        </repeat>
         <param name="code" type="text" area="true" size="5x35" label="AWK Program" help="">
+            <sanitizer sanitize="false" />
```
nsoranzo: I think that changing this `sanitizer` may introduce security issues, can you restore the previous version?

nsoranzo:
@innovate-invent It would be helpful if you can extract the security fix in a separate pull requests for a faster merge, thanks!

tools/text_processing/text_processing/awk.xml
innovate-invent: Can you explain? I need to disable the ${} substitution

innovate-invent:
@jmchilton The conditional would remove important functionality from the tool. Having a single input allows triggering mapping over the tool for each input to the single input.
Possible mappings are now:

Single Input | Multiple input | Output
--- | --- | ---
1 | 1 | 1
0 | m | 1
1 | m | 1
n | m | n

The last one would not be possible with a conditional.

innovate-invent:
@nsoranzo Extracted: https://github.com/bgruening/galaxytools/pull/850

jmchilton:
Could that not just be another case in the conditional?

innovate-invent:
@jmchilton I guess the question is: Why a conditional? Does it merit the increased complexity?

tools/text_processing/text_processing/awk.xml
nsoranzo: I wrongly thought `$code` was ending up in `<command/>`, in `<configfile/>` should be fine.

tools/text_processing/text_processing/awk.xml
```diff
@@ -9,20 +9,28 @@
...
             #for group in $infiles
+                tool_input=$tool_input
+                #set $tool_input+=1
                 #for file in $group.infile
                     "$file"
                 #end for
             #end for
+            #if $infile and $last
+                tool_input=0
+                "$infile"
```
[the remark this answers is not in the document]
innovate-invent: that is an outdated version of the PR :P

innovate-invent:
I'll take a stab at it but Galaxy doesn't really support this in the tool xml. The interface is going to get ugly with numerous nested conditionals and repeat blocks.

innovate-invent:
CI continues to fail for another tool

tools/text_processing/text_processing/awk.xml
```diff
@@ -1,48 +1,126 @@
...
     <requirements>
-        <requirement type="package" version="4.2.0">gawk</requirement>
+        <requirement type="package" version="5.0.1">gawk</requirement>
     </requirements>
     <version_command>awk --version | head -n 1</version_command>
     <command>
 <![CDATA[
+        #set $tool_input = 0
         env -i
+        #for env in $envs
```
bernt-matthias: why not set this variables with the `-v` parameter of awk?

Would `&&` or `;` after `env -i` be good?

tools/text_processing/text_processing/awk.xml
```diff
@@ -1,48 +1,126 @@
...
+            #end for
+        > '$outfile'
 ]]>
     </command>
     <configfiles>
         <configfile name="awk_script">$code</configfile>
     </configfiles>
     <inputs>
-        <param name="infile" format="txt" type="data" label="File to process" />
+        <repeat name="infiles" title="Inputs">
```
bernt-matthias: Is a repeat really necessary. You already can process multiple files with `multiple="true"`

tools/text_processing/text_processing/awk.xml
```diff
@@ -1,48 +1,126 @@
...
+                    <valid initial="string.printable">
+                        <remove value="&apos;" />
+                    </valid>
+                </sanitizer>
+            </param>
+        </repeat>
     </inputs>
     <outputs>
-        <data name="outfile" format_source="infile" metadata_source="infile"/>
+        <data name="outfile" format="txt" />
```
bernt-matthias: I think this is not a good idea, e.g. if you input a tabular.

tools/text_processing/text_processing/awk.xml
bernt-matthias: Ah, OK.

tools/text_processing/text_processing/awk.xml
bernt-matthias: In this case my preference is always to add a `when` for collection input, because IMHO repeats don't work well in workflows, i.e. only for a fixed number of input collections.

tools/text_processing/text_processing/awk.xml
bernt-matthias: What do you mean with dynamic inputs?

bernt-matthias: You can have an output for each case of the input conditional and filter them based on the chosen input option.

bernt-matthias:
My main question is: what would be a use case for this? @bgruening asked this before and I could not find an answer here.

tools/text_processing/text_processing/awk.xml
bernt-matthias: > it boiled down to Galaxy not supporting referring to inputs within a repeat block

Just one more reason to get rid of the repeat block

tools/text_processing/text_processing/awk.xml
bernt-matthias: Somethin along the following lines: In the input you have 

```
<conditional>
  <param name="intype" ... />
  <when "single"></when>
  <when "single"></when>
  <when "multiple"></when>
  <when "list"></when>
  <when "multiple-list"></when>
</conditional>
````

in the output section you have one output for each of these cases:

```
        <data ...>
            <filter>inselect['intype'] == "single" or inselect['intype'] == "multiple"</filter>
        </data>
        <collection ... format_source="...">
            <discover_datasets pattern="__name__" directory="out" visible="false" />
            <filter>inselect['intype'] == "list" or inselect['intype'] == "list_multiple"</filter>
        </collection>
```

innovate-invent: That would prevent n:m:n mappings. It would also have the issue of inconsistent output behavior as only the 'single' and 'multiple' options would allow format_source.

tools/text_processing/text_processing/awk.xml
innovate-invent: Repeat blocks are the only way to have a variable number of inputs.

PR #842, opened 2019-02-15
Column arrange update

wm75:
This tool is currently not working on usegalaxy.eu because of a missing .so file of its pandas dependency.
I think it's better to avoid depending on pandas completely and work on a pure Python file iterator basis instead.
This makes the wrapper leaner and work with arbitrarily large input without any noticeable performance cost. In my testing of a 40,000 rows x 7 columns input, the pure Python implementation here finished faster than pandas could even be imported.

This PR also introduces a new ``--discard``parameter, which makes the tool equivalent to the Cut tool (just working with column names instead of indices), and it fixes an issue with quoting.

### DIFF
--- a/tools/text_processing/column_arrange_by_header/column_arrange.py
+++ b/tools/text_processing/column_arrange_by_header/column_arrange.py
@@ -1,18 +1,30 @@
-#!/usr/bin/env python 
-import pandas as pd
+#!/usr/bin/env python
+
 import argparse
 
 parser = argparse.ArgumentParser()
 parser.add_argument('-i', '--input', help='Tabular Input File Name')
 parser.add_argument('-o','--output', help='Tabular Output File')
-parser.add_argument('-c', '--columns', nargs='+', help='Column Headers to Sort By')
+parser.add_argument(
+    '-c', '--columns', nargs='+', help='Column Headers to Sort By'
+)
+parser.add_argument(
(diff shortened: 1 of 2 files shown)


tools/text_processing/column_arrange_by_header/column_arrange.py
```diff
@@ -1,18 +1,30 @@
...
-# write without index, seperated by tabs
-sorted_table.to_csv(args.output, sep='\t', index=False)
+with open(args.input) as data:
+    hdr = next(data)
+    columns = hdr.rstrip('\n').split('\t')
+    idx = [columns.index(name) for name in args.columns]
+    if not args.discard:
+        idx += [i for i in range(len(columns)) if i not in idx]
+    rearranged_cols = [columns[i] for i in idx]
+    with open(args.output, 'w') as out:
```
bgruening: that with can be moved on indentation to the left

wm75:
Memory: lol (pandas reads in the whole file for processing)
Speed: as I commented above. Python is not *as* slow as many people think, and in this case by the time pandas catches up speed-wise, probably memory becomes limiting.

bgruening:
> Memory: lol (pandas reads in the whole file for processing)

maybe this implementation, but in general pandas can be used to not due this ...

Eitherway, I like it. Let's get it in. I also fixed hopefully the old version.

PR #847, opened 2019-02-28
Use profile version in text_processing tools

mvdbeek:
To get detect_errors="exit_code" and other good things.

### DIFF
--- a/tools/text_processing/column_arrange_by_header/columnArrange.xml
+++ b/tools/text_processing/column_arrange_by_header/columnArrange.xml
@@ -1,8 +1,8 @@
 <tool id="bg_column_arrange_by_header" name="Column arrange" version="0.2" profile="16.04">
     <description>by header name</description>
-    <command interpreter="python">
+    <command>
     <![CDATA[
-        column_arrange.py
+        python '$__tool_directory__'/column_arrange.py
           --columns
         #for token in $rep_param_columns:
             '$token.param_column'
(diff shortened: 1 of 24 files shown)


mvdbeek:
```
+Linting tool /home/travis/build/bgruening/galaxytools/tools/text_processing/text_processing/cut.xml
Applying linter tool_xsd... FAIL
.. ERROR: Invalid XML found in file: tmpqjKCNx. Errors [/tmp/tmpqjKCNx:72:0:ERROR:SCHEMASV:SCHEMAV_ELEMENT_CONTENT: Element 'conditional': This element is not expected. Expected is ( action ).]
```
That might be a valid error, I have no clue if this really works, it looks complicated.

bgruening:
It really is complicated a relict I think from the very early days. I think it still works but it was never documented. Maybe @blankenberg knows more?

bgruening:
ping @blankenberg

bernt-matthias:
Some feedback for this old issue. I try to fix the linter issue over here: https://github.com/galaxyproject/galaxy/pull/15508

I agree that output actions are barely documented (a perfect example that non-trivial features that are not documented are not used xref https://github.com/galaxyproject/galaxy/pull/15027). But I also do not see how this is doing anything useful. 

```
                   <conditional name="delimiter">
                        <when value="T">
```

There is no value `"T"` for the delimiter parameter (and never was as far as I have seen). So this never did anything.

The second part that (if I get it right) tries to set column metadata which I don't see working correctly given that we remove columns. But due to the lack of docs I'm not sure. 

My suggestion would be to get rid of this or put some serious work into output actions (understanding code, adding docs and tests to the Galaxy code base .. I kind of feel tempted).

PR #850, opened 2019-04-02
Extracted security fix for reading system ENVIRON

innovate-invent:


tools/text_processing/text_processing/awk.xml
```diff
@@ -5,10 +5,12 @@
     </macros>
     <requirements>
         <requirement type="package" version="4.2.0">gawk</requirement>
+        <requirement type="package" version="8.30">coreutils</requirement>
```
nsoranzo: Since you are not using the `-S` option of `env` which was introduced in coreutils 8.30, maybe we can leave this additional requirement for the other PR?
Otherwise this needs a tool version bump, which means the currently installed version of the tool will not be fixed (i.e. a new version will be installed).

innovate-invent: Done

bgruening:
Thanks!

bgruening:
And deployed to the TS. Thanks @innovate-invent!

PR #851, opened 2019-04-03
Add only-matching parameter

innovate-invent:


innovate-invent:
CI fails for cut tool and not this one

tools/text_processing/text_processing/grep.xml
```diff
@@ -59,6 +61,7 @@
             <option value="-i">case insensitive</option>
             <option value="">case sensitive</option>
         </param>
+        <param name="only_matching" type="boolean" argument="only-matching" label="Show only nonempty parts of lines that match" truevalue="-o" falsevalue="" checked="false" help="This outputs only the match" />
```
bgruening, line 64: `argument` should be `-o` isn't it?

innovate-invent: either: https://linux.die.net/man/1/grep

bgruening: Than the `--` are missing :)

innovate-invent: This PR sat so long I deleted the original repo. I won't be able to edit the PR.

PR #854, opened 2019-04-16
Vina params

reskyner:
Addition of a tool to calculate the box size for an input ligand (using confining box), with possibility of extending box in x, y or z direction. Creates parameter file to put into vina tool.


chemicaltoolbox/autodock_vina/prepare_box/prepare_box.xml
```diff
@@ -0,0 +1,37 @@
+<tool id="prepare_box" name="Calculate the box parameters for an AutoDock Vina job from an input mol file (confounding box)" version="0.1.0">
+    <requirements>
+        <requirement type="package">rdkit</requirement>
```
bgruening: please include versions here as well for reproducibility

chemicaltoolbox/autodock_vina/prepare_box/prepare_box.xml
```diff
@@ -0,0 +1,37 @@
...
+        <requirement type="package">numpy</requirement>
+    </requirements>
+    <command detect_errors="exit_code"><![CDATA[
+        python $__tool_directory__/calc_vina_box_params.py --ligand '$input1' --config '$output1' --bufx 1 --bufy 2 --bufz 3
```
bgruening: please also put tool_directory/calc... in single quotes. It can be that in some galaxy installation there is a space in the path ... who knows ;)

chemicaltoolbox/autodock_vina/prepare_box/prepare_box.xml
```diff
@@ -0,0 +1,37 @@
...
+    ]]></command>
+    <inputs>
+        <param type="data" name="input1" format="mol" />
```
bgruening: only mol or also other filetypes are supported?

chemicaltoolbox/autodock_vina/prepare_box/prepare_box.xml
```diff
@@ -0,0 +1,37 @@
...
+        <data name="output1" format="txt" />
+    </outputs>
+    <tests>
+        <test>
+            <param name="input1" value="NUDT5A-x0114_2.mol"/>
+            <output name="output1" file="box_params.txt"/>
+        </test>
+    </tests>
+    <help><![CDATA[
+        usage: calc_vina_box_params.py [-h] [--ligand LIGAND_PATH] [--config OUTPUT]
```
bgruening: Please add some text that indicates what this tool is doing, this will help our users.

chemicaltoolbox/autodock_vina/prepare_box/prepare_box.xml
```diff
@@ -0,0 +1,37 @@
...
+  -h, --help            show this help message and exit
+  --ligand LIGAND_PATH
+  --config OUTPUT
+  --tol TOLERANCE
+  --exh EXHAUSTIVENESS
+  --bufx BUFX
+  --bufy BUFY
+  --bufz BUFZ
+
+    ]]></help>
```
bgruening: maybe adding a citation for rdkit?

chemicaltoolbox/autodock_vina/prepare_box/prepare_box.xml
simonbray: here we should parameterise the buffer values i.e. --bufx $bufx etc.

chemicaltoolbox/autodock_vina/prepare_box/prepare_box.xml
```diff
@@ -0,0 +1,37 @@
...
+        <requirement type="package">rdkit</requirement>
+        <requirement type="package">numpy</requirement>
+    </requirements>
+    <command detect_errors="exit_code"><![CDATA[
+        python $__tool_directory__/calc_vina_box_params.py --ligand '$input1' --config '$output1' --bufx 1 --bufy 2 --bufz 3
+    ]]></command>
+    <inputs>
+        <param type="data" name="input1" format="mol" />
+    </inputs>
+    <outputs>
```
simonbray: and then add something like `<param name="bufx" type="float" value="0" label="x-axis buffer" help="Expand box dimensions by this amount on the x-axis"/>` and similar for bufy and bufz and exhaustiveness

chemicaltoolbox/autodock_vina/prepare_box/calc_vina_box_params.py
```diff
@@ -0,0 +1,59 @@
...
+            """.format(box_dims[0], box_dims[1], box_dims[2], center[0], center[1], center[2], options.exhaustiveness)
+        )
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser()
+    parser.add_argument('--ligand', dest='ligand_path')
+    parser.add_argument('--config', dest='output')
+    parser.add_argument('--tol', dest='tolerance', default=0, type=float)
+    parser.add_argument('--exh', dest='exhaustiveness', default=10, type=float)
```
simonbray: we don't need both tolerance and the 3 buffer values

chemicaltoolbox/autodock_vina/prepare_box/calc_vina_box_params.py
```diff
@@ -0,0 +1,59 @@
...
+
+    # change tuples to arrays
+    coords1 = np.array(params[0])
+    coords2 = np.array(params[1])
+
+    # get the centre of the box
+    center = np.mean((coords1, coords2), axis=0)
+
+    # calculate box dimensions
+    dims = np.abs(coords1 - coords2) + (options.tolerance,) * 3
```
simonbray: we can remove the `(options.tolerance,) * 3` thing because this is the same as adding the buffers

chemicaltoolbox/autodock_vina/prepare_box/prepare_box.xml
simonbray: oh, and the exhaustiveness

chemicaltoolbox/autodock_vina/prepare_box/calc_vina_box_params.py
reskyner: Resolved by: b046d0c685b7638818660cadfb7d1c16e7d97545

chemicaltoolbox/autodock_vina/prepare_box/calc_vina_box_params.py
reskyner: Resolved by: b046d0c685b7638818660cadfb7d1c16e7d97545

chemicaltoolbox/autodock_vina/prepare_box/prepare_box.xml
reskyner: partially resolved by: f6b92ec865f2975a6772b66ba5eb943f54057122

Will regenerate the xml file when I've finished editing code.

simonbray: It's a good idea to write a help section which is tailored for the graphical interface though, rather than just using one from argparse ([example](https://github.com/bgruening/galaxytools/blob/master/chemicaltoolbox/openbabel/ob_convert.xml)) `planemo tool_init` is a good start but the tool generally needs some further editing.

chemicaltoolbox/autodock_vina/prepare_box/prepare_box.xml
reskyner: resolved by c0d9ecf83a9a57b4011d4e47769ed950fd5ee798

chemicaltoolbox/autodock_vina/prepare_box/prepare_box.xml
reskyner: resolved by c0d9ecf83a9a57b4011d4e47769ed950fd5ee798

chemicaltoolbox/autodock_vina/prepare_box/prepare_box.xml
reskyner: probably resolved by c0d9ecf83a9a57b4011d4e47769ed950fd5ee798

chemicaltoolbox/autodock_vina/prepare_box/prepare_box.xml
reskyner: resolved by c0d9ecf83a9a57b4011d4e47769ed950fd5ee798

chemicaltoolbox/autodock_vina/prepare_box/prepare_box.xml
reskyner: resolved by c0d9ecf83a9a57b4011d4e47769ed950fd5ee798

chemicaltoolbox/autodock_vina/prepare_box/prepare_box.xml
reskyner: Currently only mol. This could be expanded later, but I think I need a better understanding of how to handle multiple file types with rdkit. I will look for an example in the current tools if we want this before the tool goes into the ToolShed.

chemicaltoolbox/autodock_vina/prepare_box/prepare_box.xml
reskyner: resolved by ce67e4feeb7c8b593ce563f893b714d6c4cbdfcc

chemicaltoolbox/autodock_vina/prepare_box/prepare_box.xml
bgruening: Galaxy knows the filetype, so we can always pass the filetype into the python script via a parameter if you like. `$input1.ext` for example gives you the Galaxy filetype extension, e.g. (sdf, mol, cml ...)

simonbray: We can write a function very easily to apply the correct rdkit function depending on the file extension, but I guess it should go into a shared library which every rdkit-based tool can access. Not sure of the best way to implement that.

chemicaltoolbox/autodock_vina/prepare_box/prepare_box.xml
```diff
@@ -0,0 +1,89 @@
...
+        <data name="output1" format="txt" />
+    </outputs>
+    <tests>
+        <test>
+            <param name="input1" value="NUDT5A-x0114_2.mol" />
+            <param name="bufx" value="1" />
+            <param name="bufy" value="2" />
+            <param name="bufz" value="3" />
+            <output name="output1" file="box_params.txt" />
```
simonbray: if you add `<param name="exh" value="10" />` the test should pass

chemicaltoolbox/autodock_vina/prepare_box/prepare_box.xml
```diff
@@ -0,0 +1,89 @@
...
+    <command detect_errors="exit_code"><![CDATA[
+        python '$__tool_directory__/calc_vina_box_params.py' --ligand '$input1' --config '$output1' --bufx '$bufx' --bufy '$bufy' --bufz '$bufz' --exh '$exh'
+    ]]></command>
+    <inputs>
+        <param type="data" name="input1" format="mol" label="input ligand mol file" help="The input ligand (mol file)"/>
+        <param name="bufx" type="float" value="0" label="x-axis buffer" help="the buffer in the x direction (float - in angs.)"/>
+        <param name="bufy" type="float" value="0" label="y-axis buffer" help="the buffer in the y direction (float - in angs.)"/>
+        <param name="bufz" type="float" value="0" label="z-axis buffer" help="the buffer in the z direction (float - in angs.)"/>
+        <param name="exh" type="float" value="0" label="exhaustiveness" help="The number of poses to return from docking job"/>
+    </inputs>
```
simonbray: `type="int"` for exhaustiveness I think

chemicaltoolbox/autodock_vina/prepare_box/calc_vina_box_params.py
```diff
@@ -0,0 +1,69 @@
...
+    This tool calculates a confounding box around an input ligand (mol file), and uses it to
+    generate the input parameters for an autodock vina job. The output file can be fed into
+    the autodock vina tool as an alternative to creating the parameter file manually. 
+    
+    Optionally, you can include a 'buffer' in each of the x,y and z directions (in angstroms),
+    which will be added to the confounding box in the appropriate direction.
+    """)
+    parser.add_argument('--ligand', dest='ligand_path', help='The input ligand (mol file)')
+    parser.add_argument('--config', dest='output', help='The output file containing calculated params (txt)')
+    parser.add_argument('--exh', dest='exhaustiveness', default=10, type=float, help='The number of poses '
```
simonbray: `type=int` here as well

bgruening:
Green! :)

chemicaltoolbox/autodock_vina/prepare_box/prepare_box.xml
reskyner: Should we create an issue for this? Happy to work with Simon to develop something in the future, if that's something useful we can help to contribute (we use a lot of rdkit stuff, so would be useful for us).

bgruening: Yes, please always helps!

I guess a very tiny package with rdkit-helper functions.

simonbray: I created an issue: https://github.com/bgruening/galaxytools/issues/855

reskyner: Thanks!

PR #857, opened 2019-04-23
Chipseeker: small fixes

mblue9:
- Found that chipseeker assumes a header is present in the input peaks file, unless the file has "bed" or related in the name (see here header https://github.com/GuangchuangYu/ChIPseeker/blob/master/R/readPeakFile.R#L60-L61), which was resulting in 1 peak/row of input not being output by this tool. So made an option for header, with default no header (assume most people would input BED)
- Fixed 2 plots that weren't being output in the PDF
- Fixed link to repo in shed.yml

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -70,7 +70,7 @@ jobs:
 
 before_install:
   - export GALAXY_REPO=https://github.com/galaxyproject/galaxy
-  - export GALAXY_RELEASE=release_18.09
+  - export GALAXY_RELEASE=release_19.01
   - export PLANEMO_CONDA_PREFIX="$HOME/conda"
   - unset JAVA_HOME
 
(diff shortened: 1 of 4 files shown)


mblue9:
This is giving the error 
```
|  Error: package or namespace load failed for \u2018ChIPseeker\u2019 in loadNamespace(i, c(lib.loc, .libPaths()), versionCheck = vI[[i]]):
|  there is no package called \u2018GenomeInfoDbData\u2019
```

nsoranzo:
@mblue9 This is probably due to https://github.com/bioconda/bioconda-recipes/issues/13846

mblue9:
Ok thanks for the info @nsoranzo

bgruening:
Thanks @mblue9!

mblue9:
Thanks for fixing!

PR #866, opened 2019-06-05
add dollar, fix small bug

bimbam23:


bgruening:
Merci @bimbam23! But you should really use methyldackle instead :)

PR #872, opened 2019-06-20
nanopolish: use tool version, support h5 ext

mmiladi:


mmiladi:
Btw where should the poly toolshed repo yaml be added?

tools/nanopolish/nanopolish_methylation.xml
```diff
@@ -7,7 +7,7 @@
     <command detect_errors="exit_code"><![CDATA[
         ln -s '$input_merged' reads.fasta && 
         
-        #if $input_reads_raw.extension == 'fast5':
+        #if $input_reads_raw.extension == 'fast5' or $input_reads_raw.extension == 'h5':
```
bgruening: the other, probably more correct version would be to not allow this tool to consume h5 files, but only fast5 ones

mmiladi: There is no fast5 format in galaxy! So an uploaded fast5 can never get a "fast5" extension I guess..

PR #874, opened 2019-07-04
Maybe this is more readily understandable?

bwlang:
could also be "EXCEPTING" or "BUT NOT some"


bgruening:
I like it. @qiagu ?

qiagu:
Sure. I like the change.

qiagu:
@bwlang Thanks for making the tools better.

PR #875, opened 2019-07-06
WIP: Update version Flye

willemdek11:


bgruening:
I guess you also need to update the macro file. Thanks @willemdek11!

AnnaSyme:
If it's ok with you @willemdek11 @bgruening could we update flye to 2.6? I can try sending a pull request with changed flye.xml as well as macros.xml (although I'm still learning how). Also: could we re-name the tool "Flye assembly" or something similar with the "flye" part in the name (not just the id)? Makes it easier to recognise in the tool panel. (Currently named "Assembly")

bgruening:
@willemdek11 yes please do so. Thanks!

willemdek11:
Hey @AnnaSyme, @bgruening & @mdcjansen working on it!

PR #876, opened 2019-07-08
update to version 1.0.7.10 with galaxy-ml conda denpendency

qiagu:
Wait the `galaxy-ml` in conda to be ready.
https://github.com/bioconda/bioconda-recipes/pull/16226


qiagu:
@bgruening could you please restart Travis?

nsoranzo:
@qiagu Jobs are failing due to this error:

Executing command: /home/travis/conda/bin/conda create -y --quiet --override-channels --channel iuc --channel conda-forge --channel bioconda --channel defaults --name mulled-v1-ab8b16d251512b10a3866bce4d1190423bd88d4e3e1d85cd491e7e19a9c467bc python=3.6 Galaxy-ML=0.7.5
UnsatisfiableError: The following specifications were found to be in conflict:
  - galaxy-ml=0.7.5
Use "conda search <package> --info" to see the dependencies for each package.

qiagu:
@nsoranzo Thanks for stopping by this PR. `~/anaconda3/bin/conda create -y --quiet --override-channels --channel iuc --channel conda-forge --channel bioconda --channel defaults --name mulled-v1-ab8b16d251512b10a3866bce4d1190423bd88d4e3e1d85cd491e7e19a9c467bc python=3.6 Galaxy-ML=0.7.5` works in my local computer. Could you please try restarting the Travis  after a few seconds, as the conda PR was merged right this morning.

nsoranzo:
@qiagu I already did!

qiagu:
@nsoranzo The conda building looks working for now. Thanks!

qiagu:
@nsoranzo @bgruening @anuprulez @@khanteymoori  I'm not hurrying in having these tools merged. If you have any suggestion or concern, please feel free to have a discussion here. Thanks.

qiagu:
Just over time limit for Travis. The tools look okay, but the gpu environment is a problem. I have never succeeded yet.

qiagu:
@bgruening Could you make the test wall time longer? Now we have couple of more tools and more tests.

khanteymoori:
Thanks for new update:
1. It seems Imputer from preprocessing does not work.
2. There is no option to load compressed transformer file (in preprocessing) and apply to new data
3. There is no advanced option to define minmaxscalar range in preprocessing tool, and we should always use default values (min=0, max=1). most of the time we need to change the range.
4. There is no advance options for "kernel center", so the button should be disable when this method is chosen.
5. Sometimes when the data have a header, we need header in the results table too. It is better if there is  an option  to create the preprocessing results with/without header

qiagu:
Thanks for recording those issues! @khanteymoori 
>1. It seems Imputer from preprocessing does not work.

I suggest removing this guy because it's DEPRECATED in latest versions. What do others think?

>2. There is no option to load compressed transformer file (in preprocessing) and apply to new data

I'm afraid that this issue can't be addressed in this PR. Please move it to a new ticket for future reference.

>3. There is no advanced option to define minmaxscalar range in preprocessing tool, and we should always use default values (min=0, max=1). most of the time we need to change the range.

Done.

>4. There is no advance options for "kernel center", so the button should be disable when this method is chosen.

Where is the "kernel center"?

>5. Sometimes when the data have a header, we need header in the results table too. It is better if there is an option to create the preprocessing results with/without header

Done except for the `PolynomialFeatures`.

tools/sklearn/estimator_attributes.xml
```diff
@@ -156,6 +167,11 @@ else:
             <param name="attribute_type" value="get_params"/>
             <output name="outfile" value="get_params.tabular"/>
         </test>
+        <test>
```
bgruening: @khanteymoori thats what I told you ... more tests are always good!

PR #878, opened 2019-07-29
Update Trim Galore to 0.6.3

abretaud:
Hi,
I was asked to update trim galore on our instance, so here it is!


abretaud:
Yep I agree, nothing very important in the changelog, and `--cores` is a bit confusing on the real number of cores used indeed

bgruening:
Merci to both of you!

PR #879, opened 2019-07-30
Bismark: fix error with gzipped fastq + update to 0.22.1

abretaud:
Some of my users are getting errors running bismark, it turns out that it comes from using gzipped fastq files as input: bismark uses the file extension to detect gzip compression (`.gz`) and as we gave them the path to `.dat` files, gzipped files are wrongly considered not gzipped. So here's a fix, using symlinks in work dir.
I also update dependencies to bismark 0.22.1 and bowtie2 2.3.5. There's a new hisat2 mode for bismark, but I don't have too much time to add this to the wrapper atm


abretaud:
Not doing methylation myself, just fixing other people's errors :)
Is bismark outdated now, or just that bwa-meth and methyldackel are an alternative?

bgruening:
It's an alternative. Way faster and imho more robust.

PR #881, opened 2019-07-31
Bismark: fix element_identifier sanitizing

abretaud:
One last (:crossed_fingers:) PR to fix an issue with latest changes in bismark: errors when element_identifier contains spaces


bgruening:
Ah, now its failing on master with a test error :(

abretaud:
Ah, I think it comes from the execution times that is printed in summary.txt output. Most of the time it's 00:00:00, but for this build there was some 00:00:01 (https://travis-ci.org/bgruening/galaxytools/jobs/565981497#L2686) which made a longer diff.
Did you try just rerunning the travis job? If we're lazy and lucky it could work I guess. Otherwise, I can make a PR to adapt the test-data/lines_diff number

bgruening:
restarted now multiple times, no luck :(

abretaud:
Arf ok, I look into it

PR #883, opened 2019-08-05
add ChEMBL search tool

simonbray:


chemicaltoolbox/chembl/chembl.py
```diff
@@ -0,0 +1,102 @@
+from chembl_webresource_client.new_client import new_client
+import argparse
+
+def open_file(filename):
+    with open(filename) as f:
+        return f.read().split('\n')[0]
```
bgruening: Can you replace this function by `file.readline()`?

chemicaltoolbox/chembl/chembl.py
```diff
@@ -0,0 +1,102 @@
...
+    """
+    Return only compounds which are approved drugs
+    """
+    return mols.filter(max_phase=4)
+
+def filter_biotherapeutic(mols):
+    """
+    Return only biotherapeutic molecules
+    """
+    return mols.filter(biotherapeutic__isnull=False)
```
bgruening: is it really two `__`?

chemicaltoolbox/chembl/chembl.xml
```diff
@@ -0,0 +1,118 @@
+<tool id="chembl" name="Search ChEMBL database" version="0.1.0">
+    <description>for compounds which are similar to a SMILES string</description>
+    <requirements>
+        <requirement type="package" version="0.9.31">chembl_webresource_client</requirement>
+    </requirements>
+    <command detect_errors="exit_code"><![CDATA[
+        python -W ignore '$__tool_directory__/chembl.py'
```
bgruening: What does this `-W ignore` filter?

chemicaltoolbox/chembl/chembl.py
```diff
@@ -0,0 +1,102 @@
...
+
+def open_file(filename):
+    with open(filename) as f:
+        return f.read().split('\n')[0]
+
+def get_smiles(res):
+    """
+    Get a list of SMILES from function results
+    """ 
+    smiles = []
```
bgruening: Should that be a set to exclude duplicates?

chemicaltoolbox/chembl/chembl.xml
simonbray: The python package raises a warning which I'd like to hide from the user. That flag suppresses it.

chemicaltoolbox/chembl/chembl.py
simonbray: yes...

chemicaltoolbox/chembl/chembl.py
simonbray: Will do.

chemicaltoolbox/chembl/chembl.py
simonbray: changed

bgruening:
Merci!

PR #885, opened 2019-08-12
add fpocket

simonbray:


chemicaltoolbox/fpocket/fpocket.xml
```diff
@@ -0,0 +1,140 @@
...
+    <description>- find potential binding sites in protein structures</description>
+    <requirements>
+        <requirement type="package" version="3.1.3">fpocket</requirement>
+    </requirements>
+    <command detect_errors="exit_code"><![CDATA[
+
+        ln -s '$input' ./input.pdb &&
+        fpocket
+            -f ./input.pdb
+            #if $inp.pocket_type == 'channel':
```
bgruening: just checking, `small_mol` is not needed here?

bgruening: this could be a elif ...

simonbray: no, small_mol is the default

simonbray: will update to elif

PR #888, opened 2019-08-21
Bismark: add --score-min option

abretaud:
One of our user asked for this option, so here it is


bgruening:
This 'rent-an-anthony-service' is awesome! I want to have it as well!

abretaud:
You should to subscribe to our new Anthony Prime program!

PR #889, opened 2019-08-22
refactor keras model config and add visualization tool

qiagu:
- refactor keras model config tool by grouping layer key words arguments.
- add visualization tool to plot `feature_importances`, `pr_curve`, `roc_curve`, `rfecv_gridscores` and `learning_curve`.
- change the saving of prediction result from vcf file to batch mode.


tools/sklearn/model_prediction.py
```diff
@@ -137,54 +137,62 @@ def main(inputs, infile_estimator, outfile_predict,
...
+        steps_done = 0
+
+        # TODO: multiple threading
+        while steps_done < len(gen_flow):
+            index_array = next(gen_flow.index_generator)
+            batch_X = gen_flow._get_batches_of_transformed_samples(
+                index_array)
+
+            if params['method'] == 'predict':
+                batch_preds = estimator.predict(
```
anuprulez: Why a `batch_predictor` being used? Because the data for prediction may not fit in memory at once while predicting?

qiagu: All Keras prediction API are built on batches. Data_generator kicks in here to perform data conversion (from vcf info to 2-D array) as prediction goes. Basically, to be MEM efficient, as the tensorflow itself is a MEM monster.

tools/sklearn/model_prediction.py
```diff
@@ -137,54 +137,62 @@ def main(inputs, infile_estimator, outfile_predict,
...
-                # probibilities will 1, the classification is considered as
-                # multi-class problem, otherwise, we take it as multi-label.
-                act = getattr(estimator.model_.layers[-1], 'activation', None)
-                if act and act.__name__ == 'softmax':
-                    classes = preds.argmax(axis=-1)
-                else:
-                    preds = (preds > 0.5).astype('int32')
+        variants = np.array(pred_data_generator.variants)
+        # predict 1600 sample at once then write to file
+        gen_flow = pred_data_generator.flow(batch_size=1600)
```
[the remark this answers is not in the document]
qiagu: Arbitrary number actually. If too big, write less, but hold more MEM, if too small, write overhead comes in due to frequent writes.

tools/sklearn/ml_visualizations.xml
```diff
@@ -0,0 +1,121 @@
+<tool id="ml_visualizations" name="Machine Learning Visualizations" version="@VERSION@">
+    <description>include various types of plotting for machine learning</description>
```
qiagu: This tool name may be too ambitious. A more proper name is appreciated. @anuprulez @bgruening

bgruening:
Gracias!

qiagu:
@bgruening Thanks!

Issue #886, opened 2019-08-17
split_file_to_collection for txt files


bernt-matthias:
Would be nice to have an option to split txt files. Then the workflow outlined here 

https://github.com/galaxyproject/galaxy/pull/8336#issuecomment-522222395

would work. Currently it does not without a manual change of the data type of `extract element identifiers` (of collection) which outputs txt .. but  split_file_to_collection requires tabular.

bernt-matthias:
ping https://github.com/DarianHole

bernt-matthias:
Just wanted to let you know that I'm working on the necessary fix for https://github.com/galaxyproject/galaxy/pull/8336#issuecomment-522222395

PR #892, opened 2019-08-27
split_file_to_collection: add txt

bernt-matthias:
fixes https://github.com/bgruening/galaxytools/issues/886


bgruening:
> Failed to find identifier [test_1.txt] for testing, tool generated collection elements [[]]

bernt-matthias:
On my local planemo tests for some reason my local python 2.7 is executed which does not know about `math.inf`

bgruening:
```
--- local_file
+++ history_data
@@ -1,4 +1,3 @@
-#This is a file
 #file   data
 foo2.mgf	bar2
 foo.mgf	bar4
```

Mh, math should be part of the standard lib :(

bernt-matthias:
Hey @bgruening what do you think about the generic splitter. 

I would have liked to set the format identical to the input, but currently only dynamic detection works: https://github.com/galaxyproject/galaxy/issues/2431

Also a macro for the regexp validator. If you like we may also remove the txt case (since generic defaults to this behaviour). 

Also was thinking that I would like to remove the extension from the element identifiers.

tools/text_processing/split_file_to_collection/split_file_to_collection.xml
```diff
@@ -1,6 +1,19 @@
-<tool id="split_file_to_collection" name="Split file" version="0.1.1">
+<tool id="split_file_to_collection" name="Split file" version="0.2.0">
     <description>to dataset collection</description>
     <macros>
+        <xml name="regex_sanitizer">
```
bgruening: I like that!

tools/text_processing/split_file_to_collection/split_file_to_collection.xml
```diff
@@ -73,6 +91,8 @@
                 <option value="fastq">FASTQ</option>
                 <option value="tabular">Tabular</option>
                 <option value="fasta">FASTA</option>
+                <option value="txt">Text files</option>
+                <option value="generic">Genric</option>
```
bgruening: This should be generic I guess ... maybe a little bit more explanation ...?
Also you might want to add an example to the help?

bernt-matthias: Sure. I extended the help text. Is this sufficient in your opinion?

bgruening:
Thanks @bernt-matthias!

PR #893, opened 2019-08-28
Bismark: fix empty unmapped/ambiguous output, add --pbat option

abretaud:
As I was asked for it, yet another PR for bismark to fix empty unmapped and ambiguous output and add the --pbat option.
With the --pbat option, input must be unzipped before being used by bismark, so I had to hack a little...


bgruening:
Thanks :)

PR #898, opened 2019-09-06
Split with maintined ordering... doesn't

erasche:
image

Did not execute the python but it's probably correct. The original code is also a bit suboptimal, it opens N file handles at once which can fail in a wide array of scenarios where filehandle limits are imposed. This might fix the ordering issue.


bgruening:
Should we add a test for the odering?

erasche:
added! :)

bgruening:
Merci, tests are failing :(

PR #901, opened 2019-09-10
Montage fixes + convert resize

erasche:
Some small fixes from the Circos work.

Made for a really nice workflow, resize images + paste them together:

image

Resulting in this lovely one:
image


tools/image_processing/graphicsmagick/montage.xml
```diff
@@ -1,41 +1,49 @@
...
+    -background none
+    -tile ${width}x
+    temp.png
 ]]>
     </command>
 
     <inputs>
-        <param format="jpg" name="input" type="data" label="Existing Image" />
-        <param format="jpg" name="input_new" type="data" label="New Image" />
+        <param format="jpg,png,bmp,gif,svg,eps,tiff" name="input" type="data" multiple="true" label="Images" />
```
bgruening: I like that, but should this be an option? X inputs --> 1 output or X inputs --> X outputs?

erasche: I had not considered that use case. Do you think it's common, or maybe should the user use something like a collection splitting operation ahead of this tool?

bgruening: I guess this can work!

PR #904, opened 2019-09-18
Salmon index version update

ieguinoa:
v0.14.1 of salmon uses a new version of indexes (q6).
The index in the prebuilt_index/ dir is correct for this version (version info can be found in https://github.com/bgruening/galaxytools/blob/master/tools/salmon/test-data/cached_locally/prebuilt_index/header.json), but the cached loc file lists it as q5....that's why the test using the index is passing ok.
I will also update the DM in iuc.


bgruening:
Mh, no clue why this is failing :(

bgruening:
@mvdbeek could you have a look here? The last 2 changes indicate that there is a bigger problem on the upcoming Galaxy 19.09 release. Are you aware of any changes to location file handlng?

PR #905, opened 2019-09-23
Updated vina docking plus other new tools

tdudgeon:
chemicaltoolbox/autodock_vina/docking/docking.xml updated to support docking of multiple ligands in a single task. SDF is now the only supported output.

chemicaltoolbox/rdock/rbdock.xml and chemicaltoolbox/rdock/rbcavity.xml tools for docking with rDock.

chemicaltoolbox/rdkit/splitter.xml for splitting SDF into chunks using RDKit.

chemicaltoolbox/rdkit/sucos.xml, chemicaltoolbox/rdkit/sucos_cluster.xml and chemicaltoolbox/rdkit/sucos_max.xml for SuCOS scoring or fragment screening ligands.


chemicaltoolbox/autodock_vina/docking/docking.xml
```diff
@@ -1,4 +1,4 @@
-<tool id="docking" name="Docking" version="0.2.1">
+<tool id="docking030" name="Docking030" version="0.3.0">
```
bgruening: the ID needs to stay the same, otherwise Galaxy does not know that it is the same tool

bgruening: I would also keep the name

chemicaltoolbox/autodock_vina/docking/docking.xml
```diff
@@ -8,49 +8,58 @@
...
+        #if $ligands.is_of_type("sdf")
+            obabel -isdf '$ligands' -O ligand.pdbqt -m -p $ph_value &&
         #else
-            && mv ./output1.dat '$file_output1'
-            && mv ./output2.dat '$file_output2'
+            ln -s '$ligands' ligand1.pdbqt &&
         #end if
+        rm -rf output &&
+        mkdir output &&
+        for file in ligand*.pdbqt;
```
bgruening: we should use gnu-parallel here to run these in parallel and not sequentially.

chemicaltoolbox/rdkit/environment.yml
```diff
@@ -0,0 +1,6 @@
+name: rdkit_sucos
```
bgruening: this file is not needed

chemicaltoolbox/rdkit/splitter.py
```diff
@@ -0,0 +1,51 @@
+import argparse
```
bgruening: I guess this tool is not needed as it can be done by other tools?

chemicaltoolbox/rdkit/sucos.py
```diff
@@ -0,0 +1,202 @@
+#!/usr/bin/env python
```
bgruening: should the sucos tool be in its own subfolder?

chemicaltoolbox/rdkit/sucos_cluster.xml
```diff
@@ -0,0 +1,65 @@
...
+
+Molecules such as an SD file dataset in history.
+
+**Output**
+
+A series of SDF files, one for each cluster containing the molecules in that cluster
+
+    ]]></help>
+    <citations>
+        <citation type="doi">10.26434/chemrxiv.8100203.v1</citation>
```
bgruening: You should use the macro citation that you already have ...

chemicaltoolbox/autodock_vina/docking/docking.xml
tdudgeon: I forgot to revert the names - now they are back to the originals.

chemicaltoolbox/autodock_vina/docking/docking.xml
tdudgeon: I need to investigate this more.

chemicaltoolbox/rdkit/environment.yml
tdudgeon: removed

chemicaltoolbox/rdkit/splitter.py
tdudgeon: I need to look at the alternatives

chemicaltoolbox/rdkit/sucos.py
tdudgeon: moved to sucos dir.

chemicaltoolbox/rdkit/sucos_cluster.xml
tdudgeon: done

chemicaltoolbox/rdkit/splitter.py
tdudgeon: I don't see any exact alternatives at present. 
The `Split file` tool has a few built in formats that it understands, but SDF is not one of them. If you use the `Text files` option for the format it doesn't do the splitting correctly (in fact it doesn't split at all you get n identical files.

The `Prepare ligands for docking` tool could be extended to handle this. Currently it seem to output each record as a separate file, but it might get a bit confusing as some of the formats it supports (e.g. molfile) do not support multiple records in a single file. 

My `Splitter` tool is based on a much more comprehensive tool that can do much more than splitting (e.g filtering, transforming data etc.) which could also be added at some stage and would be useful. See https://github.com/InformaticsMatters/pipelines-utils-rdkit/blob/master/src/python/pipelines_utils_rdkit/filter.py

So unless I'm not seeing some obvious tool that can handle this then I prefer to keep the splitter tool and then to extend it at a later stage.

bgruening: Hi Tim, for me the tool is working, but you need to use "Generic", there is even a test included that makes sure it works for SD files. Here is a history for you to double check: https://usegalaxy.eu/u/bgruening/h/file-splitting-for-tim.

> The Prepare ligands for docking tool could be extended to handle this. Currently it seem to output each record as a separate file, but it might get a bit confusing as some of the formats it supports (e.g. molfile) do not support multiple records in a single file.

Yes we could do this, but in general we try to follow Unix philosophy as much as we can. Simple tools that do one thing well - exceptions are usability.

> My Splitter tool is based on a much more comprehensive tool that can do much more than splitting (e.g filtering, transforming data etc.) which could also be added at some stage and would be useful. See https://github.com/InformaticsMatters/pipelines-utils-rdkit/blob/master/src/python/pipelines_utils_rdkit/filter.py

Same here. I think this functionality is available in other tools already. We have a ton of filtering and transformation tools. In the end its also a problem if Findability. People searching for filtering are not looking at a splitting tool and vice versa. So I would keep functionality as much as possible separated. But yes, I know it has advantages as seen with Vina to merge tools - I guess we need always to discuss this on a case-to-case basis.

simonbray:
Somehow didn't notice this PR, sorry about that.
I ran the tools and made some changes in this PR: https://github.com/tdudgeon/galaxytools/pull/1
Mostly some cosmetic things. I padded out the documentation a bit for the rdock tools and also added the receptor.prm using the `<configfile>` tag. See this commit: https://github.com/tdudgeon/galaxytools/pull/1/commits/fcb7be3f9ace53bc3179c8d89acefaf269c2b83a
Also moved the sucos dir up one level. See this commit: https://github.com/tdudgeon/galaxytools/pull/1/commits/ee1f748b9d1a4cc39de1f301cdc8a96289baa265
Looks good to me!

chemicaltoolbox/rdkit/sucos.py
simonbray: I think it should be one level up rather than in the rdkit folder though

tdudgeon: It's based heavily on RDKit.

chemicaltoolbox/autodock_vina/docking/docking.xml
tdudgeon: I modified this to use gnu parallel, but there are some bits I'm not sure about.

1. There are two bits that control the parallelism a). the `-j` option to parallel that tells parallel how may processes to run; b), the `--cpu` option to vina that tells vina how many cores to use. I'm not sure how these should be specified. There's also the `GALAXY_SLOTS` environment variable that seem to want to be used, but seems to always have a value of 1 in my hands, so I'm not sure how this should be used. 
2. The quotation of the execution string is getting pretty scary. Parallel seems to require its command to be single quoted, but galaxy also needs some arguments to be single quoted (double quotes doesn't seem to work for either). To simplify this as much as possible I'm:
2.1. only quoting params that can sensibly contain spaces
2.2 using the `$'abc\'xyz'` pattern of escaping
3. The way params are specified using a box file and directly is a bit inconsistent. When being specified some optional params have hard-coded default values which I feel is not right. Even if these happen to be the same as vina would use as defaults if the param was not specified I don't think they should be provided in the command e.g. params that have default values provided by vina should be optional in galaxy.
4. Because of the inconsistency of how params are handles the 3rd test is a bit messy. It should (assuming the params issue did not exist) generate the same output as the second test, but it does not so I'm currently providing a separate output file to compare against.

@bgruening @simonbray Your thoughts on these welcome.

tdudgeon:
I merged the PR from @simonbray which includes the changes to rdock and sucos.

chemicaltoolbox/rdkit/splitter.py
tdudgeon: @bgruening The splitting does not seem to be correct. It does not split at the end of a record. Look as the start of the 3rd part - it starts in the middle of the data section.
Also, splitting is done on the basis of how many files your want, not the number of records you want in each file.

chemicaltoolbox/autodock_vina/docking/docking.xml
bgruening: > There are two bits that control the parallelism a). the -j option to parallel that tells parallel how may processes to run; b), the --cpu option to vina that tells vina how many cores to use. I'm not sure how these should be specified. There's also the GALAXY_SLOTS environment variable that seem to want to be used, but seems to always have a value of 1 in my hands, so I'm not sure how this should be used.

Galaxy slots is by default 1. An admin can set the amount of CPU depending on given infrastructure, per tool or depending on input size etc ... so this can be configured by an Admin. Otherwise, it is 1.

So I guess --cpu should be 1 and  -j should be GALAXY_SLOTS?
We could also make this dependent on the amount of molecules. As you prefer Tim. The only thing that we need to take care is that GALAXY_SLOTS is the max amount of CPUs that can be used by this job.

> The quotation of the execution string is getting pretty scary. Parallel seems to require its command to be single quoted, but galaxy also needs some arguments to be single quoted (double quotes doesn't seem to work for either).

Can you not put those commands in a txt file as I did in freebayes?

> he way params are specified using a box file and directly is a bit inconsistent. When being specified some optional params have hard-coded default values which I feel is not right. Even if these happen to be the same as vina would use as defaults if the param was not specified I don't think they should be provided in the command e.g. params that have default values provided by vina should be optional in galaxy.

Then let's make all optional and only pass them to the CMD if given by the user.

bgruening:
@tdudgeon can you remove the splitting tool from this PR. We will try to repair the other tool or add your splitting tool in a new PR. ok?

chemicaltoolbox/autodock_vina/docking/docking.xml
bgruening: let's use `--cpu 1` and `-j ` for GALAXY_SLOTS ...

tdudgeon:
@bgruening 
Splitter tool has been removed.
Parallelism updated for vina tool (also given it a better name).

chemicaltoolbox/rdock/rbcavity.xml
```diff
@@ -0,0 +1,90 @@
...
+        <param name="sphere" type="float" value="1.0" label="Mapper small sphere radius" help="Mapper radius of small sphere in angstroms; rDock will search for cavities accessible to the small sphere but not to the large sphere"/>
+        <param name="min_volume" type="integer" value="100" label="Mapper minimum volume" help="Mapper minimum volume in cubic angstroms. rDock will ignore cavities which are smaller than this value"/>
+        <param name="vol_incr" type="float" value="0.0" label="Mapper volume increment" help="Mapper volume increment in angstroms. When excluding the volume occupied by the receptor from the cavity search, rDock will temporarily increase the radius of receptor atoms by this amount"/>
+        <param name="gridstep" type="float" value="0.5" label="Mapper grid step" help="Grid resolution for mapping in angsroms"/>
+        <param name="weight" type="float" value="1.0" label="Cavity weight" help="Cavity weight"/>
+    </inputs>
+    <outputs>
+        <data name="activesite" format="binary" label="rbcavity active site on ${on_string}"/>
```
bgruening: let's create a proper filetype for this: https://github.com/galaxyproject/galaxy/pull/8727

chemicaltoolbox/rdock/rbcavity.xml
```diff
@@ -0,0 +1,90 @@
...
+    SCORING_FUNCTION RbtCavityGridSF
+    WEIGHT $weight
+END_SECTION
+        </configfile>
+    </configfiles>
+
+    <inputs>
+        <param type="data" name="receptor" format="mol2" label="Receptor" help="Select a receptor (Mol2 format)."/>
+        <param type="data" name="ligand" format="mol,sdf" label="Reference ligand" help="Single ligand in Mol or SDF format"/>
+        <param name="radius" type="float" value="6.0" label="Mapper sphere radius" help="Mapper radius of large sphere in angstroms; rDock will search for cavities to the small sphere but not to the large sphere"/>
```
bgruening: we can add min here to all params isn't it?

chemicaltoolbox/rdock/rbcavity.xml
```diff
@@ -0,0 +1,90 @@
+<tool id="rbcavity" name="rDock cavity definition" version="0.1">
+    <description>- generate the active site definition needed for rDock docking</description>
+    <requirements>
```
bgruening: Would a macro make sense here? For the requirement, the citation and the version number?

chemicaltoolbox/rdock/rbdock.xml
```diff
@@ -0,0 +1,286 @@
...
+ln -s '$receptor' receptor.mol2 &&
+ln -s $receptor_prm receptor.prm &&
+#if $name == 'Y':
+  sdmodify -f_REC '$ligands' > ligands.sdf &&
+#else
+  ln -s '$ligands' ligands.sdf &&
+#end if
+rbdock -i ligands.sdf -r receptor.prm -p dock.prm -n $num -o output &&
+sdsort -n -s -fSCORE output.sd |
+#if len(str($score)) > 0:
```
bgruening: `#if $score and $score > 0` should work.

chemicaltoolbox/rdock/rbdock.xml
```diff
@@ -0,0 +1,286 @@
...
+  sdfilter -f'\$SCORE <= $score' |
+#end if
+#if len(str($nscore)) > 0:
```
bgruening: same here

chemicaltoolbox/rdock/rbdock.xml
```diff
@@ -0,0 +1,286 @@
+<tool id="rbdock" name="rDock docking" version="0.1">
```
bgruening: Is this tool multi-core enabled? Should GALAXY_SLOTS be used?
If not do we want to parallelize it with gnu-parallel as well?

chemicaltoolbox/sucos/sucos.xml
```diff
@@ -0,0 +1,60 @@
...
+                </assert_contents>
+            </output>
+        </test>
+    </tests>
+    <help><![CDATA[
+.. class:: infomark
+
+**What it does**
+
+This tool determines the shape and feature overlap of docked ligand poses compared to a reference molecule, usually a known ligand.
```
bgruening: I like your help very much! Thanks for that!

chemicaltoolbox/sucos/sucos.py
```diff
@@ -0,0 +1,202 @@
+#!/usr/bin/env python
+"""
+Basic SuCOS scoring. Allows a set of molecules from a SD file to be overlayed to a reference molecule,
+with the resulting scores being written as properties in the output SD file.
+
+SuCOS is the work of Susan Leung.
+GitHub: https://github.com/susanhleung/SuCOS
```
bgruening: Do you want to add that this is MIT licensed?
We could also create a conda package out of that ....

chemicaltoolbox/autodock_vina/docking/docking.xml
```diff
@@ -77,57 +72,53 @@
...
+                <!-- we check only the first 2 -->
+            </output_collection>
+        </test>
+
+        <test>
+            <param name="receptor" value="protein.pdbqt"/>
+            <param name="ligands" value="input_ligand.pdbqt"/>
+            <param name="box" value="box.txt"/>
+            <output_collection name="sdf_outputs" type="list" count="1">
+                <element name="ligand1_docked" file="ligand1_docked.sdf" lines_diff="40"/>
```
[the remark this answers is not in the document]
tdudgeon: There are date stamps in the outputs which means there are differenerces. I reduced the lines_diff to 20. https://github.com/tdudgeon/galaxytools/commit/f00d953c332d85b35b246fedd3afb98eceafb079

chemicaltoolbox/rdock/rbcavity.xml
tdudgeon: OK. looks good.

chemicaltoolbox/rdock/rbcavity.xml
tdudgeon: @bgruening Not sure what you're meaning here.

chemicaltoolbox/rdock/rbcavity.xml
tdudgeon: Use macro: https://github.com/tdudgeon/galaxytools/commit/3fe85518879a9a987112f46eb10d2a028575d973

chemicaltoolbox/rdock/rbdock.xml
tdudgeon: Changed syntax: https://github.com/tdudgeon/galaxytools/commit/3fe85518879a9a987112f46eb10d2a028575d973

chemicaltoolbox/rdock/rbdock.xml
tdudgeon: See https://github.com/tdudgeon/galaxytools/commit/3fe85518879a9a987112f46eb10d2a028575d973

chemicaltoolbox/rdock/rbdock.xml
tdudgeon: @bgruening rdock only utilises a single core. My expectation was that the input would be split into multiple files (a collection) on the galaxy side and that collection used as input to the tool.
In order to use gnu parallel I think we would need to split the input as part of the execution?

chemicaltoolbox/sucos/sucos.py
tdudgeon: Mentioned MIT license: https://github.com/tdudgeon/galaxytools/commit/10f9f7c36998af3b4104e02ddda1a1bb1ccb13eb

tdudgeon:
I did this commit (https://github.com/tdudgeon/galaxytools/commit/c7a9e50f0487b713ae1af858b4056af287db475c) to clean up the default params for VINA. In some cases where VINA has a default we we hard-coding a non-default value that was very different to VINA's default. 
Instead options that VINA will provide defaults for are optional in Galaxy.

chemicaltoolbox/rdock/rbcavity.xml
simonbray: A minimum value, probably `min="0"`.

chemicaltoolbox/rdock/rbdock.xml
simonbray: @tdudgeon you don't anticipate that this submitting a separate job for each ligand will cause similar problems to those you had with Vina?

tdudgeon: No, because each rdock execution handles multiple ligands (input is SDF), and you can tune the number to give a sensible balance between execution time and parallelism.

chemicaltoolbox/rdock/rbcavity.xml
tdudgeon: Added: https://github.com/tdudgeon/galaxytools/commit/23d7d698390b58f2bf9487a7ee8e011511ecb74c

nsoranzo:
Hi guys, sorry for not checking this before, but are you aware of https://github.com/TGAC/earlham-galaxytools/tree/master/tools/rdock which IIRC is already part of the ChemicalToolbox ?

bgruening:
@nsoranzo arg! :( 
No it's not at least it's not installed afaik. We have forgotten to check the TS I guess. 
Are you maintaining your tool? How do we proceed now?

nsoranzo:
@bgruening The person that developed the wrapper (and also `smina` and `apoc`) has left our institute some time ago, so I'd be happy to transfer any of them, in order not to break lineages and make updates easier.

> No it's not at least it's not installed afaik.

https://github.com/bgruening/docker-galaxy-chemicaltoolbox/blob/master/chemicaltoolbox_tools.yml#L70-L76

bgruening:
@nsoranzo I was referring to https://cheminformatics.usegalaxy.eu

I guess then we would have catched this earlier. I must have forgotten about rdock, sorry about that.
Can you please update the .shed.yml in this PR to fit your version and give access to `bgruening` to your rdock repository. I think then we should be all set?

nsoranzo:
Given that rDock here is split in 2 tools, on second thought I believe it is not worth to try to keep lineage, so feel free to merge from my side.

bgruening:
Ok, thanks @nsoranzo!

PR #906, opened 2019-09-24
fix generalized_linear tool fail to load header option

qiagu:
Also refactored the input options.


qiagu:
I feel this PR can be patched to v1.0.7.12 and the fix commit goes to even older versions.

bgruening:
thanks!

qiagu:
@bgruening  Thanks for merging!

PR #914, opened 2019-10-08
Add sdf format and chunk mode to splitting tool

simonbray:


tools/text_processing/split_file_to_collection/split_file_to_collection.xml
```diff
@@ -16,6 +16,16 @@
         </xml>
         <xml name="numnew_fname">
             <param name="numnew" type="integer" label="Number of new files" min="1" value="1"/>
```
simonbray: in principle it would make more sense for `numnew` to also go into the conditional with `chunking` (so they are alternatives) but would it break continuity with the old version too much to hide an option behind a conditional?

bgruening: As long as we bump the tool version I think this is fine.

bgruening: Do you want to do this still @simonbray?

simonbray: yes, just fixing this and one other thing

bgruening:
split_after.fasta is missing...

bgruening:
Cool danke!

PR #916, opened 2019-10-10
PCA tool

simonbray:
@khanteymoori, I wrote a very basic PCA tool. Maybe we can work together to add some more features?


simonbray:
@anuprulez, any specific ideas for a visualization?

PR #918, opened 2019-10-11
aggressive detect_errors was too aggressive

tdudgeon:
I changed the detect_errors property to `exit_code` as when RDKit encounters a bad molecule it writes out ERROR to stderr, but this does not stop the tool from working (that molecule is skipped) and the tool completes successfully with exit code of 0.

Not sure it there is a mechanism to give user feedback that some records were not processed?


bgruening:
The user should still see the stderr if they wish. It should be logged. Thanks for the patch!

PR #919, opened 2019-10-11
Modifying alevin output to more useable file types

astrovsky01:


astrovsky01:
Will be modifying more to make the dumpUmiGraph flag output actual graphs

tools/salmon/macros.xml
```diff
@@ -8,6 +8,10 @@
             <requirement type="package" version="1.2">seqtk</requirement>
             <requirement type="package" version="0.14.1">salmon</requirement>
             <requirement type="package" version="1.9">samtools</requirement>
+            <requirement type="package" version="0.2.0">vpolo</requirement>
+            <requirement type="package" version="1.17.2">numpy</requirement>
```
bgruening: where is all of this used?

astrovsky01: It's strange, but for some reason, even though vpolo's bioconda wrapper works properly, if I don't also import numpy, etc. in the requirements, the tool throws a fit

PR #920, opened 2019-10-15
Add sdf_to_tab tool

simonbray:
Extracts values for properties from an sdf file and saves them in tabular format.

ping @tdudgeon


chemicaltoolbox/rdkit/sdf_to_tab.py
```diff
@@ -0,0 +1,37 @@
...
+import argparse
+import pandas as pd
+from rdkit import Chem
+
+def sdf_to_tab(vars):
+    mols = Chem.SDMolSupplier(vars.inp, sanitize=False)
+    df = pd.DataFrame()  # for output
+
+    for mol in mols:
+        if mol != None:
```
bgruening:
```suggestion
        if mol:
```

chemicaltoolbox/rdkit/sdf_to_tab.py
```diff
@@ -0,0 +1,37 @@
...
+            d = mol.GetPropsAsDict()
+            # filter dict for desired props
+            if vars.props == '':  # none specified, return all
```
bgruening:
```suggestion
            if vars.props.strip() == '':  # none specified, return all
```

chemicaltoolbox/rdkit/sdf_to_tab.py
```diff
@@ -0,0 +1,37 @@
...
+                d = {prop: val for (prop, val) in d.items() if not any(x in str(val) for x in ['\n', '\t'])}  # remove items containing newlines or tabs
+            else:
+                d = {prop: val for (prop, val) in d.items() if prop in vars.props.replace(' ', '').split(',')}  # remove items not requested via CLI
+            d['Name'] = mol.GetProp('_Name')
+        else:
+            d = {'Name': 'ERROR'}
```
bgruening: Is this needed? You can not even track this down to the molecule in question. Should we error out in this case?

chemicaltoolbox/rdkit/sdf_to_tab.xml
```diff
@@ -0,0 +1,104 @@
+<tool id="sdf_to_tab" name="Extract values from an SD-file" version="@VERSION@">
+    <macros>
+        <token name="@VERSION@">2019.03.1</token>
+    </macros>
+    <description>into a tabular file using RDKit</description>
+    <requirements>
+        <requirement type="package" version="@VERSION@">rdkit</requirement>
```
bgruening: pandas?

chemicaltoolbox/rdkit/sdf_to_tab.xml
```diff
@@ -0,0 +1,104 @@
...
+        python '$__tool_directory__/sdf_to_tab.py'
+            -i '${infile}'
+            -o '${outfile}'
+            -p '$props'
+            $header
+]]>
+    </command>
+    <inputs>
+        <param name="infile" format="sdf" type="data" label="Input SD-file"/>
+        <param name="header" type="boolean" label="Include the property name as header"
```
bgruening: Mh, interesting is there a use-case where you do not want the header?

chemicaltoolbox/rdkit/sdf_to_tab.xml
```diff
@@ -0,0 +1,104 @@
...
+            truevalue="--header" falsevalue="" checked="false" />
+        <param name="props" type="text" label="Properties to extract from the SD-file" help="Separated with a comma, e.g. RMSD_UB,RMSD_LB,SCORE. To extract values for all properties in the file, leave the field blank."/>
```
bgruening: So this should be optional=True rihgt?

chemicaltoolbox/rdkit/sdf_to_tab.xml
simonbray: dependency of rdkit

chemicaltoolbox/rdkit/sdf_to_tab.xml
simonbray: Not afaik but it's more consistent with some other tools which offer the same option.

chemicaltoolbox/rdkit/sdf_to_tab.py
simonbray: At least this way there's a record that a molecule was skipped, this is a good thing right?

chemicaltoolbox/rdkit/sdf_to_tab.xml
simonbray: Originally I wanted to use `multiple="true"` but this doesn't work for text params. I don't know if there is some more officially favoured solution?

chemicaltoolbox/rdkit/sdf_to_tab.py
bgruening: but in the result file? Maybe on the stdout .... 3 molecules skipped or something like this.

chemicaltoolbox/rdkit/sdf_to_tab.py
```diff
@@ -0,0 +1,37 @@
...
+
+    for mol in mols:
+        if mol:
+            d = mol.GetPropsAsDict()
+            # filter dict for desired props
+            if vars.props.strip() == '':  # none specified, return all
+                d = {prop: val for (prop, val) in d.items() if not any(x in str(val) for x in ['\n', '\t'])}  # remove items containing newlines or tabs
+            else:
+                d = {prop: val for (prop, val) in d.items() if prop in vars.props.replace(' ', '').split(',')}  # remove items not requested via CLI
+            d['Name'] = mol.GetProp('_Name')
```
tdudgeon: This looks dangerous - many SDFs don't have a name field (first line). In those cases I think this will throw and error as the key is not in the dict? Or if is is then it will be None or empty string. I would only include the Name field in the output if it is really defined in the first line of the SDF record.
Also, I think it would be useful to always include the record index as an output field?

tdudgeon:
One nice option would be to optionally allow to write out the mol as SMILES as an extra column.
As this is already using RDKit this would be pretty simple.

bgruening:
@tdudgeon if you like you can PR to this branch. I think this is a good idea.

chemicaltoolbox/rdkit/sdf_to_tab.py
simonbray: It doesn't throw an error, it just returns an empty string. So I think this is safe enough.

Adding columns for record index and SMILES is a good idea, I'll add those.

PR #900, opened 2019-09-10
Fix to dendrogram generation for nxn clustering

simonbray:

PR #922, opened 2019-10-17
update box tool; bump nxn version

simonbray:
Update the calculate box parameter tool to allow other file types: SDF, MOL2 and PDB as well as MOL.

Also bumped the version number for nxn-clustering because this was missed in a previous PR: https://github.com/bgruening/galaxytools/pull/900

### DIFF
--- a/chemicaltoolbox/autodock_vina/prepare_box/prepare_box.xml
+++ b/chemicaltoolbox/autodock_vina/prepare_box/prepare_box.xml
@@ -1,5 +1,5 @@
 <tool id="prepare_box" name="Calculate the box parameters for an AutoDock Vina" version="0.1.1">
-    <description>job from an input mol file (confounding box)</description>
+    <description>job from a ligand input file (confounding box)</description>
     <requirements>
         <requirement type="package" version="2019.03.1">rdkit</requirement>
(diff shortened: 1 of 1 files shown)


chemicaltoolbox/autodock_vina/prepare_box/prepare_box.xml
```diff
@@ -1,12 +1,13 @@
-<tool id="prepare_box" name="Calculate the box parameters for an AutoDock Vina" version="0.1.0">
-    <description>job from an input mol file (confounding box)</description>
+<tool id="prepare_box" name="Calculate the box parameters for an AutoDock Vina" version="0.1.1">
+    <description>job from a ligand input file (confounding box)</description>
     <requirements>
-        <requirement type="package" version="2019.03.1.0">rdkit</requirement>
+        <requirement type="package" version="2019.03.1">rdkit</requirement>
```
bgruening: 2019.03.4 is already out ... but we do not need to update now. Your call.

simonbray: I made a start but it seems this could be a rabbit hole... I will make a PR at some point with all tools and all packages updated (e.g. OpenBabel 3.0 is also out).

PR #928, opened 2019-10-21
Update ribotaper_part2_create_metaplots.xml

zslastman:
ghostscript is required for the collation of those pdfs - it's not available in the root in newer galaxy versions


bgruening:
@zslastman ghostscript was never part of Galaxy. __All__ dependencies needs to be given as part of the requirements.

PR #929, opened 2019-10-30
bump 3.1.2 + simplification

martin-raden:
- intarna 3.1.2
- replace sed-based tab-separated output generation with new `--outSep` argument setup
- refactor call generation to reduce redundancy

### DIFF
--- a/tools/rna_tools/intarna/intarna.xml
+++ b/tools/rna_tools/intarna/intarna.xml
@@ -1,4 +1,4 @@
-<tool id="intarna" name="IntaRNA" version="3.1.1">
+<tool id="intarna" name="IntaRNA" version="3.1.2">
     <description>Efficient RNA-RNA interaction prediction incorporating accessibility and seeding of interaction sites.</description>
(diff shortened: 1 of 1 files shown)


martin-raden:
intarna package build on bioconda currently running...

martin-raden:
if I edit via the github web interface, a tab will be replaced by spaces.. :/

bgruening:
You can configure this in the github interface ...

> id1\tstart1\tend1\tid2\tstart2\tend2\tsubseqDP\thybridDP\tE

what do you mean?

tools/rna_tools/intarna/intarna.xml
```diff
@@ -1,4 +1,4 @@
-<tool id="intarna" name="IntaRNA" version="3.1.1">
+<tool id="intarna" name="IntaRNA" version="3.1.2">
```
bgruening: if you want to save your work in the long run, this version could also be macrofied ... so you only need to bump it once

PR #933, opened 2019-11-05
add quantiletransformer, powertransformer and kbindiscretizer to pipeline

qiagu:
Sorry one commit was missed for paracords tool. We don't need to update the tool immediately though.


qiagu:
@bgruening  The tests are badly imbalanced.
image

bgruening:
Currently, every folder is tested separately ...

PR #936, opened 2019-11-06
Clean model params before fitting

qiagu:
For security reason, remove support of estimators that contain file write handler.

Maybe need a warning?


bgruening:
security first :)

PR #937, opened 2019-11-20
Update easy_join "Join two files"

scholtalbers:
Update to accept only tabular on the second file as well. With just accepting txt, there is no dropdown to select a column. Which is inconsistent as the first file does have this dropdown.


bernt-matthias:
Just stumbled over the same problem. 

@bgruening could you merge this, maybe together with https://github.com/bgruening/galaxytools/pull/834

PR #939, opened 2019-12-03
Diamond 0.9.29

abretaud:
A new version for Diamond (I have not added new options as I don't think they're very important)
Also updated the data manager + added a test for it


data_managers/data_manager_diamond_database_builder/data_manager/data_manager_diamond_database_builder.xml
```diff
@@ -1,10 +1,10 @@
-<tool id="diamond_database_builder" name="Diamond" tool_type="manage_data" version="0.0.1">
+<tool id="diamond_database_builder" name="Diamond" tool_type="manage_data" version="0.0.2">
     <description> Database builder</description>
     <requirements>
-        <requirement type="package" version="0.6.13">diamond</requirement>
+        <requirement type="package" version="0.9.29">diamond</requirement>
```
mvdbeek: Can you also add python (I guess 2.7 if you don't want to fix data_manager_diamond_database_builder.py) as a dependency ? And in the script it'd be good to replace https://github.com/bgruening/galaxytools/blob/master/data_managers/data_manager_diamond_database_builder/data_manager/data_manager_diamond_database_builder.py#L15 with json.dumps / json.loads

data_managers/data_manager_diamond_database_builder/data_manager/data_manager_diamond_database_builder.py
```diff
@@ -12,227 +13,228 @@
...
-    #Fetch the FASTA
-    REFERENCE_SOURCE_TO_DOWNLOAD[ params['param_dict']['reference_source']['reference_source_selector'] ]( data_manager_dict, params, target_directory, database_id, database_name )
-    
-    #save info to json file
-    open( filename, 'wb' ).write( to_json_string( data_manager_dict ) )
+    # Fetch the FASTA
+    REFERENCE_SOURCE_TO_DOWNLOAD[params['param_dict']['reference_source']['reference_source_selector']](data_manager_dict, params, target_directory, database_id, database_name)
+
+    # save info to json file
+    open(filename, 'wb').write(json.dumps(data_manager_dict))
```
mvdbeek:
```suggestion
    open(filename, 'w').write(json.dumps(data_manager_dict, sort_keys=True))
```
that'll break the test output but should be more consistent between test runs.

abretaud:
hum, probably some network problem between travis and ncbi

mvdbeek:
Hmm, that's a multi-gigabyte download, we shouldn't lint this ... can we skip the link ?

bgruening:
Thanks!

PR #940, opened 2019-12-13
Graphprot update

michauhl:
Hi,
behaviour is still weird. In Galaxy (planemo serve) the tool produces outputs that are actually disabled (by output filters). E.g. in training mode (action_type="train") when disable_motifs or disable_cv is set to True, it still produces the output files (seq_motif_out_file, cv_results_out_file ..). Same is true for action_type="predict", except for genomic_peaks_out_file, where there is no output produced at all! (thus the failed test 2)
The produced command line calls in Galaxy look fine and work in my local console, producing the expected outputs stored in test-data/ (Python 3.7.3).
So maybe the output filters do not work, but I cannot see an error here (compared to other wrapper .xml files).
Any ideas?
Thanks,
Michael


tools/rna_tools/graphprot/graphprot_train_predict.xml
```diff
@@ -0,0 +1,285 @@
...
+    <outputs>
+
+        <data name="model_out_file" format="txt" from_work_dir="GraphProt.model" label="${tool.name} on ${on_string}: GraphProt model file">
+            <filter>action_type["action_type_selector"] == "train"</filter>
+        </data>
+        <data name="params_out_file" format="txt" from_work_dir="GraphProt.params" label="${tool.name} on ${on_string}: GraphProt model parameters file">
+            <filter>action_type["action_type_selector"] == "train"</filter>
+        </data>
+        <data name="cv_results_out_file" format="txt" from_work_dir="GraphProt.cv_results" label="${tool.name} on ${on_string}: GraphProt cross validation results file">
+            <filter>action_type["action_type_selector"] == "train" and disable_cv is False</filter>
```
wm75: try:

```suggestion
            <filter>action_type["action_type_selector"] == "train" and not action_type["training_options"]["disable_cv"]</filter>
```

wm75: Yes, it *is* a problem with the output filters. Sections represent a level of nesting just like conditionals (and repeats), which get expressed as nested dicts in the filters. Adjust the other filters accordingly and things should work.

The `boolean_param is False` vs `not boolean_param` is just cosmetic (the second form is more pythonic).

bgruening: Oh, nice ... it took me too long to find this as well. I'm still not sure of all are working. 

@michauhl can you please add https://github.com/galaxyproject/tools-iuc/blob/91b28adf44f425970469cc52fc5c13e75409ed27/tool_collections/samtools/samtools_markdup/samtools_markdup.xml#L61  to the test to check if the expected output number is correct?

tools/rna_tools/graphprot/gplib.py
```diff
@@ -102,7 +102,7 @@ def graphprot_profile_get_top_scores_median(profile_file,
 
 
 ################################################################################
-
+#!/usr/bin/env python3
```
wm75: Did you mean to remove the existing shebang instead of duplicating it?
The former would make sense for a library with nothing to execute.

michauhl: It's the old one. You want me to remove it?

michauhl: Since there's no main to execute, I could remove it true :)

wm75: I would remove both this one and the one on top. It's a lib, not a script.

michauhl: Oh, were did this come from? Consider it removed ..

michauhl: I think everything works fine now

michauhl: Thank u very much Wolfgang + Björn

michauhl: Anything else?

michauhl: Tool required memory I would set to 16 GB

PR #943, opened 2020-01-08
update sklearn tools to v1.0.8.2 and add image deep learning tool

qiagu:
Main changes:
- add image deep learning tool
- API improvements for deep learning models: introduced a new parameter `validation_fraction` which would be useful in DNN models, let `verbose` expose to users, etc.
- some bug fixes

PR #944, opened 2020-01-15
Add LightGBM tool

simonbray:
This is a tool @khanteymoori and I need for our cheminformatics project.


tools/sklearn/lightgbm.xml
```diff
@@ -0,0 +1,293 @@
...
+    </command>
+    <configfiles>
+        <inputs name="inputs"/>
+        <configfile name="lightgbm_script">
+<![CDATA[
+import json
+import lightgbm as lgb
+import numpy as np
+import pandas
+import pickle
```
qiagu: `pickle` is not used.

simonbray: Good point, thanks!

simonbray:
@bgruening, can you have a look at this?

There is one test failing, but it's not related to this tool. Not sure why - it completes successfully locally as well as in @qiagu's PR.

tools/sklearn/lightgbm.xml
```diff
@@ -0,0 +1,293 @@
...
+import sys
+
+from scipy.io import mmread
+import sklearn.ensemble
```
qiagu: And `ensemble `

qiagu:
This is great. 
I had an idea to add `lightgbm`, maybe `catboost` as well, to the pipeline tool long time ago. Here we have two options: 1) directly add to the pipeline tool, like we did for `xgboost`; 2) extend this tool to output unfitted pickle object and load it into the pipeline tool via `custom model`.

simonbray:
Hm, I think the disadvantage of 1) is that the tool will need an extra requirement (`lightgbm`) which will not be relevant for most users.

tools/sklearn/lightgbm.xml
```diff
@@ -0,0 +1,291 @@
...
+if params["selected_tasks"]["selected_task"] == "train":
+    algorithm = params["selected_tasks"]["selected_algorithms"]["selected_algorithm"]
+    options = params["selected_tasks"]["selected_algorithms"]["options"]
+    if "max_leaf_nodes" in options:
+        options["num_leaves"] = options["max_leaf_nodes"]
+        options.pop("max_leaf_nodes")
+
+    X, y = get_X_y(params, infile1, infile2)
+    lgb_train = lgb.Dataset(X, y)
+    gbm = lgb.train(options, lgb_train)
```
qiagu: It seems these estimators don't comply with sklearn `fit`, `predict` APIs. Wrapper is needed to make them work in pipeline.

qiagu:
https://github.com/microsoft/LightGBM/blob/master/python-package/lightgbm/sklearn.py
I'm glad they support sklearn APIs.
Pipeline has advantages over single estimator if preprocessing steps are needed, which happens most time in my practices. It's well discussed that applying precessing on whole dataset before train test split can lead to incorrect evaluation.

qiagu:
https://github.com/bgruening/galaxytools/pull/943/files#diff-cee8a67d2b46be46df04d32f0a71234cL1-L5
The same error can be solved by updating the test data.
@simonbray You're right. Adding to pipeline would lead to the dependencies change. We can leave that later.

simonbray:
@qiagu, what do you think needs to be done before this can be merged? Should I make an option to output a pickle object, or should this be done later?

tools/sklearn/lightgbm.xml
```diff
@@ -0,0 +1,291 @@
...
+import lightgbm as lgb
+import numpy as np
+import pandas
+import sys
+
+from scipy.io import mmread
+from galaxy_ml.utils import load_model, get_X_y
+
+
+N_JOBS = int(__import__('os').environ.get('GALAXY_SLOTS', 1))
```
anuprulez: Do you use this variable (`N_JOBS`) anywhere else in the tool definition?

tools/sklearn/lightgbm.xml
```diff
@@ -0,0 +1,291 @@
...
+    </tests>
+    <help><![CDATA[
+**What it does**
+
+LightGBM is a gradient boosting framework that uses tree based learning algorithms. It 
+is designed to be distributed and efficient with the following advantages:
+
+* Faster training speed and higher efficiency.
+* Lower memory usage.
+* Better accuracy.
```
anuprulez: `Faster, better ...` than which algorithm? If you know, please mention it

tools/sklearn/lightgbm.xml
```diff
@@ -0,0 +1,291 @@
...
+        <inputs name="inputs"/>
+        <configfile name="lightgbm_script">
+<![CDATA[
+import json
+import lightgbm as lgb
+import numpy as np
+import pandas
+import sys
+
+from scipy.io import mmread
```
anuprulez: do you use `mmread` imported method anywhere else?

tools/sklearn/lightgbm.xml
```diff
@@ -0,0 +1,291 @@
...
+from galaxy_ml.utils import load_model, get_X_y
```
anuprulez: same with `load_model`?

tools/sklearn/lightgbm.xml
```diff
@@ -0,0 +1,291 @@
...
+    python '$lightgbm_script' '$inputs'
+]]>
+    </command>
+    <configfiles>
+        <inputs name="inputs"/>
+        <configfile name="lightgbm_script">
+<![CDATA[
+import json
+import lightgbm as lgb
+import numpy as np
```
anuprulez: I do not see `np` being used anywhere? Please check

tools/sklearn/lightgbm.xml
simonbray: You are right, this isn't very objective. I have removed this section. Thanks for the comments!

qiagu:
This is fine for this tool and PR. We can add lightGBM into galaxy-ml utils in next version. Then add the classifier and regressor in pipeline tool and be able to tune hyperparameters

PR #945, opened 2020-01-16
Update plotly_ml tool

simonbray:
Updating tool versions + test-data


tools/plotly_ml_performance_plots/plot_ml_performance.py
```diff
@@ -140,7 +140,7 @@ def main(infile_input, infile_output, infile_trained_model):
         plotly.offline.plot(fig_roc, filename="output_roc.html", auto_open=False)
 
     except Exception as exp:
-        pass
+        print("Plotting the ROC-AUC graph failed. This exception was raised: {}".format(exp))
```
simonbray: I discovered the reason that the reason the ROC-AUC curves were not being created: an error was being silently caught here. I added this line to make debugging easier in the future.

bgruening:
But this means we need to bump tool versions as well.

simonbray:
Whoops - done now

PR #950, opened 2020-01-31
bump to v3.1.5 (bugfix release)

martin-raden:
[bioconda upload](https://github.com/bioconda/bioconda-recipes/runs/420126221) currently running ...


martin-raden:
so check might need a restart after the conda package is available... mmhh.. was a bit early for the PR.. ;)

PR #952, opened 2020-02-28
remove file with header only in the output collection

sunyi000:
the output of the original tool contains a file with header only. This PR remove this file from the output


tools/text_processing/split_file_on_column/split_file_on_column.xml
```diff
@@ -6,7 +6,8 @@
     <command>
 <![CDATA[
         mkdir tmp_out &&
-        awk -F'\t' '{print > "tmp_out/"\$$column".$infile.ext" }' '$infile'
+
+        awk -F '\t' 'NR==1{hdr=$0;next}f!="tmp_out/"\$$column".$infile.ext"{if(f) close(f); f="tmp_out/"\$$column".$infile.ext";print hdr>f} {print >> f}' '$infile'
```
bgruening: can you add this as an option?

I could imaging a boolean asking the user to prevent the header line (or even the number of header lines)

bgruening:
please also add a test-case with headers, thanks

sunyi000:
@bgruening 
I have updated GALAXY_RELEASE variable in .travis.yml 
could you please review again and let me know if there's anything else need to be changed

tools/text_processing/split_file_on_column/split_file_on_column.xml
```diff
@@ -1,17 +1,41 @@
-<tool id="tp_split_on_column" name="Split file" version="0.3">
+<tool id="tp_split_on_column" name="Split file" version="0.2">
```
bgruening: you need to increase the version number :)

tools/text_processing/split_file_on_column/split_file_on_column.xml
```diff
@@ -1,17 +1,41 @@
...
     <description>according to the values of a column</description>
     <requirements>
-        <requirement type="package" version="4.1.3">gawk</requirement>
+        <requirement type="package" version="4.1.0">gnu_awk</requirement>
```
bgruening: please use the conda-forge version .... https://anaconda.org/conda-forge/gawk

tools/text_processing/split_file_on_column/split_file_on_column.xml
```diff
@@ -1,17 +1,41 @@
...
+        <conditional name="condition_has_header">
+            <param name="has_header" type="select" display="radio" label="Does the input file has header?">
+                <option value="n">No</option>
+                <option value="y">Yes</option>
+            </param>
+            <when value="y">
+                <param name="include_header" type="select" display="radio" label="Include header in all splitted files?" value="n">
+                  <option value="n">No</option>
+                  <option value="y">Yes</option>
+	        </param>
```
bgruening: indentation is wrong here, please use 4 spaces

tools/text_processing/split_file_on_column/split_file_on_column.xml
```diff
@@ -1,17 +1,41 @@
...
     <inputs>
         <param format="tabular" name="infile" type="data" label="File to select" />
         <param name="column" label="on column" type="data_column" data_ref="infile" accept_default="true" />
+        <conditional name="condition_has_header">
+            <param name="has_header" type="select" display="radio" label="Does the input file has header?">
+                <option value="n">No</option>
+                <option value="y">Yes</option>
+            </param>
+            <when value="y">
+                <param name="include_header" type="select" display="radio" label="Include header in all splitted files?" value="n">
```
bgruening: why is the conditional needed? This param is enough isn't it?

tools/text_processing/split_file_on_column/split_file_on_column.xml
sunyi000: done. increased it back to 0.3

tools/text_processing/split_file_on_column/split_file_on_column.xml
sunyi000: just the CDATA section?

tools/text_processing/split_file_on_column/split_file_on_column.xml
sunyi000: i'm just thinking users could upload a file with header but do not want to include header in the splitted files.
or upload a file with header and also include header in the split files.

if user upload a file without header, it functions as before...

bgruening:
I will fix this on the master branch.

bgruening:
@sunyi000 does the last commit make sense to you?

sunyi000:
> @sunyi000 does the last commit make sense to you?

make sense but the below scenario will produce incorrect results,
for example:
if source file contains a header, and user select do not include header, then the output is wrong

i still think the condition is better though...its more bullet proof imho @bgruening

bgruening:
>  if source file contains a header, and user select do not include header, then the output is wrong

But the header will not be on all splitted files, only in the first. If the user does not want it in the first file, there is a tool to remove th first line. Would that work?

sunyi000:
> > if source file contains a header, and user select do not include header, then the output is wrong
> 
> But the header will not be on all splitted files, only in the first. If the user does not want it in the first file, there is a tool to remove th first line. Would that work?

@bgruening 
please correct me if I'm wrong..here are the scenarios I can think of
 
1:if source file looks like below(with header) , split on the 3rd column

H1 H2 H3
a b 1
a c 1
b c 2

include_header=false, it will execute the orignal awk cmd which produces 3 files. 
file1:
a b 1
a c 1

file 2:
b c 2

file 3:
H1 H2 H3

so the file3 needs to be removed..... this is why we created this PR

with the same source file,
include_header=true, it executes the new awk and hence the correct output files.

2:
if source file is like this(without header)
a b 1
a c 1
b c 2

include_header=false, original awk will generate correct output

include_header=true (this combination does not make sense, but it could happen by user mistake)
In this case the output file will omit the first line, which is incorrect.

so I think the condition is really just to add some restrictions to the UI. 
we can also remove the condition but add another tool to remove the unwanted file.

Let me know if I understand this wrongly. Thanks

PR #957, opened 2020-03-09
Chemicaltoolbox, sygma 1.1.1+galaxy1

andzajan:
Dear @simonbray and @bgruening, we would like to make some changes in current implementation of the SyGMa tool.

Main aim of this refinement is to make outputs to be easier to apply for metabolomics studies.

Changes we have made so far:

* `--detailed` argument added to the python script, so that tool can return information about metabolic pathways and and chemical formula. [6c53cdb](https://github.com/andzajan/galaxytools/commit/6c53cdb7fd07c244a48883f10fc867e0173ec75c);

*  Tab is used for value field separator instead of space in the output file. [f2cf5aca](https://github.com/andzajan/galaxytools/commit/f2cf5acaa9c247f095f681628d78777eed1feb71);

* Changed order of the columns in output, molecule label is the first followed by SMILES string. [672fbf4](https://github.com/andzajan/galaxytools/commit/672fbf492173ce42d4b447b674f877224e1b2651);

* Added column labels for output files. [8166711c](https://github.com/andzajan/galaxytools/commit/8166711cff52df619cf5c84514bde5249548a43a) and [950e6d](https://github.com/andzajan/galaxytools/commit/950e6d838be3f7aa736728699f9098d4af5d1c8b);

*  Updated xml file and added tests. [ade7b14b](https://github.com/andzajan/galaxytools/commit/ade7b14b81a03012c3c693205ea0c5c4c6827be5), [6e15f969](https://github.com/andzajan/galaxytools/commit/6e15f9694f2a492c834d080800b9f28b4f60eb98) and [df7a2d1](https://github.com/andzajan/galaxytools/commit/df7a2d198849f4006f974cbcfb975985664895d0). 

Open questions we have so far:

* Are you happy with changes made to the output format? Or should we keep default output as it was? `smi` format with smiles in the first column and molecule identifier after space?

* What convention we should use for tool version numbering?

* We want to introduce another output where detailed output is summarised by the unique molecular formula. Easiest way to implement this is to use `pandas` instead of `numpy`. Are you happy for us to change a script to use `pandas`?

Ping @RJMW


simonbray:
Hey @andzajan, thanks a lot for the changes. This looks good to me.

> Are you happy with changes made to the output format? Or should we keep default output as it was? smi format with smiles in the first column and molecule identifier after space?

The default SMILES format has the string in the first column. I think it would be preferable to stick to this. Do you have a usecase where you need to swap the columns?

> What convention we should use for tool version numbering?

I don't have a strong view about this - maybe 1.1.1.1?

> We want to introduce another output where detailed output is summarised by the unique molecular formula. Easiest way to implement this is to use pandas instead of numpy. Are you happy for us to change a script to use pandas?

Yes, fine.

bgruening:
For the versions, we do have these recommenations: https://galaxy-iuc-standards.readthedocs.io/en/latest/best_practices/tool_xml.html#tool-versions

@andzajan thanks a lot. This looks great, I will modify the version and get it in.

Thanks,
Bjoern

simonbray:
> The default SMILES format has the string in the first column. I think it would be preferable to stick to this. Do you have a usecase where you need to swap the columns?
@andzajan can you reply to this before we merge?

andzajan:
Hi @simonbray and @bgruening, I agree with Simon. Happy to change column order and add extra argument to the python code and tool itself, to summarise outputs by chemical formula.

bgruening:
@andzajan cool, please ping us if you are ready.

chemicaltoolbox/sygma/sygma.xml
```diff
@@ -1,27 +1,31 @@
 <tool id="sygma" name="Generate possible metabolites with SyGMa" version="@VERSION@">
+    <description>by performing common reactions on one or more parent molecule(s)</description>
     <macros>
-        <token name="@VERSION@">1.1.1</token>
+        <token name="@VERSION@">1.1.1+galaxy1</token>
```
bgruening: if you putthis +galaxy1 into the `<tool>` than we can keep the macro.

bgruening:
ping @simonbray

bgruening:
@andzajan thanks a lot!

PR #959, opened 2020-03-17
initial version of SDF filter + sort tool

tdudgeon:
Docs not done yet, and tests need improving.


chemicaltoolbox/rdock/sort_filter.xml
```diff
@@ -0,0 +1,60 @@
+<tool id="rdock_sort_filter" name="SDF sort and filter" version="0.1.0">
+    <description>- sort and filter SDF</description>
+    <macros>
+        <import>rdock_macros.xml</import>
+    </macros>
+    <expand macro="requirements"/>
+    <command><![CDATA[
+
+sdsort -n $descending -s -f$sort_field -id$name_field '$input' | sdfilter -f'\$_COUNT <= $top' -s$name_field | sdsort -n $descending -f$sort_field > '$output'
```
bgruening: Sort field and name field need to be quoted for security reasons

chemicaltoolbox/rdock/sort_filter.xml
```diff
@@ -0,0 +1,60 @@
...
+
+    ]]></command>
+
+    <inputs>
+        <param type="data" name="input" format="sdf" label="Molecules" help="Molecules in SDF format"/>
+        <param name="top" type="integer" value="1" label="Number of best poses" help="Number of best scoring records to keep"/>
```
simonbray:
```suggestion
        <param name="top" type="integer" value="1" label="Number of records to keep in output" help="Number of best scoring records to keep"/>
```

simonbray: Can we make the label a bit more general, I guess this tool can be used for filtering for other variables than docking scores?

simonbray:
Can work on improving the test if this would be helpful.

PR #960, opened 2020-03-17
add the most important PDB output

chrisbarnettster:
Hi @bgruening !

I've modified fpocket. The PDB file *_out.pdb is the most important output (see page 31 of the fpocket manual to confirm - https://github.com/Discngine/fpocket/blob/master/doc/manual_fpocket2.pdf)
I've included this output and made sure that this and the all_verts output are selected by default. 
These two outputs are needed for visualisation of pockets and are used by the pymol and VMD visualisation scripts that are usually used with fpocket. 
Note that the PDB collection (atoms_output) does not contain the same information as found in *_out.pdb. This file is unique and important :smiley: Also updated the testcase.

All the best,
Chris


chrisbarnettster:
thanks @bgruening @simonbray

simonbray:
Thanks Chris!

PR #963, opened 2020-03-21
update split_fasta

bernt-matthias:
- remove file extensions from identifiers
- update to biopython 1.76 (since 1.65 seems not installable -- maybe
because biopython is now in conda-forge which does not include this
version?)
- remove tool_dependencies.xml

no idea why the test data changed, butr all sequences are in.


bgruening:
Merci!

mvdbeek:
@guerler would like to use this tool, but it seems it never made it into the tool shed. Can you upload this please @bgruening ?

bgruening:
Done!

PR #966, opened 2020-03-29
initial version of diff command wrapper

davelopez:
Simple wrapper for the GNU diff command


tools/diff/macros.xml
```diff
@@ -0,0 +1,5 @@
+<macros>
+    <token name="@TOOL_VERSION@">3.6</token>
```
bgruening: If this is just a few lines, you could move this macro directly into the main file.

tools/diff/diff.xml
```diff
@@ -0,0 +1,91 @@
+<tool id="diff" name="diff" version="@TOOL_VERSION@+@WRAPPER_VERSION@">
+
+    <description>analyzes two files and generates a text file with information about the lines that are different.</description>
```
bgruening: Do you know if `diff` is part of the coreutils package? If so we should probably depend on this package.

tools/diff/macros.xml
davelopez: Done!

tools/diff/diff.xml
```diff
@@ -0,0 +1,87 @@
+<tool id="diff" name="diff" version="3.6+galaxy0">
+
+    <description>analyzes two files and generates a text file with information about the lines that are different.</description>
```
beatrizserrano:
```suggestion
    <description>analyzes two files and generates a text file with information about the lines that are different</description>
```

tools/diff/diff.xml
davelopez: Added conda-forge package for diffutils 3.6 and requirement 😃

bgruening:
Merci! Great contribution.

PR #967, opened 2020-04-01
Update chembl version to 0.10.1

simonbray:


simonbray:
Thanks @bgruening, could you add the new version to the server if you have a spare minute? I noticed the old version is causing the docking workflow from the CTB paper to fail.

PR #968, opened 2020-04-02
add seed and filter-options to rdock, terminate if job gets stuck in a loop

bgruening:


chemicaltoolbox/rdock/rbdock.xml
```diff
@@ -13,15 +13,33 @@ ln -s $receptor_prm receptor.prm &&
...
+       -r receptor.prm
+       -p dock.prm
+       -n $num
+       -o output
+       #if str($seed):
+        -s $seed
+       #end if
+&&
+
+
```
nsoranzo, line 27:
```suggestion
```

chemicaltoolbox/rdock/rbdock.xml
```diff
@@ -13,15 +13,33 @@ ln -s $receptor_prm receptor.prm &&
...
+sdsort -n -s -fSCORE output.sd 
+
+#if $filter.filter_select == "filter":
+    |
```
nsoranzo, line 31:
```suggestion
```

chemicaltoolbox/rdock/rbdock.xml
```diff
@@ -13,15 +13,33 @@ ln -s $receptor_prm receptor.prm &&
...
+    #if $filter.score:
+      sdfilter -f'\$SCORE <= $filter.score' |
```
nsoranzo, line 33:
```suggestion
        | sdfilter -f'\$SCORE <= ${filter.score}'
```

chemicaltoolbox/rdock/rbdock.xml
```diff
@@ -13,15 +13,33 @@ ln -s $receptor_prm receptor.prm &&
...
+    #end if
+    #if $filter.nscore:
+      sdfilter -f'\$SCORE.norm <= $filter.nscore' |
```
nsoranzo, line 36:
```suggestion
        | sdfilter -f'\$SCORE.norm <= ${filter.nscore}'
```

chemicaltoolbox/rdock/rbdock.xml
```diff
@@ -13,15 +13,33 @@ ln -s $receptor_prm receptor.prm &&
...
+    #end if
+    sdfilter -f'\$_COUNT <= $filter.top'
```
nsoranzo, line 38:
```suggestion
    | sdfilter -f'\$_COUNT <= $filter.top'
```

bgruening:
Thanks @nsoranzo ... we had some problems with molecules that do not terminate. There is now a test that should catch this.

bgruening:
Arg, this seems to work with 19.09, but not with 20.01 ...

simonbray:
What about the idea of wrapping rbdock in the python script? Or is this overkill?

Really don't get why this doesn't work. Just out of curiosity, I'll check my grep solution with Galaxy 19.09 and see if it works.

simonbray:
> What about the idea of wrapping rbdock in the python script?

Have written something for this, just testing it now.

simonbray:
Okay, whether this is a good solution is very debatable but tests are passing for me locally.

simonbray:
I'm okay to merge I think, what do you think @bgruening @tdudgeon?

chemicaltoolbox/rdock/rbdock.py
```diff
@@ -0,0 +1,29 @@
+import subprocess
+import argparse
+
+def main():
+    parser = argparse.ArgumentParser(description='Simple wrapper for rbdock')
+    parser.add_argument('-n', '--num', type=int, help='Equilibrated system as input')
```
tdudgeon, line 6: Description needs updating?

tdudgeon:
I think we'll need to test this carefully, but let's go for it and deploy?

chemicaltoolbox/rdock/rbdock.py
simonbray: Oh yes, thanks for the catch

PR #971, opened 2020-04-07
Added diff2html template report

davelopez:
In addition to the diff raw text file, an HTML report is generated with a more friendly visual representation of the differences.

Any improvement suggestions and ideas are greatly appreciated :)


tools/diff/diff.xml
```diff
@@ -12,40 +13,93 @@
     </stdio>
 
     <command ><![CDATA[
-        diff -u '$input1' '$input2' > '$output1'
+        diff -u '$input1' '$input2' > '$diff_file';
+        #if $generate_report:
+            sed -e 's/@@title@@/Diff report for $input1.name and $input2.name/g'
```
bgruening, line 18: $input1.name --> use $input.element_identifier this should also work for collections. And please put them into single quotes.

davelopez: Done! Thanks for the comments!
If you see anything else or have some more usability ideas, please let me know :)

tools/diff/diff.xml
```diff
@@ -12,40 +13,92 @@
     </stdio>
 
     <command ><![CDATA[
-        diff -u '$input1' '$input2' > '$output1'
+        diff -u '$input1' '$input2' > '$diff_file';
+        #if $generate_report:
+            sed -e 's/@@title@@/Diff report for '$input1.element_identifier' and '$input2.element_identifier'/g'
```
bgruening, line 18: lets see if this passes now, are you not closing the outer `'` here?

tools/diff/diff.xml
```diff
@@ -12,40 +13,93 @@
...
+        #end if
+        
     ]]>
     </command>
 
     <inputs>
         <param name="input1" type="data" format="txt" label="First input file" />
         <param name="input2" type="data" format="txt" label="Second input file" />
+        <param name="generate_report" type="boolean" checked="true" truevalue="true" falsevalue="false" label="Generate HTML report" help="Generates an HTML report to visualize the differences"/>
+        <section name="adv" title="Advanced Report Options" expanded="false">
```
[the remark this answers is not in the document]
davelopez: Yeah, I got a bit carried away...

tools/diff/diff.xml
davelopez: I thought the same, but... It works and passes the tests :)

bgruening:
Cool, thanks a lot!

PR #972, opened 2020-04-07
transfs with multiple models

tdudgeon:
@bgruening Here is an initial attempt at an updated transfs tool.
The key change is that the user can choose from a number of models to use for the predictions.
Those models are present as files in the container in `/train/fragalysis_test_files/resources`.
There are 4 in total to choose from.

This obviously needs a new container image (`informaticsmatters/deep-app-ubuntu-1604:1.2`) that is a bit different from the one currently being used. The relevant files were moved from `/root/train/fragalysis_test_files` to `/train/fragalysis_test_files` and permissions on the files/dirs changed so that a normal user can run things. This means that most of your hacks to get round the original permissions problems won't be needed now, but I've not changed anything related to that, and I'm obviously not able to test anything at this stage.

So this PR is a long way off being able to merge, but hopefully it's an early heads up on what is needed.

### DIFF
--- a/chemicaltoolbox/xchem-deep/transfs.py
+++ b/chemicaltoolbox/xchem-deep/transfs.py
@@ -210,14 +210,14 @@ def generate_predictions_filename(work_dir, predict_file_name):
     return "{0}{1}{2}".format(work_dir, os.path.sep, predict_file_name)
 
 
-def run_predictions():
+def run_predictions(model):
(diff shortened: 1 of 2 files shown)


bgruening:
Version to test: https://usegalaxy.eu/root?tool_id=xchem_transfs_scoring

PR #974, opened 2020-04-09
change folder name regex in NameAndTypes module in common.xml

sunyi000:
update 'select folder regex' to use 'select' instead of text. similar to the file regex implementations in the same module


bgruening:
Thanks.

PR #975, opened 2020-04-09
Openduck

simonbray:


.travis.yml
```diff
@@ -123,10 +123,10 @@ script:
   #- while sleep 9m; do echo "=====[ $SECONDS seconds still running ]====="; done &
   - |
     if [ -s changed_tools_chunk.list ]; then
-        planemo test --galaxy_branch "$GALAXY_RELEASE" --galaxy_source "$GALAXY_REPO" $(cat changed_tools_chunk.list)
+        planemo test --conda_channels omnia,bioconda,conda-forge,defaults --galaxy_branch "$GALAXY_RELEASE" --galaxy_source "$GALAXY_REPO" $(cat changed_tools_chunk.list)
     elif [ -s changed_repositories_chunk.list ]; then
         while read -r DIR; do
-            planemo test --galaxy_branch "$GALAXY_RELEASE" --galaxy_source "$GALAXY_REPO" "$DIR"
+            planemo test --conda_channels omnia,bioconda,conda-forge,defaults --galaxy_branch "$GALAXY_RELEASE" --galaxy_source "$GALAXY_REPO" "$DIR"
         done < changed_repositories_chunk.list
```
simonbray, line 130: these changes need to be reverted before merging

simonbray:
This also needs to run on GPU: so far it's just being tested on CPU.

Each of the scripts has a parameter `--gpu-id` which if set creates a parameter `CudaDeviceIndex` that gets fed to openmm.

bgruening:
@simonbray use an env variable like `OPENDUCK_GPU_PARAM`. Then the Galaxy admin can overwrite the setting with `OPENDUCK_GPU_PARAM="--gpu-id"`

simonbray:
@bgruening, I think this is ready, though the tests will fail.

Maybe you can check I got the GPU param right here: https://github.com/bgruening/galaxytools/pull/975/files#diff-cb622fa2ade3568f8dd20924c3268638R16 and https://github.com/bgruening/galaxytools/pull/975/files#diff-dc5c8d6701beea182fcda184aa480d2eR15

chemicaltoolbox/openduck/openduck_run_smd.xml
```diff
@@ -0,0 +1,80 @@
...
+    <expand macro="requirements" />
+    <command detect_errors="exit_code"><![CDATA[
+        duck_smd_runs
+            -i $input
+            -p $pickle
+            -n $num
+            -l $mdlen
+            -d $start_dist
+            -v $vel
+            #if $OPENDUCK_GPU_PARAM:
```
bgruening, line 15: this variable does not exist ... if you what to consider an environment variable you need to include something like the GALAXY_SLOTS ... eg. `\${OPENDUCK_GPU_PARAM:-1}`

I would even say, the admin need to set `export OPENDUCK_GPU_PARAM="--gpu-id foo" `. So you can save here the if.

chemicaltoolbox/openduck/openduck_system_prep.xml
```diff
@@ -0,0 +1,90 @@
+<tool id="openduck_system_prep" name="OpenDUck system preparation" version="@VERSION@">
+    <description>for dynamic undocking</description>
+    <macros>
+        <import>macros.xml</import>
+    </macros>
+    <expand macro="requirements" />
+    <command detect_errors="exit_code"><![CDATA[
+        #set OPENDUCK_GPU_PARAM=1
```
bgruening, line 8: the above, you are setting here a cheetah variable. But you want to have an evironment variable `\${OPENDUCK_GPU_PARAM:-1}` --> note the backslash.

chemicaltoolbox/openduck/macros.xml
```diff
@@ -0,0 +1,36 @@
+<macros>
+    <token name="@VERSION@">0.1.1</token>
+    <xml name="requirements">
+        <requirements>
+            <requirement type="package" version="3.7">python</requirement>
+            <requirement type="package" version="0.25.2">yank</requirement>
```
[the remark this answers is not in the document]
simonbray: bioconda will not allow dependencies from omnia

chemicaltoolbox/openduck/openduck_run_smd.xml
```diff
@@ -0,0 +1,80 @@
...
+                --gpu-id $OPENDUCK_GPU_PARAM
+            #end if &&
+        tar -czf allfiles.tar.gz --exclude=allfiles.tar.gz * &&
+        get_wqb -l $ligand -o $ligand_wqb
+    ]]></command>
+    <inputs>
+        <param argument="input" type="data" format='binary' label="Chunk file from DUck equilibration"/>
+        <param argument="pickle" type="data" format='binary' label="Pickle file from DUck equilibration"/>
```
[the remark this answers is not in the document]
simonbray: Pff. I am just implementing the tool the way it's been written.

One option could be to combine the prep and run tools together. I don't think there is much advantage to using them individually anyway.

chemicaltoolbox/openduck/openduck_run_smd.xml
simonbray: Ah okay, makes sense. Just to be clear, if `--gpu-id` is not set the tool is run on CPU only. So in this case the admin would do `export OPENDUCK_GPU_PARAM=""`.

chemicaltoolbox/openduck/openduck_run_smd.xml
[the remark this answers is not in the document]
bgruening: That would be one option, just squeeze them together :+1:

chemicaltoolbox/openduck/openduck_run_smd.xml
bgruening: yes, or does not set this ENV at all. I guess `\${OPENDUCK_GPU_PARAM:-''}` should work in this case.

chemicaltoolbox/openduck/macros.xml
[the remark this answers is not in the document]
bgruening: and networkx?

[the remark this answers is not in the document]
simonbray: yeah, okay... networkx could be in the bioconda recipe. Would you like me to make a PR?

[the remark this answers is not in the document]
simonbray: yeah, okay... networkx could be in the bioconda recipe. Would you like me to make a PR?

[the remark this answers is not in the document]
bgruening: I will take care or skip it ... we either way need get this al of those into conda-forge in the long-run.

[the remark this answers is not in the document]
simonbray: I agree, unfortunately it's not really in our hands. There were already huge debates about this on conda-forge. Hopefully the omnia people will manage to resolve this soon.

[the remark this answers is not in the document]
simonbray: Made a new build of the conda package with networkx as a dep: https://github.com/bioconda/bioconda-recipes/pull/21528

simonbray:
Hopefully those things are all fixed in the latest commit @bgruening 

Tests are passing locally.

chemicaltoolbox/openduck/openduck_chunk.xml
```diff
@@ -0,0 +1,79 @@
...
+    <command detect_errors="exit_code"><![CDATA[
+        cp '$ligand' ./ligand.sdf &&
+        duck_chunk
+            -p '$protein'
+            -l ./ligand.sdf
+            -c '$cutoff'
+            $ignore_buffers
+            ## -i is interaction, e.g. A_ASP_156_OD2
+            -i '${ia.chain_sel}'_'${ia.res_sel}'_'${ia.resid_sel}'_'${ia.atom_sel}' &&
+        tar -czf allfiles.tar.gz --exclude=allfiles.tar.gz *
```
bgruening, line 16: this taring can be conditioned with the `return_tar`

chemicaltoolbox/openduck/openduck_chunk.xml
```diff
@@ -0,0 +1,79 @@
...
+        <expand macro="interaction_params" />
+        <expand macro="tar_param" />
+
+    </inputs>
+    <outputs>
+        <data name="chunk" format="pdb" from_work_dir="protein_out_prot.pdb" label="${tool.name} PDB file"/>
+        <expand macro="tar_output" />
+    </outputs>
+    <tests>
+        <test>
```
bgruening, line 35: add expect_num_outputs?

chemicaltoolbox/openduck/openduck_run_smd.xml
```diff
@@ -0,0 +1,105 @@
...
+        ## to run on GPU set OPENDUCK_GPU_PARAM to something like "--gpu-id 1"
+        cp '$ligand' ./ligand.sdf &&
+        
+        duck_prepare_sys
+            -p '$protein'
+            -l ./ligand.sdf
+            -c '$chunk'
+            ## -i is interaction, e.g. A_ASP_156_OD2
+            -i '${ia.chain_sel}'_'${ia.res_sel}'_'${ia.resid_sel}'_'${ia.atom_sel}' 
+            `echo "\${OPENDUCK_GPU_PARAM}"` &&
```
bgruening, line 17:
```suggestion
            \${OPENDUCK_GPU_PARAM} &&
```

bgruening: I hope this works.

chemicaltoolbox/openduck/openduck_run_smd.xml
```diff
@@ -0,0 +1,105 @@
...
+
+        duck_smd_runs
+            -i eql.chk
+            -p cs.pickle
+            -n '$num'
+            -l '$mdlen'
+            -d '$start_dist'
+            -v '$vel'
+            `echo "\${OPENDUCK_GPU_PARAM}"` &&
+        tar -czf allfiles.tar.gz --exclude=allfiles.tar.gz * &&
```
bgruening, line 30: make that conditional?

chemicaltoolbox/openduck/openduck_run_smd.xml
simonbray: Yes, it's pretty nasty. Otherwise python gives something like `error: unrecognized arguments:''`

chemicaltoolbox/openduck/openduck_chunk.xml
simonbray: Whoops, will add

chemicaltoolbox/openduck/openduck_run_smd.xml
simonbray: will try your version out

simonbray:
Thanks for the shed.yaml file!

PR #978, opened 2020-04-10
add metadata tags field1...6 to group category

sunyi000:
because regex for file name was changed to 'field1....field6', these are cellprofiler metadata tags and must be made available for selection in the 'group category' parameter within "Group" section of common.xml


bgruening:
Thanks @sunyi000 ... new release is coming.

PR #980, opened 2020-04-13
Rdock filter

tdudgeon:
This enhances the rdock sort tool by adding the ability to filter using arbitrary expressions, and makes sorting and grouping optional.


chemicaltoolbox/rdock/sort_filter.xml
```diff
@@ -1,38 +1,65 @@
...
+> '$output'
 
     ]]></command>
 
     <inputs>
         <param type="data" name="input" format="sdf" label="Molecules" help="Molecules in SDF format"/>
-        <param name="top" type="integer" value="1" label="Number of records to keep in output" help="Number of best scoring records to keep"/>
-        <param name="sort_field" type="text" label="Field to sort on" optional="false" help="Name of the field to sort records by">
+        <repeat name="filters" title="Filters">
+            <param name="filter" type="text" label="Filter expression" optional="true" help="Perl expression for filter">
```
bgruening, line 30: The "Perl" here relevant? Its a simple math-expression, isn't it?
Do we need the repeat? Is that easier than just writing `or`?

chemicaltoolbox/rdock/sort_filter.xml
```diff
@@ -1,38 +1,65 @@
...
                 <valid initial="string.printable">
                     <remove value="&apos;"/>
                 </valid>
                 <mapping initial="none"/>
             </sanitizer>
-        </param>    
-        <param name="descending" type="boolean" label="Sort descending" truevalue="-r" falsevalue="" checked="true"
-               help="Generate the name field (first line) for cases where this is empty"/>
+        </param>
+        <param name="top" type="integer" value="1" label="Number of records to keep in output" optional="false" help="Number of best scoring records to keep"/>
```
bgruening, line 61: The help text is confusing I think. Is this just returning 1 record by default?

chemicaltoolbox/rdock/sort_filter.xml
tdudgeon: Technically it is a Perl expression. Any valid Perl code should be possible here. From the sdfilter help:
```
Filters SD records by data fields

Usage:	sdfilter -f'$<DataField> <Operator> <Value>' [-s<DataField>] [sdFiles]
or	sdfilter -f<filename> [-s<DataField>] [sdFiles]

Note:	Multiple filters are allowed and are OR'd together.
	Filters can be provided in a file, one per line.

	Standard Perl operators should be used. e.g.
	eq ne lt gt le ge for strings
	== != < > <= >= for numeric

	_REC (record #) is provided as a pseudo-data field
	if -s option is used, _COUNT (#occurrences of DataField) is provided as a pseudo-data field

	If SD file list not given, reads from standard input
	Output is to standard output
```

And yes, after I'd implemented the repeat I realised it wan't that useful.
A better approach might be to allow a multi-line text field and write that content to a config file. Again see the sdfilter help.

chemicaltoolbox/rdock/sort_filter.xml
tdudgeon: Using default values yes. Within each group of records (grouped by field $name_field) you end up with the $top best records (according to $sort_field and $descending).

chemicaltoolbox/rdock/sort_filter.xml
bgruening: @tdudgeon we need to make sure, __no__ arbitrary perl code can be executed here.
Maybe just allow those expressions (== != < > <= >=) all numbers and letters?

I don't mind the repeat, this is up to you.

chemicaltoolbox/rdock/sort_filter.xml
bgruening: But field_name is also optional, would it make sense to also keep "top" then optional?

tdudgeon: True, but top is ignored is field_name is not defined.

bgruening: On the commandline, but the user does not know that ... maybe at least adding this to the help?
I guess the proper way would be to model it with a conditional.

chemicaltoolbox/rdock/sort_filter.xml
tdudgeon: I removed the repeat from the filter as it didn't really add anything.

tdudgeon: > Maybe just allow those expressions (== != < > <= >=) all numbers and letters?

I'm not sure how best to restrict what can be entered as a filter. Certainly we must allow $ and text operators (eq ne lt gt le ge), probably round braces. Might also be nice to allow string matching (beginsWith, endsWith, contains - maybe using regexps?). But this might open things up too much.

chemicaltoolbox/rdock/sort_filter.xml
tdudgeon: I made all params optional.

chemicaltoolbox/rdock/sort_filter.xml
bgruening: I'm hesitant to allow all those. In the end you can do a lot with perl and escape the commandline. I would rather restrict it and if you really have advanced needs like beginsWith, endsWith, contains the user should convert it into tabular and proceed from there.

What do you think?

tdudgeon: I'm happy with that. At least for starters. For now I just need to use simple expressions like '$SCORE  > 0.5'.
We can extend later if we need.

PR #985, opened 2020-04-22
Easyjoin: we can change the temporary directory

FredericBGA:
We can set the --temporary-directory option for sort in XML. And the perl script will write also the result file in this directory.
It is useful for large files that don't fit in /tmp: `/tmp/sortA3aLjF: No space left on device`

### DIFF
--- a/tools/text_processing/text_processing/easyjoin.xml
+++ b/tools/text_processing/text_processing/easyjoin.xml
@@ -1,4 +1,4 @@
-<tool id="tp_easyjoin_tool" name="Join" version="@BASE_VERSION@.2">
+<tool id="tp_easyjoin_tool" name="Join" version="@BASE_VERSION@.3">
     <description>two files</description>
     <macros>
         <import>macros.xml</import>
@@ -13,6 +13,7 @@
         chmod +x sort-header &&
         perl $__tool_directory__/easyjoin
             $jointype
+            -T '$temporary_directory'
             -t $'\t'
             $header
             -e '$empty_string_filler'
@@ -53,6 +54,7 @@
                 </valid>
             </sanitizer>
         </param>
+        <param name="temporary_directory" type="text" value="/tmp" label="GNU sort's --temporary-directory option (can be changed to avoid 'No space left on device' error)."/>
     </inputs>
     <outputs>
         <data name="output" format_source="infile1" metadata_source="infile1"/>
@@ -65,6 +67,7 @@
             <param name="column2" value="1" />
             <param name="header" value="True" />
             <param name="jointype" value="-a 1 -a 2" />
+            <param name="temporary_directory" value="/tmp" />
             <output name="output" file="easyjoin_result1.tabular" />
         </test>
     </tests>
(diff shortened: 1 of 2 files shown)


tools/text_processing/text_processing/easyjoin.xml
```diff
@@ -53,6 +54,7 @@
                 </valid>
             </sanitizer>
         </param>
+        <param name="temporary_directory" type="text" value="/tmp" label="GNU sort's --temporary-directory option (can be changed to avoid 'No space left on device' error)."/>
```
bgruening, line 57: I don't think such an option should be exposed to the user. In theory, a user can enter here anything and Galaxy will try to write there ... this will also never be cleaned up.

bgruening:
@FredericBGA in your Galaxy instance you could simply export the `$TMP` variable to something that fits your needs better. This should work I guess. Like here: https://github.com/usegalaxy-eu/infrastructure-playbook/blob/master/files/galaxy/dynamic_rules/usegalaxy/tool_destinations.yaml#L745

tools/text_processing/text_processing/easyjoin.xml
FredericBGA: ok. As a quick fix I had set -T option(with our /workspace/tmp dir that can handle large files) in the `<command>` bloc. So I can do that again (but each time this tool will be updated, I will have to apply the patch, that's why I prefer creating PR)

Is the perl program update fine to you?

> this will also never be cleaned up

No, it's cleaned up I think. I've tried in my home dir, without `--debug` option the temp files are removed by `easyjoin`

FredericBGA:
> 
> 
> @FredericBGA in your Galaxy instance you could simply export the `$TMP` variable to something that fits your needs better. This should work I guess. Like here: https://github.com/usegalaxy-eu/infrastructure-playbook/blob/master/files/galaxy/dynamic_rules/usegalaxy/tool_destinations.yaml#L745

Ok, Il will try. Thank you.

tools/text_processing/text_processing/easyjoin.xml
bgruening: See my other comment. I think you can easily work-around this by setting the env variable for this tool at the admin level.

So if your /tmp is too small just re-set $TEMP to /var/tmp or something similar and it should work I assume. This is also what we do on your cluster by default.

FredericBGA:
> 
> 
> > @FredericBGA in your Galaxy instance you could simply export the `$TMP` variable to something that fits your needs better. This should work I guess. Like here: https://github.com/usegalaxy-eu/infrastructure-playbook/blob/master/files/galaxy/dynamic_rules/usegalaxy/tool_destinations.yaml#L745
> 
> Ok, Il will try. Thank you.

It works (of course). I've set `TMPDIR `in `config/local_env.sh`. Thank you. I will try to first think about configuring my env rather than trying to modifying Galaxy! (https://github.com/galaxyproject/planemo/issues/981#issuecomment-616647036)

PR #987, opened 2020-04-22
Bam output from salmon quant is unsorted

mvdbeek:
And leaving this at bam fails the metadata job because the file can't be
indexed.


bgruening:
Should we just sort it in this case?

bgruening:
Danke

PR #991, opened 2020-04-28
Change name of cut tool to avoid confusion with built-in cut tool

shiltemann:
I have been getting a lot of reports from people getting stuck in tutorials because they are using this cut tool instead of the built-in one. 

image

Maybe renaming this tool would avoid confusion about which to use?

The problem arises because this tool does not rearrange columns if you specify e.g. "c2,c5,c3", but keeps them in order. In the ATAC-Seq tutorial for instance, this leads to an invalid BED file. (we could also update this tool to do the column rearranging like the other cut tool)

### DIFF
--- a/tools/text_processing/text_processing/cut.xml
+++ b/tools/text_processing/text_processing/cut.xml
@@ -1,4 +1,4 @@
-<tool id="tp_cut_tool" name="Cut" version="@BASE_VERSION@.0">
+<tool id="tp_cut_tool" name="Advanced Cut" version="@BASE_VERSION@.0">
     <description>columns from a table (cut)</description>
     <macros>
         <import>macros.xml</import>


nsoranzo:
> we could also update this tool to do the column rearranging like the other cut tool

Unfortunately that's not so simple, since this tool is based on UNIX cut, which does not support column rearranging, see e.g. https://stackoverflow.com/questions/2129123/rearrange-columns-using-cut . We would need to rewrite the tool to use `awk` or Perl.

This is clearly confusing to users and we need to decide what is the better long-term solution for these 2 similar tools.

bgruening:
Thannks @shiltemann!

bgruening:
> This is clearly confusing to users and we need to decide what is the better long-term solution for these 2 similar tools.

Any idea about how to proceed here is very welcome!

nsoranzo:
> > This is clearly confusing to users and we need to decide what is the better long-term solution for these 2 similar tools.
> 
> Any idea about how to proceed here is very welcome!

I think the embedded tool is going to always be distributed with Galaxy and does everything the majority of users need. So I would just check that that's what we use in the training materials.

PR #1007, opened 2020-06-11
Add choice from Scipy

anuprulez:
Note: This does not change the output model

Thanks!


bgruening:
@anuprulez I think we should still bump the tool version isn't it?

anuprulez:
Ok, I will do it :)

PR #1011, opened 2020-06-27
update diamond to 2.0.8

bernt-matthias:
maybe the data manager could also be updated and the taxonomy options be included?


tools/diamond/diamond_view.xml
```diff
@@ -1,16 +1,17 @@
...
     <expand macro="requirements" />
     <expand macro="stdio" />
     <expand macro="version_command" />
-    <command><![CDATA[
+    <command detect_errors="aggressive"><![CDATA[
     ## need to link because diamont tries to open dataset_xxx.dat.daa
     ln -s '$daa' input.daa &&
     diamond
         view
+        --threads 1
```
bgruening, line 14: GALAXY_SLOTS?

bernt-matthias:
> I think at least the data manager needs to be updated and probably also versioned :(

you mean to version the entries in the data table or just to use a more useful tool version?

tools/diamond/diamond_view.xml
bernt-matthias: okay

bernt-matthias:
Hey @bgruening .. nearly there. 

Two questions: 
- wondering how to include the name mapping data from here: https://ftp.ncbi.nlm.nih.gov/pub/taxonomy/accession2taxid/ ? Separate data manager? Seems to be several GB of data.
- what did you mean with the versioning?

bernt-matthias:
Hi @bgruening did you have time to think about my questions in the last comment?

bgruening:
> wondering how to include the name mapping data from here: https://ftp.ncbi.nlm.nih.gov/pub/taxonomy/accession2taxid/ ? Separate data manager? Seems to be several GB of data.

mh, damn, yes probably yes :(

>     what did you mean with the versioning?

Diamond DBs are not compatible afaik with all tool versions. So I guess we need to do a similar trick as with snpeff or salmon.

bernt-matthias:
> mh, damn, yes probably yes :(

yep probably .. the only good alternatives that I could think of:
- take from history (not so nice)
- include the file in the taxonomy data manager (would require also some kind of flag column)

> Diamond DBs are not compatible afaik with all tool versions.

According to the changelog (https://github.com/bbuchfink/diamond/releases/) this happened last for v0.9.19. So maybe we can postpone this. Otherwise: you mean to add a new column?

bgruening:
> According to the changelog (https://github.com/bbuchfink/diamond/releases/) this happened last for v0.9.19. So maybe we can postpone this. Otherwise: you mean to add a new column?

ok, if you have tested is, then we can postpone it. And yes, otherwise we need to add a new column to the diamond.loc file with the version constrained. The tool filters then the DBs based on this column.

bgruening:
@bernt-matthias I think we should add `--ignore=E501 ` to the linting of python files.

nsoranzo:
> @bernt-matthias I think we should add `--ignore=E501 ` to the linting of python files.

This would usually go into `setup.cfg`:

```
[flake8]
ignore = D,E501,W503
```

bernt-matthias:
created setup.cfg in a separate PR: https://github.com/bgruening/galaxytools/pull/1026

abretaud:
Hey, what's the status here @bernt-matthias ? I need version 2.0.8 (that still uses the same index version), I almost started a new PR until I saw this one
Can I help?

bernt-matthias:
@abretaud for me the last open point in this PR is the data manager. In particular the `--taxonmap` parameter which needs the data from https://ftp.ncbi.nlm.nih.gov/pub/taxonomy/accession2taxid/. The best would be to integrate this in the NCBI taxonomy data manager. But I could not wrap my head around the question how to do this in the best way.

Quite a big version jump .. exciting. According to the changelog it seems that are a couple of parameters that need to be adapted or added?

Feel free to add commits. Updating to 2.x sounds like a good idea.

Comments/ideas regarding the DM problem are very welcome.

abretaud:
Ok thanks!
Yes the version jump looks scary but I don't think there are breaking changes, I'll review the param changes.
For the DM, well I didn't plan to use it as we use biomaj here, but I can have a look

bgruening:
If we have so many people now interested in this and its so much better than my hack at the beginning I'm happy if you move this to IUC - if you want?

Thanks all!

abretaud:
Yep I can do the migration :+1:

bernt-matthias:
+1 

Maybe we should also move https://github.com/galaxyproject/tools-iuc/tree/master/data_managers/data_manager_fetch_ncbi_taxonomy/ .. since this is somehow connected.

abretaud:
Hum, it's already in IUC :)
By the way I'm working on it ATM to optionally download accession2taxid data

bgruening:
Please feel free to move the good part here to IUC. The more eyes we have the better ... and let's remove tools here from the repo when we migrated them ... I think we usually put a readme into the folder stating where the tool is now living.

bernt-matthias:
> By the way I'm working on it ATM to optionally download accession2taxid data

Out of curiosity: What is your plan?

abretaud:
Here's the PR: https://github.com/galaxyproject/tools-iuc/pull/3537
Any comments welcome of course!

PR #1012, opened 2020-06-27
split_file_to_collection: allow large number of files

bernt-matthias:
by only opening one file at a time

fixes https://github.com/bernt-matthias/galaxytools/pull/new/topic/split-many

### DIFF
--- a/tools/text_processing/split_file_to_collection/split_file_to_collection.py
+++ b/tools/text_processing/split_file_to_collection/split_file_to_collection.py
@@ -3,8 +3,8 @@
 import argparse
 import math
 import os
-import re
 import random
+import re
 
 # configuration of the splitting for specific file types
 # - regular expression matching the record separator ('' if not splitting by regex but by number of lines)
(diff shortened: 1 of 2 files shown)


bgruening:
@bernt-matthias I guess this should fix 1009? I had something similar in https://github.com/bgruening/galaxytools/pull/new/fix_1009   maybe its useful?

bernt-matthias:
Jep. Looks more or less the same.

bernt-matthias:
I compared the branches .. guess this one is slightly better (e.g. if files are written successively then each one is opened only once .. compared to for each record)

bgruening:
Please bump the tool version and feel free to merge. Thanks!

bernt-matthias:
Sounds good, but I can't merge :)

mvdbeek:
@bgruening This didn't upload correctly (see version 0.4.0 in https://toolshed.g2.bx.psu.edu/view/bgruening/split_file_to_collection/d57735dd27b0), but there was a minor update to 0.4.0 that broke the existing tool (outputs are not written to out/, and so the dataset discovery fails on main). So could you first upload the old revision of 0.4.0 (you can checkout e6267aab728dafc7c7d50b899113dedf1366671b, that should be the correct version), and then upload from master the 0.5.0 revision ?

PR #1015, opened 2020-07-07
antismash5

bgruening:


bgruening:
@abretaud is there any interest to update this wrapper? Do you follow the latest updates of Antismash? What is fungismash? So many questions :)

abretaud:
Hello!
There is probably interest, but not particularly from me as I don't use it :) So no, I'm not following the updates, but we can merge this one I guess as long as the tests pass
I would say fungismash is the antismash of fungi (antismash being for bacteria), but I don't know more than that

abretaud:
ah, antismash was in tt_skip, I just removed it to see what it does

simonbray:
Maybe I can (should?) adopt this wrapper?

bgruening:
@simonbray all help is welcome. I upgraded the wrapper, but I can not test. For whatever reason, every install fails to download the databases ... :(

bgruening:
Ok, in the Container that I build it is running and test pass locally. I guess this is a travis issue as during installation 4Gb will be downloaded and several indices will be installed, e.g. via Diamond.

PR #1017, opened 2020-07-20
XCos and featurestein tools

tdudgeon:
Initial versions of 3 new tools:

XCos: a standalone tool that can be used to compare the overlap of ligand poses with a set of fragment screening hits
FeatureStein: a pair of tools, one that generates a merged RDKit features map from a set of fragments and the second that uses this merged feature map to score a set of ligand poses.

This IS NOT yet in a state to merge. The XCos tool is working OK.
The FeatureStein tools need a new data type for the merged feature map file, which is a Python pickle of the merged feature map.
Currently this is using `binary` file type but this is not working for the scoring tool. I think a new file type needs to be added?
@bgruening How does this get done?


chemicaltoolbox/im-pipelines/featurestein-score.xml
```diff
@@ -0,0 +1,71 @@
...
+    </macros>
+    <expand macro="requirements" />
+    <command detect_errors="exit_code">
+<![CDATA[
+        python -m pipelines.xchem.featurestein_score -i '$input' -if sdf -f '$featuremaps' -o output -of sdf --no-gzip &&
+        mv output.sdf '$output'
+]]>
+    </command>
+    <inputs>
+        <param name="input" format="sdf" type="data" label="Ligands to score" help="In SDF format"/>
```
simonbray, line 15: Using a pickle as input is not a good idea from the security point of view.

chemicaltoolbox/im-pipelines/featurestein-generate.xml
```diff
@@ -0,0 +1,67 @@
+<tool id="featurestein_generate" name="Generate 'FeatureStein' merged features maps" version="0.1">
```
simonbray, line 1: suggestion: use the same version number as the im-pipelines repo, i.e. 1.1.3

chemicaltoolbox/im-pipelines/featurestein-generate.xml
```diff
@@ -0,0 +1,67 @@
...
+    <description>feature overlay scoring</description>
+    <macros>
+        <import>macros.xml</import>
+        <token name="@GALAXY_VERSION@">0</token>
+    </macros>
+    <expand macro="requirements" />
+    <command detect_errors="exit_code">
+<![CDATA[
+        python -m pipelines.xchem.featurestein_generate -i '$input' -if sdf -f '$output'
```
simonbray, line 10:
```suggestion
        featurestein_generate -i '$input' -if sdf -f '$output'
```

chemicaltoolbox/im-pipelines/featurestein-generate.xml
simonbray: oh I see you did already in the macro - then just use `@TOOL_VERSION@+@GALAXY_VERSION@`

tdudgeon:
@simonbray Thanks for those comments. I'll make the changes, but first I think we need to decide on how to handle saving the featuremaps for the featurestein tools. They need to be saved somehow, and a pickle is the only way I know how to do this.
And I think we need a new data type for this?

bgruening:
@tdudgeon we should not use pickles. Never. They are insecure by default, as they can execute arbitrary code. On a multi-user-system this is not good. What kind of data is that? Can we store it in hdf5?

bgruening:
@tdudgeon can you rebase this PR. We already included xcos in a nother PR. Thanks.

tdudgeon:
@bgruening What is being saved is a RDKit FeatMap (http://rdkit.org/docs/source/rdkit.Chem.FeatMaps.FeatMaps.html?highlight=featmap#rdkit.Chem.FeatMaps.FeatMaps.FeatMap)
RDKit does not seem to provide any way to save these, hence why they are pickled.

tdudgeon:
@simonbray Combining the tools is one option, but there is good reason to be able to save the feature maps.
I've asked on the RDKIt mailing list if there is a better way of saving.

simonbray:
@tdudgeon, producing a pickle output is not a problem, it's using one as an input.

So you could have a combined tool which (perhaps optionally) provides the pickle as well.

tdudgeon:
I created a single combined featurestein tool, and made some changes to the xcos tool to reflect changes in the upstream code.

PR #1024, opened 2020-09-03
Add note to Flye and Canu

willemdek11:
NOTE:  This tool may take a long time depending on the size and characteristics of your dataset.


bgruening:
Upps, sorry, missed that. Do you still want to add it @willemdek11 ?

willemdek11:
@bgruening I think it is fine without it, don't worry :)

PR #1028, opened 2020-09-14
fixes for pgchem

bernt-matthias:
just stumbled over this one. 

would need a loc file to make the test work... any suggestions @bgruening ?


bgruening:
Are you using this tool with an external PG database? This tool is really old and I think we should invest more in https://github.com/bgruening/galaxytools/pull/642 and the direction.

bernt-matthias:
> Are you using this tool with an external PG database?

No .. just stumbled over it -- Not using it at all. Then lets remove the test and take this minor enhancement -- even if not really relevant.

bgruening:
Thanks.

PR #1036, opened 2020-10-02
Lint sklearn python scripts, test without --biocontainers, install PCA tool

simonbray:


.github/workflows/pr.yaml
```diff
@@ -300,11 +300,10 @@ jobs:
     - name: Planemo test tools
       run: |
         if grep -lqf .tt_biocontainer_skip changed_tools_chunk.list changed_repositories_chunk.list; then
-                PLANEMO_OPTIONS="--job_config_file job_conf.xml_condor"
+                PLANEMO_OPTIONS="--job_config_file job_conf_no_biocontainers.xml_condor"
```
bgruening, line 303: Clever! However, I guess I need to find out why this container is failing.

bgruening:
@simonbray thanks so much for working on that. I already pushed the missing tool to the TS and it will be installed tm.

.github/workflows/pr.yaml
simonbray: This change still seems to be needed, right? Otherwise the biocontainers skiplist is doing nothing

bgruening:
@simonbray did you figure out why the container is not working?

simonbray:
No, sorry, not yet.

Where is it coming from? It is not from the biocontainers repo afaict:

```

2020-10-02 13:50:47,751 INFO  [galaxy.tool_util.deps.containers] Checking with container resolver [CachedMulledDockerContainerResolver[namespace=biocontainers]] found description [None]
2020-10-02 13:50:47,829 INFO  [galaxy.tool_util.deps.containers] Checking with container resolver [CachedMulledDockerContainerResolver[namespace=local]] found description [ContainerDescription[identifier=quay.io/local/mulled-v2-3e165e0a0c9317555ab38ffaec2066080f4117fa:c6973b6230c9931cf47fae10af628285bc85f748-0,type=docker]]
```

simonbray:
Shouldn't this be merged? The biocontainers skiplist is doing nothing at the moment.

PR #1030, opened 2020-09-22
[awk] Using right awk in biocontainers

almahmoud:
This commit: https://github.com/bgruening/galaxytools/commit/7bfc12bbe0f36602125eddde0e3b162ebe3345bf makes the corresponding biocontainer fall over with the kubernetes runner cause it uses Busybox's `awk` instead of `gawk`.

```
(base) BIOL-JT-ALEX:~ alex$ docker run -it --rm quay.io/biocontainers/gawk:4.2.0 'bash'
bash-4.2# env -i awk --help
BusyBox v1.22.1 (2014-05-23 01:24:27 UTC) multi-call binary.
Usage: awk [OPTIONS] [AWK_PROGRAM] [FILE]...
    -v VAR=VAL    Set variable
    -F SEP        Use SEP as field separator
    -f FILE        Read program from FILE
    -e AWK_PROGRAM

bash-4.2# awk --help
Usage: awk [POSIX or GNU style options] -f progfile [--] file ...
Usage: awk [POSIX or GNU style options] [--] 'program' file ...
POSIX options:        GNU long options: (standard)
    -f progfile        --file=progfile
    -F fs            --field-separator=fs
    -v var=val        --assign=var=val
[...]
gawk is a pattern scanning and processing language.
By default it reads standard input and writes standard output.
Examples:
    gawk '{ sum += $1 }; END { print sum }' file
    gawk -F: '{ print $1 }' /etc/passwd
```

```
bash-4.2# which awk
/usr/local/bin/awk
bash-4.2# env -i which awk
/usr/bin/awk
```

Tool fails with
```
awk: unrecognized option '--sandbox'
BusyBox v1.22.1 (2014-05-23 01:24:27 UTC) multi-call binary.

Usage: awk [OPTIONS]
etc...
```

PR #1041, opened 2020-10-22
[tp_sort_rows] print py2->py3

almahmoud:
Containerized tool test was failing due to the old print. This makes it work. Alternative is to set py2 hard requirement potentially, but this print should be compatible with 2 and 3.
(This could be considered a paper-cut, but I can't set the label in this repo)


jmchilton:
The last PR that was merged had similar errors which seem unrelated to the changes here: https://github.com/bgruening/galaxytools/pull/1030.

PR #1042, opened 2020-10-22
replace <> by != for python3 galaxy instance

scholtalbers:
Untested but I believe this correctly updates some wrappers to work on a python3 hosted Galaxy instance.

Seems also some autoformatting/whitespace removal happened - could remove that if need be.


bgruening:
Das ist cool. Jelle hast du Zeit das linting zu fixen? Kannst du bitte auch noch die tool version anheben? Danke.

scholtalbers:
@bgruening to what version should these be bumped, currently they are the same as the underlying tool. Of course, feel free to bump them yourself :)

PR #1044, opened 2020-11-02
adding fastq.gz to flye ext parser

NatPRoach:
Current version of flye doesn't work on fastq.gz files because it doesn't set the extension, this should fix that.


bgruening (approved):

bgruening:
Thanks @NatPRoach

PR #1047, opened 2020-11-10
Modify OB Remove counterions and fragments to optionally output indexed table

trachtok:
Make tool Remove counterions and fragments (OpenBabel) output tab-separated indexed table keeping original order of molecules, including molecules that did not pass filtering for heavy atoms. Currently, if user has list of multiple compounds and some of them have < 5 heavy atoms after tool removes counterions and fragments, such molecules are skipped when output is being written and user cannot track back which ones were actually removed. With this modification, user can choose to have indexed table as an output (default is old behaviour). Works only for InChI and SMILES, ignored for all other formats.


chemicaltoolbox/openbabel/ob_remIons.py
```diff
@@ -15,20 +16,25 @@ def parse_command_line():
...
     parser.add_argument('-i', '--input', required=True, help='input file name')
     parser.add_argument('-o', '--output', required=True, help='output file name')
+    parser.add_argument('-idx', default=False, action='store_true', help='should output be an indexed text table? works only for inchi/smiles, otherwise is ignored')
     return parser.parse_args()
 
 
 def remove_ions(args):
-    outfile = pybel.Outputfile(args.iformat, args.output, overwrite=True)
-    for mol in pybel.readfile(args.iformat, args.input):
+    outfile = open(args.output, 'w')
```
simonbray, line 24: Maybe a `with` block here?

chemicaltoolbox/openbabel/ob_remIons.xml
```diff
@@ -12,10 +12,12 @@
       -i '${infile}'
       -iformat '${infile.ext}'
       -o '${outfile}'
+      $index
 ]]>
     </command>
     <inputs>
         <expand macro="infile_all_types"/>
+        <param name="index" type="boolean" checked="false" truevalue="-idx" falsevalue="" label="Should output be indexed tab-separated table?" help="This option will work only when input format is InChI or SMILES, otherwise will be ignored!" />
```
simonbray, line 20:
```suggestion
        <param name="index" type="boolean" checked="false" truevalue="-idx" falsevalue="" label="Output indexed tab-separated table?" help="Add an index column to the output file. This option will work only when input format is InChI or SMILES, otherwise it will be ignored." />
```

chemicaltoolbox/openbabel/ob_remIons.py
```diff
@@ -15,20 +16,25 @@ def parse_command_line():
...
-            if mol.OBMol.NumHvyAtoms() > 5:
-                outfile.write(mol)
-    outfile.close()
+
+        mol = mol.write(args.iformat) if mol.OBMol.NumHvyAtoms() > 5 else '\n'
+
+        if args.idx and args.iformat in ['inchi', 'smi']:
+            outfile.write(f'{index}\t{mol}')
+        elif mol != '\n':
+            outfile.write(f'{mol}')
```
simonbray, line 37: Do you know if this works for other file formats as well? It's a pity this tool doesn't have a test for SD-files.

simonbray (commented):
This looks good to me, thanks @trachtok!

simonbray:
Björn just reminded me that if we are using f-strings (I am pro) we should also update the requirements to use only python 3.6 or greater. Can you therefore modify this line: https://github.com/trachtok/galaxytools/blob/add_index_table_remIons_tool/chemicaltoolbox/openbabel/macros.xml#L7

chemicaltoolbox/openbabel/ob_remIons.py
trachtok: In case of formats that are not one-liners this table setup with columns would not make much sense, so it works only for one-line formats.

chemicaltoolbox/openbabel/ob_remIons.xml
trachtok: Thanks, I will modify it ;)

chemicaltoolbox/openbabel/ob_remIons.py
trachtok: I thought about having it in `with` block but then removed it because it did not seem necessary. Do you have some point against it, why it would be better with `with` block? I am not exactly python expert so I am happy to learn :)

simonbray: The with block ensures the file gets closed properly, here you are just opening it. See here: https://www.python-course.eu/python3_file_management.php#Reading-and-Writing-Files-in-Python

trachtok: Ok, point made, I will use `with`, thanks for noticing :)

simonbray (approved):

bgruening:
Cool thanks @trachtok! This tool will automatically be deployed at EU on Saturday, if you need it earlier please let me know.

trachtok:
Thanks @simonbray and @bgruening!

bgruening:
Thanks to you!! :)

PR #1048, opened 2020-12-10
Replace column fixes

wm75:
- The tool would previously behave unexpectedly with empty columns at the end of lines (drop columns, retain lines that should  be skipped, etc)
- The input datatype is now preserved on output
- The code has also been cleaned a bit


tools/replaceColumn/replaceColumn.xml
```diff
@@ -35,24 +35,23 @@ with open("output_file", 'w') as output:
...
-                        out.append(line_content_column)
-                    elif unk_strat == "error":
-                        raise Exception('ERROR: Encountered a value [%s] in the file that is not in the replacements file and is not commented with [%s]' % (line_content_column, comment_str))
-                else:
-                    out.append(line_content_column)
-
-            if len(out) == len(line_content):
-                output.write('%s\n' % delimiter_local.join(out))
+            if column >= len(line_content):
+                continue
```
wm75, line 41: This is somewhat questionable, but preserves the behavior of the previous version.

bgruening: we could fail instead?

wm75: Yes, that's the simple alternative. We could also couple the behavior for short lines to the general mode (skip/error/pass), which would also not be too complicated to do.
I was just reluctant to change this cause it may force people to revise WFs when they update to the new tool version.
But if you think it's worth it I can change it to either of the two behaviors above.

bgruening: if we bump the version I think this is fine and afaik the supplied value, if this fails, is invalid

wm75: Ok, I'll change this. Regarding to what:
I can see use-cases for the more complex implementation where users would want to skip/keep unmodified truncated lines instead of erroring out (like certain lines of the input lacking an annotation so there's nothing to replace, but you don't want to lose these lines either) so I'd prefer being able to handle such situations.

bgruening:
I'm not sure why, but in containers, those tests all fail.

wm75:
Yes, but it seems that job execution as such is broken.

bgruening:
Rerunning tests now.

PR #1052, opened 2021-01-14
Update flye to 2.8.2

hexylena:
Genome size is no longer required, unless the --asm-coverage option is selected.

### DIFF
--- a/tools/flye/flye.xml
+++ b/tools/flye/flye.xml
@@ -84,7 +84,6 @@
         <test>
             <param name="inputs" ftype="fasta" value="nanopore.fasta"/>
             <param name="mode" value="--pacbio-raw"/>
-            <param name="g" value="10000"/>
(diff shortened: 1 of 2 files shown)


tools/flye/flye.xml
```diff
@@ -91,9 +100,10 @@
         <test>
             <param name="inputs" ftype="fasta" value="nanopore.fasta"/>
             <param name="mode" value="--nano-corr"/>
-            <param name="g" value="10000"/>
             <param name="i" value="2"/>
-            <param name="asm" value="40"/>
+            <param name="asm|asm_select" value="true" />
```
bgruening, line 104: can you please use the `<conditional name="">` syntax?

hexylena: oh, sure, can do.

bgruening:
Tool version also needs to be bumped as long as we not use the new IUC macro for versions.

Thanks @hexylena

bgruening:
Merci!

PR #1054, opened 2021-01-18
Added cmd_info.json to output from Alevin.

FloWuenne:
Hello,

This is a very small pull request to add 1 additional output file to Alevin when the user selects to output all files.

I wrote a galaxy wrapper for alevinQC (https://testtoolshed.g2.bx.psu.edu/view/florian.wuennemann/alevinqc/71f60bf852f6) to enable quality control plots for Alevin runs. alevinQC requires the cmd_info.json file from alevin, which is currently not saved to galaxy. I added the cmd_info.json file as an output with the filter option allout.

I tested the option with a local docker container running galaxy and it worked as expected.

Thanks for considering this PR, so the alevinQC wrapper can work downstream of this Alevin wrapper!


FloWuenne:
I would love to fix this PR so that all the tests are successful and this could be merged into main. 
Could anyone give me guidance on what is missing/causing the tests to fail?

PR #1062, opened 2021-02-03
Fix diff tool error when generating report

### LINKED ISSUE #1061: possible bug in diff html reports
Running [diff](https://github.com/bgruening/galaxytools/blob/master/tools/diff/diff.xml) on two files with the same content with requested html report produces the following output in the stderr.

```
sed: -e expression #1, char 78: unknown option to `s'
```

The tool does not fail. Report is empty. Maybe the report could say such.

davelopez:
There was an error in the quotes used in the `sed` command that was breaking the HTML report generation when comparing files containing spaces in the path.

Added a test case for this.

This should fix #1061


bgruening (approved):

bgruening:
Thanks for the quick fix.

PR #1066, opened 2021-02-16
Add option to calculate formal molecular charge to rdkit descriptor tool

simonbray:
Also, lint python files in the rdkit dir


chemicaltoolbox/rdkit/rdkit_get_charge.xml
```diff
@@ -0,0 +1,81 @@
+<tool id="ctb_rdkit_get_charge" name="Formal molecular charge" version="@TOOL_VERSION@+galaxy@GALAXY_VERSION@">
```
bgruening, line 1: from a user-point of view, should this be added to https://github.com/bgruening/galaxytools/blob/master/chemicaltoolbox/openbabel/ob_genProp.xml --> even if we cross the rdkit/openbabel boundary here (which is purely technical and does not make much sense to the user maybe?)

simonbray: I guess we could but I'm not so enthusiastic. I think our users are quite well informed about different packages and it should be clear from the tool which one they are using.

Or is your concern generally about creating a lot of tiny tools with very specific functionality?

bgruening: More the latter. More tools means more maintenance and more things that can break.

We could still offer them the information (charge (ob) vs. charge (rdkit)) but having two tools for charge calculation, just with a different implementation seems weird.

But I understand that this is not an important topic. So I'm happy to merge that if you prefer.

simonbray: There is also the rdkit descriptor tool. I can add it to that instead.

> We could still offer them the information (charge (ob) vs. charge (rdkit)) but having two tools for charge calculation, just with a different implementation seems weird.

I don't think charge can be calculated with the openbabel tool?

bgruening: > There is also the rdkit descriptor tool. I can add it to that instead.

Also good! You see we have already so many places. If you concerned about performance, because to many things are calculated we can make those optional. 

> I don't think charge can be calculated with the openbabel tool?

No it can't to my knowledge, it was just an example.

simonbray:
do not merge yet, I need to test again tomorrow

chemicaltoolbox/rdkit/rdkit_descriptors.py
```diff
@@ -1,76 +1,76 @@
...
-from rdkit.Chem import Descriptors
-from rdkit import Chem
-import sys, os, re
 import argparse
 import inspect
+import sys
+
+from rdkit import Chem
+from rdkit.Chem import Descriptors
+from rdkit_util import get_supplier
```
bgruening, line 9: We still have a problem with pulsar that those extra files (get_supplier.py) are not moved to the pulsar endpoint if they are not specified/called in the XML file.

So either you include (touch $__tool_directory__/get_supplier.py) in the command section or you leave this function in this main file.

chemicaltoolbox/rdkit/rdkit_descriptors.xml
```diff
@@ -14,14 +14,220 @@
             -i '${infile}'
             --iformat '${infile.ext}'
             -o '${outfile}'
+            -s '${select_multiple}'
```
bgruening, line 17: Will this be None in case of blank? `#if` might be more reliable?

chemicaltoolbox/rdkit/rdkit_descriptors.py
simonbray: Ah, good to know

bgruening (approved):
Good from my side! Thanks!

PR #1067, opened 2021-02-25
various bug fixes to the cellprofiler tool

sunyi000:
fixing bugs in cellprofiler modules

### DIFF
--- a/tools/cellprofiler/export_to_spreadsheet.xml
+++ b/tools/cellprofiler/export_to_spreadsheet.xml
@@ -1,7 +1,8 @@
-<tool id="cp_export_to_spreadsheet" name="ExportToSpreadsheet" version="@CP_VERSION@">
+<tool id="cp_export_to_spreadsheet" name="ExportToSpreadsheet" version="@CP_VERSION@+galaxy@VERSION_SUFFIX@">
     <description>export measurements into one or more files</description>
     <macros>
         <import>macros.xml</import>
+        <token name="@VERSION_SUFFIX@">1</token>
         <xml name="metadata_sample_row" >
             <param name="metadata_category" type="select" label="Select the metadata to use as the identifier">
                 <option value="FileName">File name</option>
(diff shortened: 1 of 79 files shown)


tools/cellprofiler/export_to_spreadsheet.xml
```diff
@@ -216,6 +216,7 @@ with open(input_pipeline) as fin:
                         <option value="Experiment">Experiment</option>
                         <option value="Object relationship">Object relationship</option>
                         <option value="Do not use">Do not use</option>
+                        <option value="Embryos">Embryos</option>
```
beatrizserrano, line 219: no I'm confused @sunyi000... shouldn't we remove that option?

sunyi000: ah sorry I forgot to remove this....

beatrizserrano: but here the user still needs to select the data_to_export, left? and no matter what it's selected, everything will be exported?

sunyi000: there's a yes/no param 'export all measurement' before this one , if user select 'yes', this data_to_export won't appear.

beatrizserrano: if the user selects 'no' in export all measurements, then needs to choose: `Experiment`, `Object relationship` or `Do not use`, left? The dynamically generated ones won't appear and users won't have the option of entering the name of the image/object that they want to get the features from. If I remember well, those 3 options available mean:
- `Experiment`: rather general metadata, such as number of images, ...
- `Object relationship`: object id - object id relationships. We needed this option to identify which nucleus contained which nucleoli for example.
- `Do not use`: no features computed I guess.

What I don't see is why someone would use any of this because:

- `Experiment`: no features from previous tools computed
- `Object relationship`: since dynamically generated ones don't appear, features associated to the objects being related here won't be computed.
- `Do not use`: means 'no' in export all measurements?

Now if someone wants to extract the features of just a segmentation mask (remember nuclei_mask for instance), the functionality is not there. 

My suggestion was: since we can't offer all the functionality, we just don't ask, we export everything. Wouldn't that mean that this whole select is not needed?

sunyi000: it's removed now..so export will always = all.
i'm just trying to preserve whatever options from Cellprofiler UI version in this Galaxy version, but yes if the options are not necessary here, we remove it..
cheers

beatrizserrano (approved):
thanks so much @sunyi000!

bgruening (approved):
Thanks both!

PR #1069, opened 2021-03-03
Add support for multiple input datasets.

davebx:


bgruening:
@davebx can we add any test with multiple input files? I guess supplying the same file twice could work?

davebx:
@bgruening sure, I can just split their test dataset into a few files, one moment

bgruening:
Ok, let the dance begin :)

bgruening:
Thanks @davebx!

PR #1071, opened 2021-03-04
add stress-ng

mtekman:
resumes https://github.com/galaxyproject/tools-iuc/pull/3492


mtekman:
should I rename it with underscores only?

bgruening:
@mtekman I just looked briefly over it, but could not see tests with memory? Maybe I overlooked it. What I'm after is to allocated XGB of memory, please make sure this works and we can merge this. Thanks.

mtekman:
the first test allocates 128MB of memory

there are actually two types of memory that can be allocated: "vm" memory and "memfd" memory, both which appear to take up RAM in my system

mtekman:
I can add more RAM tests if you like, I was basing the tests mostly off the documentation examples

bgruening:
Ok, we can start with that. Thanks.

> Output file did not contain expected text 'write128: interrupted early' (output 'stress-ng: info:  [9] apparmor: stressor will be skipped, AppArmor is not available

mtekman:
some examples:

`stress-ng --cpu 1 --cpu-load 5 --memfd 1 --memfd-bytes 4G`
`stress-ng --cpu 1 --cpu-load 5 --vm 1 --vm-bytes 4G`

(although the CPU load does not always seem to stay low)

bgruening:
CPU load is not interesting at the moment.

mtekman:
Yeah, but I was hoping we could keep the CPU usage low whilst the RAM was tested but this does not seem to reliably be the case

mtekman:
https://stackoverflow.com/questions/52107976/strees-ng-stress-memory-and-dont-want-to-let-cpu-loading-go-100

It looks like the `--cpu-load` option *does* work, but it still appears as using 100%

bgruening (approved):

bgruening:
Thanks. I will install this tool now and come back to you :)

tools/stress_ng/.shed.yml
```diff
@@ -0,0 +1,11 @@
+categories:
+  - Web Services
+description: stress test a computer system in various selectable ways
+long_description: |
+  stress-ng will stress test a computer system in various selectable
+  ways. It was designed to exercise various physical subsystems of a
+  computer as well as the various operating system kernel interfaces.
+owner: bgruening
```
mvdbeek, line 8: Could you maybe upload this as a different owner ? My concern in the IUC applies here as well, if you just install all tools by a trusted owner like iuc or bgruening you may end up with this tool on your server without intending to.

bgruening: I can do this if you like, but honestly, this "user" is active for 10 years and has a lot of repos that should not be installed, e.g. with old tool_dependencies etc ... we still do not install all IUC tools, but offer curated tool.yaml files.

mvdbeek: My thinking is that missing dependencies is different from a tool that could be used for somewhat malicious operations ... but I guess in the scenario of someone just installing everything there;d be many more problems anyway. I guess it's ok.

bgruening: Oh I hope that tool can not do anything malicious. They are still running in a cgroup context and the scheduler will kill it if they use too many resources. Do I miss something?

mvdbeek: But it can simulate many different things that are probably not controlled for by a scheduler, like IO, network and HDD activity, which I could see impacting other jobs on shared resources.

bgruening: Yes, this is true, but not different from Blast :sob:  ... on a positive note, this is exactly what I would like to test. How can we jail our jobs better and make our cgroup measurements better. ping @natefoo ... I wanted to eitherway inform you about this when we have installed the tool.

@mvdbeek let me know if I should move this to a different user, really happy to do it.

mvdbeek: It's what I would do, but I agree that you could find other tools do the same thing if you were intent on it.

bgruening: Ok, done.

PR #1077, opened 2021-03-12
Updating chemicaltoolbox/rdock tool version(s)

planemo-autoupdate:
Hello! This is an automated update of the following tool repo: **chemicaltoolbox/rdock**. I created this PR because I think the tool's main dependency is out of date, i.e. there is a newer version available through conda.

### DIFF
--- a/chemicaltoolbox/rdock/rdock_macros.xml
+++ b/chemicaltoolbox/rdock/rdock_macros.xml
@@ -1,5 +1,5 @@
 <macros>
-    <token name="@TOOL_VERSION@">2013.1-0</token>
+    <token name="@TOOL_VERSION@">2013.1</token>
     <xml name="citations">
         <citations>
             <citation type="doi">10.1371/journal.pcbi.1003571</citation>


chemicaltoolbox/rdock/rdock_macros.xml
```diff
@@ -1,5 +1,5 @@
 <macros>
-    <token name="@TOOL_VERSION@">2013.1-0</token>
+    <token name="@TOOL_VERSION@">2013.1</token>
```
bgruening, line 2: I'm not sure, if this number is higher than the previous one.

simonbray: Was the old version deleted? `conda search rdock` is only providing `2013.1`.

Anyway, getting the most recent version number is done with galaxy's conda_util, I didn't reimplement this.

bgruening: No clue how this could happen or ever worked. But this tool is either way deprecated left? Should we move it to deprecated?

PR #1081, opened 2021-03-12
Updating tools/diff tool version(s)

planemo-autoupdate:
Hello! This is an automated update of the following tool repo: **tools/diff**. I created this PR because I think the tool's main dependency is out of date, i.e. there is a newer version available through conda.

### DIFF
--- a/tools/diff/diff.xml
+++ b/tools/diff/diff.xml
@@ -1,12 +1,12 @@
 <tool id="diff" name="diff" version="@TOOL_VERSION@+galaxy@GALAXY_VERSION@">
     <description>analyzes two files and generates an unidiff text file with information about the differences and an optional Html report</description>
     <macros>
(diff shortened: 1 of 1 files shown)


bgruening:
this is your tool @davelopez  ;)

davelopez (approved):

simonbray:
I'm not sure if multiple versions of this tool are really useful... we can consider putting this on the autoupdate skiplist if you think so too.

davelopez:
Yeah, I don't think there will be many version changes either, actually for some reason I was using the previous diffutils version instead of the last, so the bot did a great job 😆 
You can add it to the skiplist if you like.

simonbray:
> this is your tool @davelopez ;)

Another thought: if we added a 'maintainer' in the .shed.yml file, we could have automatic pings. But I guess some IUC policy has to be changed for this again.

PR #1083, opened 2021-03-12
Include auxiliary files

gallardoalba:
This PR updates Salmon, including an option for generating auxiliary files.


bgruening:
What is the use-case of those files? They can't used inside of Galaxy isn't it?

Wouldn't it be better to tar them all up and just offer one giant tarball with all those files?

gallardoalba:
> What is the use-case of those files? They can't used inside of Galaxy isn't it?
> 
> Wouldn't it be better to tar them all up and just offer one giant tarball with all those files?

Some of them can be used in Multiqc (by including this module). At least some of there. I think that it is a good idea to include int a tar file those that cannot be used. 

https://github.com/ewels/MultiQC/blob/master/multiqc/modules/salmon/salmon.py

tools/salmon/macros.xml
```diff
@@ -259,6 +259,9 @@
             <param name="writeUnmappedNames" type="boolean" truevalue="--writeUnmappedNames" falsevalue="" checked="False"
             label="Write the names of un-mapped reads to the file unmapped_names.txt."
             help=""/>
+            <param name="auxiliary_files" type="boolean" truevalue="true" falsevalue="false" checked="false" 
+            label="Generate auxiliary files."
```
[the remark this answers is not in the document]
bgruening: The multiqc needed one for example seems more useful than the rest ...

tools/salmon/macros.xml
```diff
@@ -1,5 +1,5 @@
 <macros>
-    <token name="@VERSION@">1.3.0</token>
+    <token name="@VERSION@">1.4.0</token>
```
bgruening, line 2: @gallardoalba turns out the update is not that complicated. Most tests have passed.

tools/salmon/macros.xml
```diff
@@ -12,11 +12,48 @@
             <requirement type="package" version="1.4.1">scipy</requirement>
         </requirements>
     </xml>
+
+    <xml name="geometry">
```
bgruening, line 16: this is the only new parameter I would include ... can you work on this? Currently, this is not complete not is it actually used.

tools/salmon/macros.xml
```diff
@@ -12,11 +12,48 @@
...
+            ln -s "$geneMap" ./geneMap.${geneMap.ext} &&
+        #end if
+       
+        ]]>
+    </token>
+
+
     <xml name="stranded">
         <param name="strandedness" type="select" label="Specify the strandedness of the reads">
-                <option value="U" selected="True">Not stranded (U)</option>
```
bgruening, line 17: everything below is just my stupid editor and style changes, nothing important.

tools/salmon/salmonquant.xml
```diff
@@ -1,4 +1,4 @@
-<tool id="salmon" name="Salmon quant" version="@VERSION@+galaxy1">
+<tool id="salmon" name="Salmon quant" version="@VERSION@+galaxy2">
```
bgruening, line 1: we can reset this to galaxy0 with the new upgrade to version 1.4

tools/salmon/macros.xml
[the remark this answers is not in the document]
gallardoalba: Sure, I can do it. But in any case, the multiqc module requires all of them. But that's true that the ambig_info.tsv file provides interesting information even if not interested in multiqc analysis.

tools/salmon/macros.xml
gallardoalba: Sure!

tools/salmon/macros.xml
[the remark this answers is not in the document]
bgruening: All 3? Oo, what a mess. The MQC interface will look super complicated for salmon :(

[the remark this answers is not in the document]
gallardoalba: I think that those auxiliary files can be provided (at least the ambig_info.tsv) even if Salmon is not included in Multiqc; I consider it can provides interesting info.

PR #1088, opened 2021-03-13
Updating chemicaltoolbox/sucos from version 2020.03.4 to 2024.03.5

ghost:
Hello! This is an automated update of the following tool repo: **chemicaltoolbox/sucos**. I created this PR because I think the tool's main dependency is out of date, i.e. there is a newer version available through conda.

### DIFF
--- a/chemicaltoolbox/sucos/sucos.xml
+++ b/chemicaltoolbox/sucos/sucos.xml
@@ -2,7 +2,7 @@
     <description>- compare shape and feature overlap of docked ligand poses to a reference molecule</description>
     <macros>
         <import>sucos_macros.xml</import>
-        <token name="@GALAXY_VERSION@">1</token>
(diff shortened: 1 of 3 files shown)


bgruening:
@simonbray this looks good to me. I guess we should ignore the linting here, as this is upstream code from @tdudgeon ... correct?

Or should we ignore such tools as the real functionality (the python code shipped with this tool) is not updated?
@tdudgeon any change we can package sucos up into a pypi packge?

tdudgeon:
Personally I would not upgrade anything unless there is a major security issue with SciPy. There's a risk that something will break and no actual benefit in making the change.

@bgruening Putting the code into PyPi would take some work. What would be the benefit?

bgruening:
One advantage would be that this bot here would pick up any new version of your package and update the Galaxy tool as well.

PR #1095, opened 2021-03-22
Deprecate Diamond here, moved to IUC

abretaud:
Now that https://github.com/galaxyproject/tools-iuc/pull/3541 is merged


bgruening:
Nice!

PR #1100, opened 2021-04-06
Allow special characters as Create text file input

wm75:
This change allows arbitrary characters as tool input without
compromising safety against malicious text.

The added test revealed a bug in the existing tests, which only tested
against default settings.

The newly added "sanitize_text" macro may be useful for other tools here, too.

### DIFF
--- a/tools/text_processing/text_processing/macros.xml
+++ b/tools/text_processing/text_processing/macros.xml
@@ -16,6 +16,16 @@
             <regex match="Exception:" />
         </stdio>
     </xml>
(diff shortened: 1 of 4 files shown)


bgruening:
Do you want to add this sanitizer to the other tools as well?

wm75:
I'm planning to (that's why I put it into the macro file), but not in this PR.

Actually, the other tool I'd be interested in fixing soon is add_line_to_file, which lives a level higher in its own directory so using the same macro is not possible unfortunately.

PR #1106, opened 2021-04-13
Sucos fix

simonbray:


bgruening:
@simonbray please use `black` to fix the lining issues, otherwise the tool will not be pushed to the TS.

bgruening (approved):

simonbray:
Thanks @bgruening :+1:

PR #1107, opened 2021-04-14
Lint all python files

simonbray:
I guess a lot of tests will fail, not necessarily because of the linting.


simonbray:
Looks like the only tests which are failing here but passing in #1108 are `AREsite2_REST`, `antarna` and `ctb_chemfp_nxn_clustering`. (Tool linting also needs fixing)

bgruening:
What about merging this PR but stopping the CI. So those changes gets only deployed when there is a real tool update in the future?

simonbray:
> What about merging this PR but stopping the CI. So those changes gets only deployed when there is a real tool update in the future?

That sounds good to me, I just need to fix those failing tool tests as well :+1:

simonbray:
Okay, all the tests are passing that were passing before, so we can merge IMO.

bgruening:
Super cool, danke!

bernt-matthias:
Hi @simonbray in order to make this sustainable we should add this to the CI workflow: https://github.com/bgruening/galaxytools/blob/ca87db9c038a6fcf96aa39da50f384865fd932ff/.github/workflows/pr.yaml#L168 .. (or replace flake8)

ideally starting at the IUC repo.

simonbray:
I think the current `flake8` checks are fine. The problem was that most of the python code was failing linting because it was written before the CI checks were introduced. The aim of this PR was just to fix that.

PR #1109, opened 2021-04-20
Updating tools/rna_tools/miRDeep2/miRDeep2 tool version(s)

planemo-autoupdate:
Hello! This is an automated update of the following tool repo: **tools/rna_tools/miRDeep2/miRDeep2**. I created this PR because I think the tool's main dependency is out of date, i.e. there is a newer version available through conda.

### DIFF
--- a/tools/rna_tools/miRDeep2/miRDeep2/mirdeep2.xml
+++ b/tools/rna_tools/miRDeep2/miRDeep2/mirdeep2.xml
@@ -1,6 +1,6 @@
 <tool id="rbc_mirdeep2" name="MiRDeep2" version="@TOOL_VERSION@+galaxy@GALAXY_VERSION@">
     <macros>
-        <token name="@TOOL_VERSION@">2.0.1.2</token>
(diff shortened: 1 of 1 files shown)


simonbray (approved):

simonbray:
@bgruening this is an interesting example of how the Galaxy tests catch a problem which the tests in the conda recipe miss.

(see https://github.com/bioconda/bioconda-recipes/pull/29583)

bgruening:
Oh, fantastic detective work! Thanks for the fix. I'm wondering if we can improve all this testing and the "deployment pipeline" but I don't know how ...

simonbray:
It might be too much, but you could in theory run the Galaxy tool tests in the bioconda CI. We already annotated some of the bioconda recipes with the corresponding Galaxy tools, which would be the necessary first step. It would make the bioconda testing much more robust.

I don't know if it is worth investing time into this though.

bgruening:
I don't think the bioconda folks would like this overhead. We could maybe make it easier to use the prebuild container from the bot? Or trigger a Galaxy update from bioconda?

PR #1116, opened 2021-05-14
Add bionano

gallardoalba:
This PR adds the Bionano Hybrid Scaffold tool.


tools/bionano/.shed.yml
```diff
@@ -0,0 +1,9 @@
+categories:
+- Assembly
```
bgruening, line 2: Cris can you make a suite out of this please, in case we need to add more tools.

gallardoalba: Sure!

tools/bionano/bionano_scaffold_edited.xml
```diff
@@ -91,7 +99,7 @@
         <flag attr="hashgen" val0="5" val1="7" val2="2.4" val3="1.5" val4="0.05" val5="5.0" val6="1" val7="1" val8="4"/>
         <flag attr="hash" val0="-hashdelta" val1="26" val2="10" val3="46"/>
         <flag attr="hashMultiMatch" val0="30" val1="10"/>
-        <flag attr="insertThreads" val0="$cores"/>
+        <flag attr="insertThreads" val0="54"/>
```
bgruening, line 102: why do you revert this?

tools/bionano/bionano_scaffold_edited.xml
```diff
@@ -65,7 +73,7 @@
         <flag attr="RAmem" val0="3" val1="1"/>
     </global>
     <fasta2cmap>
-        <flag attr="enzyme" val0="BspQI" display="Enzyme" group="FASTA to CMAP digestion" description="Define single enzyme for in-silico FASTA to CMAP digestion. Avalible enzymes: BspQI, BbvCI, BsmI, BsrDI, BssSI, DLE1."/>
+        <flag attr="enzyme" val0="CTTAAG" display="Enzyme" group="FASTA to CMAP digestion" description="Define single enzyme for in-silico FASTA to CMAP digestion. Avalible enzymes: BspQI, BbvCI, BsmI, BsrDI, BssSI, DLE1."/>
```
bgruening, line 76: Why is this changed?

tools/bionano/bionano_scaffold_edited.xml
```diff
@@ -178,180 +186,192 @@
...
+        <flag attr="outlier" val0="0.0001"/>
+        <flag attr="endoutlier" val0="0.001"/>
+        <flag attr="PVendoutlier"/>
+        <flag attr="deltaX" val0="12"/>
+        <flag attr="deltaY" val0="12"/>
+        <flag attr="xmapchim" val0="12"/>
+        <flag attr="hashgen" val0="5" val1="7" val2="2.4" val3="1.5" val4="0.05" val5="5.0" val6="1" val7="1" val8="4"/>
+        <flag attr="hash" val0="-hashdelta" val1="26" val2="10" val3="46"/>
+        <flag attr="hashMultiMatch" val0="30" val1="10"/>
+        <flag attr="insertThreads" val0="54"/>
```
bgruening, line 203: why this? we should use $cores

tools/bionano/bionano_scaffold_edited.xml
```diff
@@ -420,6 +456,15 @@
                     <has_archive_member path=".*/status.txt"/>
                 </assert_contents>
             </output>
+                        <assert_stdout>
```
bgruening, line 459: indentation? Are you not using the galaxy-language-server?

tools/bionano/bionano_scaffold_edited.xml
```diff
@@ -0,0 +1,524 @@
...
+        #if $configuration_options.configuration == 'vgp'
+            && cp '${vgp_mode}' ./config.xml
+        #else
+            #set cores = os.environ.get('GALAXY_SLOTS', 2)
+            #set memory = os.environ.get('GALAXY_MEMORY_MB', 4000) / 1000
+            && cp '${configuration_file}' ./config.xml
+            && sed -i 's|attr="maxmem" val0=.* display|attr="maxmem" val0="$memory" display|' ./config.xml
+            && sed -i 's|attr="maxthreads" val0=.* display|attr="maxthreads" val0="$cores" display|' ./config.xml
+            && sed -i 's|attr="maxvirtmem" val0=.*/>|attr="maxvirtmem" val0="$cores"/>|' ./config.xml
+            && cat ./config.xml
```
bgruening, line 23: this cat should happen in both cases.... and then we can have tests that check that the core settings are correct

tools/bionano/bionano_scaffold_edited.xml
gallardoalba: Yes, sorry, It was a fail. I changed all the configuration file because the one provided in the test correspond to the samples used on it. I'll fix it.

gallardoalba: I didn't realize that there were more parameters that required to be modified.

bgruening: There are a few more ...

tools/bionano/bionano_scaffold_edited.xml
gallardoalba: Because the VGP configuration uses that sequence (corresponding to a specific restriction enzyme); the input files used in the test uses a different one.

tools/bionano/bionano_scaffold_edited.xml
gallardoalba: Yes, I'll fix it.

tools/bionano/bionano_scaffold_edited.xml
gallardoalba: The problem is that the VGP configuration file cannot be used with the input files used in the tests because the restriction enzymes are different.

tools/bionano/bionano_scaffold_edited.xml
bgruening: ok, maybe that parameter should be made configureable?

gallardoalba: I think that It wouldn't be necessary (except for testing purposes). But if you consider so I'll include it.

tools/bionano/bionano_scaffold_edited.xml
gallardoalba: Yes, I just copied this block from the one which you included.

tools/bionano/bionano_scaffold_edited.xml
```diff
@@ -76,7 +75,7 @@
         <flag attr="RAmem" val0="3" val1="1"/>
     </global>
     <fasta2cmap>
-        <flag attr="enzyme" val0="CTTAAG" display="Enzyme" group="FASTA to CMAP digestion" description="Define single enzyme for in-silico FASTA to CMAP digestion. Avalible enzymes: BspQI, BbvCI, BsmI, BsrDI, BssSI, DLE1."/>
+        <flag attr="enzyme" val0="${configuration_options.enzyme}" display="Enzyme" group="FASTA to CMAP digestion" description="Define single enzyme for in-silico FASTA to CMAP digestion. Avalible enzymes: BspQI, BbvCI, BsmI, BsrDI, BssSI, DLE1."/>
```
gallardoalba, line 78: This is not possible because the $configuration_options.enzyme is only set when  $configuration_options == 'VGP'

bgruening: Than add an if-condition.

tools/bionano/bionano_scaffold_edited.xml
```diff
@@ -0,0 +1,614 @@
...
+            <param name="conflict_filter_genome" value="3"/>
+            <param name="conflict_filter_sequence" value="3"/>
+            <conditional name="configuration_options">
+                <param name="configuration" value="file"/>
+                <param name="configuration_file" value="configuration.xml"/>
+            </conditional>
+            <param name="zip_file" value="true"/>
+            <output name="ngs_contigs_scaffold_fasta" ftype="fasta">
+                <assert_contents>
+                    <has_size value="4753369" delta="300" />
```
bgruening, line 445: you can test a FASTA file in a smarter way than size. E.g you can add line count, assert some header line etc in addition.

tools/bionano/bionano_scaffold_edited.xml
```diff
@@ -0,0 +1,614 @@
...
+            <param name="bionano_cmap" value="colormap_assembly.cmap"/>
+            <param name="conflict_filter_genome" value="2"/>
+            <param name="conflict_filter_sequence" value="2"/>
+            <conditional name="configuration_options">
+                <param name="configuration" value="file"/>
+                <param name="configuration_file" value="configuration.xml"/>
+            </conditional>
+            <output name="ngs_contigs_scaffold_fasta" ftype="fasta">
+                <assert_contents>
+                    <has_size value="4753369" delta="300" />
```
bgruening, line 484: same here

bgruening:
Cool, do you add this to EU?

PR #1121, opened 2021-05-24
Add MitoHiFi

gallardoalba:


tools/mitohifi/macros.xml
```diff
@@ -0,0 +1,22 @@
+<macros>
+    <token name="@TOOL_VERSION@">2</token>
+    <token name="@GALAXY_TOOL_VERSION@">galaxy0</token>
+    <xml name="edam_ontology">
+        <edam_topics>                                                                                  
+            <edam_topic>topic_0196</edam_topic>
```
bgruening, line 6: do you check thos topics?

tools/mitohifi/mitohifi.xml
```diff
@@ -0,0 +1,349 @@
+<tool id='mitohifi' name='MitoHiFi' version='@TOOL_VERSION@+@GALAXY_TOOL_VERSION@' profile='20.01'>
+    <description>assembly mitogenomes from Pacbio HiFi reads</description>
+    <macros>
+        <import>macros.xml</import>
+    </macros>
+    <expand macro='edam_ontology' />
+    <expand macro='requirements' />
+    <version_command>python mitohifi_v2.py -h</version_command>
```
bgruening, line 8: This is not returning any version information.

tools/mitohifi/mitohifi.xml
```diff
@@ -0,0 +1,349 @@
...
+            #if $operation_mode.min_length
+                --min_length $operation_mode.min_length
+            #end if
+            $operation_mode.exact_specie
+            && mv ./*fasta ./reference_genome.fasta
+            && mv ./*gb ./reference_genome.gb
+        #elif $operation_mode.command == 'mitohifi'
+            mitohifi_v2.py 
+            #if $operation_mode.input_option.input == 'pacbio'
+                -r '${ ' '.join(["%s" % $x for $x in $operation_mode.input_option.input_reads]) }'
```
bgruening, line 24: single-quotes, I think you generate here double-quotes

tools/mitohifi/mitohifi.xml
```diff
@@ -0,0 +1,349 @@
...
+    ]]>    </command>
+    <inputs>
+        <conditional name="operation_mode">
+            <param name="command" type="select" label="Operation type selector" help="Select a type of operation">
+                <option value="find_reference">Find a close-related mitochondrial reference genome</option>
+                <option value="mitohifi">Run MitoHiFi</option>
+            </param>
+            <when value="find_reference">
+                <param argument="--species" type="text" value="" label="Specie name" help="Specie name in binomial nomenclature">
+                    <sanitizer invalid_char="">
```
bgruening, line 52: @gallardoalba general question, those should be validators not sanitizers, is that correct? https://docs.galaxyproject.org/en/latest/dev/schema.html#tool-inputs-param-validator

tools/mitohifi/mitohifi.xml
```diff
@@ -0,0 +1,349 @@
...
+                        <valid initial="string.letters,string.digits">
+                            <add value=" "/>
+                        </valid>
+                    </sanitizer>
+                </param>
+                <param argument="--email" type="text" value="" label="Email" help="Required for NCBI database query">
+                    <sanitizer invalid_char="">
```
bgruening, line 59: validator?

bgruening (commented):
Only a few minor comments. Thanks Cris.

tools/mitohifi/macros.xml
gallardoalba: Yes, assembly https://bioportal.bioontology.org/ontologies/EDAM?p=classes&conceptid=topic_0196

tools/mitohifi/mitohifi.xml
gallardoalba: It is fine I think, this is the output command that it generates `mitohifi_v2.py -r '/tmp/tmp_brg1r0y/files/000/dataset_91.dat /tmp/tmp_brg1r0y/files/000/dataset_92.dat' -f '/tmp/tmp_brg1r0y/files/000/dataset_3.dat' -g '/tmp/tmp_brg1r0y/files/000/dataset_2.dat' -p 50 -o 5 -t ${GALAXY_SLOTS:-1} `

tools/mitohifi/mitohifi.xml
gallardoalba: Fixed

tools/mitohifi/mitohifi.xml
gallardoalba: I try to restrict the use of the wrong parameters which could be dangerous (hacking?).  Then I think that I should use sanitize, but not sure about it.

tools/mitohifi/mitohifi.xml
gallardoalba: Same as before, probably validators would be useful too.

bgruening (approved):

bgruening:
Nice! Thanks @gallardoalba ... please add this tool to EU and also to the tool_destinations (we need to give it a docker destination).

PR #1125, opened 2021-05-28
Bionano modify outputs

gallardoalba:
This PR updates the outputs; scaffolded and not scaffolded files are provided independently instead of concatenating them. It also provides the conflict.txt file as output.

### DIFF
--- a/tools/bionano/bionano_scaffold.xml
+++ b/tools/bionano/bionano_scaffold.xml
@@ -55,7 +55,6 @@
         -f
         $zip_file
         -o ./
-        && cat $output_file_NCBI $output_file_not_scaffolded > total_contigs_raw.fasta
         #if $trim_cut_sites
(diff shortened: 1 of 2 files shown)


tools/bionano/bionano_scaffold.xml
```diff
@@ -420,22 +420,26 @@
         -->
     </inputs>
     <outputs>
-        <data name="ngs_contigs" format="fasta" from_work_dir="total_contigs_raw.fasta" label="${tool.name} on ${on_string}: NGScontigs">
```
bgruening, line 423: @Delphine-L is that everything that you need?

bgruening (commented):
Cool thanks @gallardoalba!

tools/bionano/bionano_scaffold.xml
gallardoalba: Yes, I think so, those are the required files that she told me.

PR #1135, opened 2021-06-22
changes the way files are concating, now appending files one by one instead of altogether

anilthanki:
changes the way files are concating, now appending files one by one instead of altogether to avoid large command when a collection is too big

### DIFF
--- a/tools/text_processing/text_processing/cat.xml
+++ b/tools/text_processing/text_processing/cat.xml
@@ -11,12 +11,15 @@
     </version_command>
     <command>
     <![CDATA[
-        cat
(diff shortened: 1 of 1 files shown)


anilthanki:
Hi @bgruening 

The error is for another tool called  `cut` not from my PR

Thanks

Anil

bgruening:
Thanks @anilthanki!

nsoranzo:
Thanks for the manual upload @bgruening !

anilthanki:
Thanks, @bgruening for merging this quickly..

PR #1118, opened 2021-05-18
Made asm parameter optional in flye

Slugger70:
Addresses issue #1117

Issue #1117, opened 2021-05-18
Flye versions > 2.6 are listed in the toolshed as not installable.


Slugger70:
The toolshed lists the error as `parameter 'asm': the attribute 'value' must be set for non optional parameters`

PR #1137, opened 2021-06-30
Update Flye 3.8.3

gallardoalba:
Thus PR updates Flye up to version 3.8.3. It includes the following modfiications:

- A new mode: PacBio HiFi
- New option: Expected HiFi reads error rate (--hifi-error)
- New option: Keep haplotypes (--keep-haplotypes)
- Fix bug https://github.com/bgruening/galaxytools/pull/1118/files
- Include additional information in the help option


bgruening:
I see that here: [2021-06-30 02:29:07] ERROR: No disjointigs were assembled - please check if the read type and genome size parameters are correct

tools/flye/flye.xml
```diff
@@ -1,146 +1,390 @@
...
         <conditional name="asm">
-            <param name="asm_select" type="select" label="description" help="">
+            <param name="asm_select" type="select" label="Reduced contig assembly coverage">
                 <option value="true">Enable reduced coverage for initial disjointing assembly</option>
                 <option value="false" selected="true">Disable reduced coverage for initial disjointing assembly</option>
             </param>
             <when value="true">
-                <param name="asm" argument="--asm-coverage" type="integer" optional="true" label="reduced coverage for initial disjointing assembly" />
-                <param argument="-g" type="text" label="estimated genome size (for example, 5m or 2.6g)">
+                <param argument="--asm-coverage" type="integer" min="0" value="30"
```
bgruening, line 84: According to this issue https://github.com/bgruening/galaxytools/issues/1117

the param needs either a `value` or needs to be optional.

gallardoalba: I used that value because of that _For a typical assembly, 30x is enough to produce good initial contigs. Regardless of this parameter, all reads will be used at the later pipeline stages_ . Do you think it should be optional?

bgruening: No, that is fine. You explanation makes sense.

PR #1138, opened 2021-07-13
update the main module cellprofiler to version 4.2.0

sunyi000:
- updated the tests in main module cellprofiler.xml to work with latest 4.2.0
- updated version in macros.xml
- tested backward compatibility with two version 3 workflow, obtained from workflow hub and 'object tracking' training tutorial


bgruening:
@sunyi000 this will also update all the other modules to the 4.2 version, is that the idea? So the training and Bea's workflow with the many modules are working with the new 4.2 version? That would be awesome!

sunyi000:
> @sunyi000 this will also update all the other modules to the 4.2 version, is that the idea? So the training and Bea's workflow with the many modules are working with the new 4.2 version? That would be awesome!


I tested her training workflow with the updated cellprofiler.xml module, it seems the backward compatibility of 4.2.0 is pretty good, it works well with old pipeline format..

or to be safe, should we create another separate cp4 tool, and put a deprecation warning to the cp3 tool?

bgruening:
We can leave it like that, but are the modules are also working? I mean have you tested the entire workflow with all the modules or only the one cellprofiler.xml  tool? It would be super cool if the modules would also work. I just thought that this will not be the case as everything has changed a lot.

bgruening:
grafik

tests are failing :(

PR #1139, opened 2021-07-19
Update salmon

mvdbeek:
We've got an invitation to submit an example galaxy workflow that uses salmon for a review paper, so it would be good to support the latest release.
I've added the `--ont` flag, but 1.4 also has new features for alevin that I haven't implemented. Not sure if these are helpful without alevin-fry ?


mvdbeek:
@bgruening it's green now, can you take a look please ?

bgruening (approved):

bgruening:
Thanks @mvdbeek!

mvdbeek:
Thank you @bgruening!

PR #1143, opened 2021-07-29
Add tool-data/all_fasta.loc.sample to nanopolish directory because nanopolish tools are marked as invalid in the toolshed

cat-bro:
The toolshed has marked a number of nanopolish tools as invalid because there is no sample file all_fasta.loc.  This change adds that and also renames test-data/all_fasta.loc.test to test-data/all_fasta.loc to be consistent with its path in tool_data_table_conf.xml.test.

This change doesn't affect to tool tests, which always refer to the test .fa files by name rather than going through the data tables.


bgruening (commented):
I can be wrong but I think we need the tools/nanopolish/test-data/all_fasta.loc.test file for the test.

cat-bro:
Hi @bgruening, thanks for having a look.  The CI was stuck so I've re-pushed the commits to make it restart.

The test data table refers to test-data/all_fasta.loc, not test-data/all_fasta.loc.test.  I don't know whether or not .test would be a recognised pattern here.  The tests don't use the data tables at all, they use reference data from_history.

https://github.com/cat-bro/galaxytools/blob/master/tools/nanopolish/tool_data_table_conf.xml.test
```
<tables>
    <!-- Locations of all fasta files under genome directory -->
    <table name="all_fasta" comment_char="#" allow_duplicate_entries="False">
        <columns>value, dbkey, name, path</columns>
        <file path="${__HERE__}/test-data/all_fasta.loc" />
    </table>
</tables>
```

I could drop the renaming commit.  The issue in the toolshed is explicitly about the absence of all_fasta.loc.sample and the name of test-data/all_fasta.loc* probably does not matter.

bgruening:
Ok that makes sense. Thanks.

PR #1144, opened 2021-08-03
Update sklearn citations

qiagu:
Add galaxy-ml paper into citations.

### DIFF
--- a/tools/sklearn/main_macros.xml
+++ b/tools/sklearn/main_macros.xml
@@ -1922,10 +1922,28 @@
     <xml name="sklearn_citation">
         <citations>
             <citation type="bibtex">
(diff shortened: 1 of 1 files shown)


tools/sklearn/main_macros.xml
```diff
@@ -1922,10 +1922,28 @@
     <xml name="sklearn_citation">
         <citations>
             <citation type="bibtex">
-          @article{scikit-learn, title={Scikit-learn: Machine Learning in {P}ython}, author={Pedregosa, F. and Varoquaux, G. and Gramfort, A. and Michel, V.
```
bgruening, line 1925: can you please use DOIs instead?

qiagu: Is there DOI for scikit-learn?

bgruening: No but ours :-)

PR #1149, opened 2021-08-29
racon: fix file extensions

bernt-matthias:
must be fastq/fasta but they aren't for fastqsanger ...

fixes https://github.com/galaxyproject/tools-iuc/issues/3856


bgruening:
Very cool, thanks.

PR #1157, opened 2021-09-10
Save datasets in Galaxy history from jupyterlab notebook 

anuprulez:
This PR adds a Galaxy tool for processing long-running jobs emitted from a script in the Jupyterlab notebook such as created [here](https://github.com/galaxyproject/galaxy/pull/11484). This tool will be useful for running scripts as a job such as training a machine/deep learning model that takes a long time to finish. The trained model (as ONNX format) will be available in Galaxy history after the job finishes. 

Steps to run this tool:

1. Serve Galaxy via planemo locally containing this tool
2. Generate an API key from running Galaxy instance
3. Execute the notebook https://github.com/anuprulez/jupyterlabtool/blob/master/upload_jltools.ipynb after updating the API key. This notebook will connect to the local Galaxy instance via bioblend and execute the saved script on Galaxy. Once the job finishes, the trained model will be available in a newly created Galaxy history as an ONNX file.
4. Note: ONNX file format is now available in Galaxy in the dev branch.

Currently, this work is in progress, tools tests are not yet written. Your comments are welcome!! Thanks!

tool_output


mvdbeek:
What do you think about using a collection with `discover_outputs`, this should have a similar effect and be better performance-wise, it would work better in workflows, history extractions etc ?

tools/jupyter_job/run_jupyter_job.xml
```diff
@@ -0,0 +1,80 @@
+<tool id="run_jupyter_job" name="Run long running jupyterlab job" version="0.0.1">
```
bgruening, line 1: @anuprulez can you check if there exists a hidden=true option here?

tools/jupyter_job/run_jupyter_job.xml
```diff
@@ -0,0 +1,80 @@
...
+    <description>on GPUs</description>
+    <requirements>
```
bgruening, line 3: I thought this needs to run in a Docker container? We should make this explicit, as we don't want to run this in conda, for security reasons?

tools/jupyter_job/run_jupyter_job.xml
```diff
@@ -0,0 +1,80 @@
...
+    </command>
+    <inputs>
+        <param name="select_file" type="data" label="Load file" format="txt"/>
+    </inputs>
+    <outputs>
+        <data format="onnx" name="outfile_output_model" label="Saved model"></data>
+        <data format="h5" name="outfile_output_arrays" label="Saved arrays"></data>
+    </outputs>
+    <tests>
+        <test>
```
bgruening, line 29: can you a test for num expected outputs?

tools/jupyter_job/run_jupyter_job.xml
```diff
@@ -0,0 +1,80 @@
...
+            <param name="select_file" value="tf-script.py"/>
+            <output name="outfile_output_model" file="tf-script-model.onnx" ftype="onnx" compare="sim_size" delta="100" />
+        </test>
+        <test>
+            <param name="select_file" value="scikit-script.py"/>
+            <output name="outfile_output_model" file="scikit-script-model.onnx" ftype="onnx" compare="sim_size" delta="50" />
+        </test>
+        <test>
+            <param name="select_file" value="tf-script.py"/>
```
bgruening, line 38: why not combining this test with the first one?

tools/jupyter_job/run_jupyter_job.xml
```diff
@@ -0,0 +1,80 @@
...
+            <output name="outfile_output_arrays">
+                <assert_contents>
+                    <has_h5_keys keys="loss_history,mnist_images,mnist_labels,tot_loss"/>
+                </assert_contents>
+            </output>
+        </test>
+        <test>
+            <param name="select_file" value="scikit-script.py"/>
```
bgruening, line 46: and this test with the second?

tools/jupyter_job/run_jupyter_job.xml
bgruening: please also add a recent profile=21.05

tools/jupyter_job/run_jupyter_job.xml
bgruening: ping @anuprulez

anuprulez: This dynamic code execution is the same as executing code written in any other interactive tool. The execution command from this tool can be sent to any secluded VM for running as far as I understand.

bgruening: Yes, but this tool is way more insecure than any other tool, left? So making sure it runs in Docker and not by accident in Conda is making it a little bit more secure?

anuprulez: I agree, having a docker container running this tool will automatically enable it to run in a secluded environment.

mvdbeek: Moving to an explicit container requirement does not in any way guarantee this is safe. The right way *right now* is to make it an interactive tool. The medium term solution is a tag that means we require containerized execution and pick a destination that can run it.

anuprulez: Current tool is supposed to be executed on jupyterlab notebook that is already an interactive tool. I am not sure if interactive tools interact with one another?

anuprulez:
> What do you think about using a collection with `discover_outputs`, this should have a similar effect and be better performance-wise, it would work better in workflows, history extractions etc ?

This tool is very specific that executes Python code having ML scripts, extracted from Jupyterlab ML interactive notebook, dynamically and saves the trained models, numpy arrays, lists, dataframes. I am not sure if we have such a use case involving workflows or history extractions.

tools/jupyter_job/run_jupyter_job.xml
mvdbeek: No problem, they're just tools.

mvdbeek:
> This tool is very specific that executes Python code having ML scripts, extracted from Jupyterlab ML interactive notebook, dynamically and saves the trained models, numpy arrays, lists, dataframes. I am not sure if we have such a use case involving workflows or history extractions.

These are the advantages from the user-perspective, the main point for admins and someone who works on Galaxy is that you don't need to bypass Galaxy and use an API key. That has a bunch of issues that should really make this a last-resort.

anuprulez:
Steps to run this tool (after this is merged to Galaxy EU):

1. Open Galaxy's interactive tool (https://live.usegalaxy.eu/?tool_id=interactive_tool_ml_jupyter_notebook)
2. Write a script as `.ipynb` file (or copy the entire script from: https://github.com/anuprulez/galaxytools/blob/run_job/tools/jupyter_job/test-data/tf-script.py)
3. Open another `.ipynb` tab and write and execute the following script:
```
url = <<Galaxy's URL>> e.g. https://usegalaxy.eu/
key = <<your Galaxy's API key>>
file_path = <<ipynb script relative path with respect to the notebook>>
tool_output = run_script_job(file_path)
```

bgruening (approved):
nice, thanks a lot @anuprulez

PR #1161, opened 2021-09-12
Cmfinder fix

pavanvidem:
graphclust param type to text


bgruening:
tests are failing ...

PR #1162, opened 2021-09-13
Update Peakachu v0.2.0

gallardoalba:
This PR updates Peakachu up to version 0.2.0 and fixes a but in the _--size_factors_ parameter.


bgruening:
@gallardoalba thanks. Can you please add a test next time. But this is great!

gallardoalba:
Sure, I tried, but didn't know which input was required in that case.

PR #1163, opened 2021-09-15
add cellprofiler v4.2.1 tool

sunyi000:
- main module of cellprofiler 4.2.1 as galaxy tool


sunyi000:
@bgruening I still get all tests passed on my local machine with clean setup, but failed during CI.
- the only difference I see is that the ci workflow uses py3.7 during test and combined_test..
- since cellprofiler only works with 3.8 and above (https://github.com/CellProfiler/CellProfiler/blob/master/setup.py#L96) I just want to try upgrading the python version in ci.yaml but it seems that circleci cached the old version somewhere.
- could you help me clear the cache, or do I simply just need to rename the cache key in ci.yml

just want to see if changing python version in ci.yml makes any difference

thanks very much

bgruening:
Please run your tests locally with `planemo test --docker`. Are they failing when you do that?

sunyi000:
> Please run your tests locally with `planemo test --docker`. Are they failing when you do that?

it succeeded. all 4 tests passed with `planemo test --docker cellprofiler.xml`

tools/cellprofiler_v4/cellprofiler.xml
```diff
@@ -0,0 +1,244 @@
...
+            <assert_contents>
+                <has_n_lines n="@N@" />
+            </assert_contents>
+        </xml>
+    </macros>
+    <expand macro="cp_requirements" />
+    <command detect_errors="aggressive"><![CDATA[
+        bash '$script_file' &&
+        cellprofiler -c -r --file-list list.txt -o ./output -p '$pipeline' 
+        #if '$detailed_output'
```
gallardoalba, line 16:
```suggestion
        #if $detailed_output
```

gallardoalba: I tried to run the test, but all of them failed. I got this error:

Screenshot from 2021-09-22 16-41-52

sunyi000: Thanks  @gallardoalba 
I couldn't find where is this /home/laptop directory, to be honest.
I tested it with planemo and the results are green. Could you please point me to the right direction?
Thanks very much

sunyi000: but if you look at the output of the CI in github, the actual error is that during CI process, the tests in xml failed because cellprofiler doesn't produce any output files. 

My guess is that this probably caused by the python version in ci.yml, as it is 3.7 in CI but cellprofiler 4 only works with 3.8.

I tested it locally with clean env, both using `planemo t cellprofiler.xml` and `planemo t --docker cellprofiler.xml` all get green test results. so Im out of ideas on this :)


ubuntu@galaxytools-dev_ ~_galaxytools_tools_cellprofiler_v4 22_09_2021 17_39_17

sunyi000: now I saw the same /home/runer/.cache permission error in Github CI output...
hmm...

tools/cellprofiler_v4/cellprofiler.xml
```diff
@@ -0,0 +1,244 @@
...
+        <token name="@VERSION_SUFFIX@">0</token>
+        <xml name="test_assert_content" token_n="291">
+            <assert_contents>
+                <has_n_lines n="@N@" />
+            </assert_contents>
+        </xml>
+    </macros>
+    <expand macro="cp_requirements" />
+    <command detect_errors="aggressive"><![CDATA[
+        bash '$script_file' &&
```
gallardoalba, line 14:
```suggestion
        export MPLCONFIGDIR=\$TEMP &&
        bash '$script_file' &&
```

gallardoalba:
When I ran it in local, I get this error:

Screenshot from 2021-09-22 18-22-32

But it works after downgrading the NumPy version:

Screenshot from 2021-09-22 18-23-24

Then I guess that it is necessary to modify the conda recipe.

tools/cellprofiler_v4/cellprofiler.xml
gallardoalba: It fixes the matplotlib problem, but there is still a problem with some dependencies.

sunyi000:
> When I run it in local, I get this error:
> 
> Screenshot from 2021-09-22 18-22-32
> 
> But it works after downgrading the NumPy version:
> 
> Screenshot from 2021-09-22 18-23-24
> 
> Then I guess that it is necessary to modify the conda recipe.

ah.. I got numpy 1.19 in my environment..I will go change the recipe then...
thanks

sunyi000:
I just updated the recipe, and created a PR #30645 
@gallardoalba Thank you.

gallardoalba:
> I just updated the recipe, and created a PR   [30645](https://github.com/bioconda/bioconda-recipes/pull/30645) 
> @gallardoalba Thank you.

You can ask here for a review https://gitter.im/bioconda/Lobby

simonbray:
Could you rename the `@CP_VERSION@` token to `@TOOL_VERSION@`? This will allow automated updates of the cellprofiler tool. :)

tools/cellprofiler_v4/cellprofiler.xml
```diff
@@ -0,0 +1,245 @@
+<tool id="cp_cellprofiler4" name="CellProfiler" version="@TOOL_VERSION@+galaxy@VERSION_SUFFIX@">
```
simonbray, line 1: Is there a reason here to create a completely new wrapper rather than updating the existing tool?

sunyi000: cellprofiler 3 tools has other modules only works with version3. Those module in cp3 suite will not work with cp4.

simonbray: I don't think that is a reason to create a new wrapper. It looks like the function of the tool is the same as the old one, even if the versions are not compatible with each other.

Actually this will create more confusion for the user, because they will search for the tool in Galaxy and see two options, and it won't be immediately clear which to use.

gallardoalba:
This is the error @sunyi000 https://github.com/scikit-image/scikit-image/issues/4664

sunyi000:
thanks for the link.. I will create a docker
 but I'm still confused why this is working in a local env...

gallardoalba:
> thanks for the link.. I will create a docker
> but I'm still confused why this is working in a local env...

In that case, the reason is that in local don't exist such problems with folder permissions, but in Galaxy, due to security reasons, there are restrictions. Regards.

gallardoalba:
> I was looking for it in the toolshed and realized it wasn't there yet. It looks like all checks are fine. Is there any other thing that's holding up the merge?

Done!

jkh1:
Thanks a lot.

PR #1181, opened 2021-12-03
Suppress report printing to standard out in case of UnicodeEncodeError

dnbenso:
Report contents are now printed to stdout and after trim_galore completes sqlalchemy in the Galaxy venv may complain about latin-1 codec. This is because "micro seconds" for example are written with the standard Greek mu symbol.  Users who desire this output can still get it by selecting to print the report and Galaxy may render the character with '\ufffd'.

### DIFF
--- a/tools/trim_galore/trim_galore.xml
+++ b/tools/trim_galore/trim_galore.xml
@@ -52,6 +52,9 @@
 
         trim_galore
 
+        ## prevent report from being printed to stdout as it may contain greek chars like mu
+        --suppress_warn
+
         ## according the develpers 4 cores could be a sweet spot, anything above has diminishing returns
         --cores \${GALAXY_SLOTS:-4}
 


bgruening:
@dnbenso this should be fixed on the Galaxy side I think, isn't it?

dnbenso:
@bgruening this tool does seem to work on usegalaxy.eu. But on my release_21.05 it doesn't.  I get a traceback starting with:

  File "lib/galaxy/jobs/runners/__init__.py", line 535, in _finish_or_resubmit_job

and ending with:
UnicodeEncodeError: 'latin-1' codec can't encode character '\ufffd' in position 2420: ordinal not in range(256)

If it is just something peculiar with my implementation of Galaxy we can discard the change.

bgruening:
I think so. Its some DB setting, we had this issue as well I remember. Just not the fix at the moment. Maybe you can ask the admins what this can be? You probably just need to change the encoding to UTF-8 somewhere.

dnbenso:
Thanks @bgruening.  I'll have another look at the UTF-8 encoding and database..  Cheers.

PR #1183, opened 2022-01-17
update bionano_scaffold tests so that they do not fail based on values of 'threads' and 'memory'

cat-bro:
The tests contain assertions that stdout contains the exact text `attr="maxmem" val0="8"` for all test jobs.  Quite often this value will not be 8.  Update this and other values to assert that it matches an integer, not a specific integer.


bgruening (approved):

bgruening:
Merci!

PR #1187, opened 2022-01-27
add pharmcat

astrovsky01:


tools/pharmcat/pharmcat.xml
```diff
@@ -0,0 +1,80 @@
+<tool id="pharmcat" name="pharmcat" version="@WRAPPER_VERSION@+@VERSION_SUFFIX@" profile="20.01">
+    <macros>
+        <token name="@WRAPPER_VERSION@">1.3.1</token>
+        <token name="@VERSION_SUFFIX@">galaxy0</token>
+    </macros>
+    <requirements>
+        <requirement type="package" version="@WRAPPER_VERSION@">pharmcat</requirement>
```
bgruening, line 7: Since there is no conda package the container can not be built. So we need something like that: https://github.com/bgruening/galaxytools/blob/f69dd59785c9674ebd58506d5d8b31289d7e9532/tools/bionano/macros.xml#L26

tools/pharmcat/pharmcat.xml
```diff
@@ -0,0 +1,80 @@
...
+    </requirements>
+    <command detect_errors="exit_code"><![CDATA[
+        pharmcat
+        -vcf $input
```
bgruening, line 11: single quotes

tools/pharmcat/pharmcat.xml
```diff
@@ -0,0 +1,80 @@
...
+            -f output.json
+            #if $function_select.outside_call:
+                -o $function_select.outside_call
+            #end if
+        #else if $function_select.function == "report":
+            -p $function_select.pheno_data
+            -j output.json
+            -o output,html
+            #if $function_select.title:
+                -o $function_select.title
```
bgruening, line 26: single quote

tools/pharmcat/pharmcat.xml
```diff
@@ -0,0 +1,215 @@
+<tool id="pharmcat" name="pharmcat" version="@WRAPPER_VERSION@+@VERSION_SUFFIX@" profile="20.01">
```
bgruening, line 1: a better name and description would be nice

tools/pharmcat/pharmcat.xml
```diff
@@ -0,0 +1,215 @@
...
+
+    ## PHENOTYPE CALLER FUNCTION
+        #else if $function_select.function == 'pheno':
+            #if $function_select.method.pheno_function == 'vcf':
+                ln -s '$function_select.method.input' ./input.vcf &&
+                java -cp /pharmcat/pharmcat.jar
+                org.pharmgkb.pharmcat.phenotype.Phenotyper
+                -vcf input.vcf
+            #else:
+                ln -s '$function_select.method.input' ./input.json &&
```
bgruening, line 27: those ln can be moved out of the if/else

astrovsky01: Each of those have a different file type and require their extensions be correct. Since the input is inconsistent, it needs to be in there

bgruening: ok

bgruening (approved):

PR #1188, opened 2022-01-27
update wtdbg to wtdbg2.5

GraceAHall:
Hi Björn

Here is the wtdbg2 update. 
There is a substantial difference between the wtdbg and wtdbg2.5 options, so the whole wtdbg.xml file was rewritten.


bgruening (approved):

bgruening:
Great, thanks a lot!

PR #1190, opened 2022-02-05
update bionano

bgruening:


bgruening:
@gallardoalba do you have any other changes, UX wise, that you want to get included?

gallardoalba:
> @gallardoalba do you have any other changes, UX wise, that you want to get included?

Not, I think that the wrapper is fine.

PR #1193, opened 2022-02-12
fix hic mode

bgruening:


tools/hifiasm/hifiasm.xml
```diff
@@ -354,12 +355,14 @@
             <output name="hic_acontig_graph" file="hifiasm-out-hifi-a.gfa" ftype="gfa1" />
             <output name="hic_balanced_contig_hap1_graph" ftype="gfa1">
                 <assert_contents>
-                    <has_size value="0"/>
```
bgruening, line 357: @gallardoalba using has_size 0 is dangerous as we oversee real errors like this one here.

gallardoalba: Sorry!

PR #1201, opened 2022-03-30
Adding single and double quotes to valid values for key value args used in Keras tools.

kxk302:
I am seeing a problem in ‘Create a deep learning model architecture’ tool, used in workflow for CNN tutorial: 

https://training.galaxyproject.org/training-material/topics/statistics/tutorials/CNN/tutorial.html

In 'Create a deep learning model architecture’ step of the tutorial, in layer 2, I have:

"padding='same'" for “Type in key words arguments if different from the default”

Any text argument (not just padding) fails to parse correctly, whether I use single or double quotes. I get the following error message for double quotes:

NameError
   dict(padding=__dq__same__dq__)
                 ^^^
name '__dq__same__dq__' is not defined
  File "/jetstream2/scratch/test/jobs/1464482/tool_files/keras_deep_learning.py", line 397, in <module>
    config_keras_model(inputs, outfile)
  File "/jetstream2/scratch/test/jobs/1464482/tool_files/keras_deep_learning.py", line 264, in config_keras_model
    model = get_sequential_model(layers_config)
  File "/jetstream2/scratch/test/jobs/1464482/tool_files/keras_deep_learning.py", line 165, in get_sequential_model
    options.update(kwargs)
TypeError: 'NoneType' object is not iterable

This seems to be due to kwargs not allowing single and double quotes. The code change allows that. This is a bit puzzling a the same workflow worked fine in a January 2022 workshop presentation. There are recent changes to lib/galaxy/tools/evaluation.py:build_param_dict(), but I'm not sure whether thats the culprit or not.

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,17 +42,17 @@ pip install planemo
 
 To run the tests for a specific tool (e.g., pca) in a specific folder (e.g. sklearn)
 ```
-planemo test --docker tools/sklearn/pca.xml
+planemo test --biocontainers tools/sklearn/pca.xml
 ```
 
 To run the tests for all tools in a specific folder (e.g. sklearn)
 ```
-planemo test --docker tools/sklearn
+planemo test --biocontainers tools/sklearn
 ```
 
 To run the tests for all tools in all folders 
 ```
-planemo test --docker tools
+planemo test --biocontainers tools
 ```
 
 3 steps to get your tool into Galaxy - A real-world example
--- a/tools/sklearn/clf_metrics.xml
+++ b/tools/sklearn/clf_metrics.xml
@@ -250,7 +250,6 @@ with open("$outfile", 'w+') as out_file:
                 <expand macro="clf_inputs" multiple1="true" multiple="true" />
                 <section name="options" title="Advanced Options" expanded="False">
                     <expand macro="average">
-                        <option value="macro" selected="true">Calculate metrics for each label, and find their unweighted mean. (macro)</option>
                     </expand>
                 </section>
             </when>
--- a/tools/sklearn/keras_macros.xml
+++ b/tools/sklearn/keras_macros.xml
@@ -701,7 +701,14 @@
   <!--Simple key words text parameters, conbined to reduce UI latency-->
 
   <xml name="simple_kwargs" token_help="Leave blank for default.">
-    <param argument="kwargs" type="text" value="" label="Type in key words arguments if different from the default" help="@HELP@"/>
+    <param argument="kwargs" type="text" value="" label="Type in key words arguments if different from the default" help="@HELP@">
+      <sanitizer>
+        <valid initial="default">
+          <add value="&apos;" />
+          <add value="&quot;" />
+        </valid>
+      </sanitizer>
+    </param>
   </xml>
 
   <!-- Keras CallBacks -->
--- a/tools/sklearn/keras_macros.xml
+++ b/tools/sklearn/keras_macros.xml
@@ -36,7 +36,7 @@
       <option value="tanh" selected="@TANH@">tanh</option>
       <option value="sigmoid">sigmoid</option>
       <option value="hard_sigmoid">hard_sigmoid</option>
-      <option value="exponential">tanh</option>
+      <option value="exponential">exponential</option>
     </param>
   </xml>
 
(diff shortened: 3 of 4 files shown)


bgruening:
Can you please fix the linting as well?

The problem.with the quotes is that they are potentially dangerous when exposed to the command line. I'm only on my mobile , can you check that this parameter is never exposed to CMD directly?

kxk302:
> Can you please fix the linting as well?
> 
> The problem.with the quotes is that they are potentially dangerous when exposed to the command line. I'm only on my mobile , can you check that this parameter is never exposed to CMD directly?

Will do. 

Will also check if param is directly exposed to the command. 

It might make sense to localize this change, that is instead of applying it to 'kwargs', create something called 'text_kwargs', that allows quotes, and use text_kwargs in Keras tools instead.

kxk302:
> Can you please fix the linting as well?
> 
> The problem.with the quotes is that they are potentially dangerous when exposed to the command line. I'm only on my mobile , can you check that this parameter is never exposed to CMD directly?

@bgruening the kwargs parameter, which must allow single/double quotes, goes into a file, containing a parameter dictionary. The file is then passed to the command line.

kxk302:
Also, when I run lint locally, I do not get the errors under the Checks tab.
I run the following command: 

planemo shed_lint --tools --ensure_metadata --urls --report_level warn --fail_level error --recursive tools/sklearn

My Python version is 3.9 though.

kxk302:
> > Can you please fix the linting as well?
> > The problem.with the quotes is that they are potentially dangerous when exposed to the command line. I'm only on my mobile , can you check that this parameter is never exposed to CMD directly?
> 
> @bgruening the kwargs parameter, which must allow single/double quotes, goes into a file, containing a parameter dictionary. The file is then passed to the command line.

@bgruening Is this setup ok? thanks!

bgruening:
Yes, I think so, hope so :)

The tests are still failing, can you look at that?

anuprulez:
I investigated this a bit and found out that in `keras_macros.xml` file, there are multiple options defined for (at least) one select box having the same value. For example - in lines
https://github.com/kxk302/galaxytools/blob/keras_kwargs/tools/sklearn/keras_macros.xml#L906 and https://github.com/kxk302/galaxytools/blob/keras_kwargs/tools/sklearn/keras_macros.xml#L907, `option='mae'` is mapped to two different select options. The values of each option should be unique.

```
<option value="mae">mae / MAE / mean_absolute_error</option>
<option value="mae">mape / MAPE / mean_absolute_percentage_error</option>
```
In my opinion, for this option `<option value="mae">mape / MAPE / mean_absolute_percentage_error</option>`, it should be written as `<option value="mape">mape / MAPE / mean_absolute_percentage_error</option>` or anything different.

The linter is pointing to different locations (different tools such as `keras_model_config.xml`, `keras_model_builder.xml` and `clf_metrics.xml`) where I hope the same issue is present.

ping @kxk302 @bgruening

bgruening:
Well spotted @anuprulez!

kxk302:
I posted this in the wrong ticket!

Thanks @anuprulez, I fixed a few errors in options and now all the lint tests pass. Can this be reviewed/approved now?

anuprulez:
@kxk302 I found out (from here: https://github.com/bgruening/galaxytools/actions/runs/2202878953 that provides all tool test results) that tests from this tool (sklearn/lightgbm) are failing. Not sure why these tests are failing, we need to find the fix though. 

test_results_loghtgbm

anuprulez:
However, both the tests for the `lightgbm` tool pass locally using Planemo

kxk302:
Thanks @anuprulez, I'm also running the lightgbm tests now locally. Have had multiple issues with Planemo :(
Is there a way to re-start the tests on Github to see if they pass now?

kxk302:
I updated the README.md so biocontainers are used for Planemo testing. Let's see if the new build succeeds.

kxk302:
@anuprulez the tests pass now! Can you please merge this? Thx

anuprulez:
@kxk302 Looks good to me. @bgruening can you merge it? Thanks!

kxk302:
@bgruening When you get a chance, could you please merge this? Thanks.

bgruening:
Thanks!

PR #1202, opened 2022-04-01
[RFC] Select compute resource for training deep learning model

anuprulez:
I would like to add a section in the "Deep learning training and evaluation conduct deep training and evaluation" tool to take input from users on which compute resources they want to train their deep learning model - CPUs or GPU. The default would be CPUs but if someone wants to use GPUs, it should also be possible. Currently, on the EU server, the default processing is GPU which may cause issues when all of them are occupied and users who would like to run this deep learning tool cannot process their jobs. Therefore, this choice of selection of computing resources would help users optimize their choices of job processing. If they want to use GPUs and none are free at the moment, still they can switch to CPUs and process their jobs. If GPUs are free, users would be able to process their jobs faster. I have tested this in my local machine having CPUs, it works fine. I will run it on a VM to test GPU settings as well. 

However, I am not sure what should we change in the tool_destination YAML file so that it recognises the compute resource type input coming from this tool and automatically creates the right compute resource based on this user preference. Let me know what you think. Thanks!

A similar feature is already implemented here: https://usegalaxy.eu/tool_runner?tool_id=toolshed.g2.bx.psu.edu%2Frepos%2Fchemteam%2Fgmx_sim%2Fgmx_sim%2F2022%2Bgalaxy0

UI changes:

compute_resource_dl

ping @gmauro @bgruening @qiagu @kxk302

### DIFF
--- a/tools/sklearn/keras_macros.xml
+++ b/tools/sklearn/keras_macros.xml
@@ -917,6 +917,16 @@
     </section>
   </xml>
 
+  <!-- Compute resource -->
+  <xml name="tf_compute_resource">
+    <section name="compute_resource" title="Compute resources" expanded="true">
+      <param name="type_compute_resource" type="select" label="Select processing unit" help="Default processing unit is CPU. Select GPU for faster training of a deep learning model.">
+        <option value="tf_cpu" selected="true">CPU (default)</option>
+        <option value="tf_gpu">GPU</option>
+      </param>
+    </section>
+  </xml>
+
  <!--Citation-->
   <xml name="keras_citation">
     <citation type="bibtex">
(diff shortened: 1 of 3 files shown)


gmauro:
There is no need to implement anything in the wrapper's UI. It's made by Galaxy with a special flag into  tools configuration.
Appending [this](https://github.com/usegalaxy-eu/infrastructure-playbook/blob/master/files/galaxy/config/job_resource_params_conf.xml) to the wrapper.
The job dispatching will be managed by SortingHat, you need only to handle the `GPU_AVAILABLE` variable (possible values 0,1) like [here](https://github.com/galaxycomputationalchemistry/galaxy-tools-compchem/blob/5536c9fbb2f680a5f296db81362099e5a68e0587/tools/gromacs/sim.xml#L46)
We can discuss it better next week

anuprulez:
The discussion moved to https://github.com/usegalaxy-eu/infrastructure-playbook/pull/367. Closing this one

PR #1208, opened 2022-05-02
Antismash: update version 6.1.1

gallardoalba:
NOTE: As the antiSMASH-DB links have changed slightly to be more future proof, the clusterblast database format has also changed slightly to take advantage of that. It is necessary to update the clusterblast database at the same time as updating to 5.2 from previous versions.

This PR includes the following modifications:
- Parameters --cf-create-clusters and --cf-borders-only have been removed (deprecated)
- New parameters: --rre, --cc-mibig, --hmmdetection-strictness, --cb-nclusters, cb-min-homology-scale, --tigrfam
- Option have been reorganized
- Include new tests and the test 02 has been fixed.


tools/antismash/antismash.xml
```diff
@@ -1,9 +1,10 @@
-<?xml version='1.0' encoding='utf-8'?>
-<tool id="antismash" name="Antismash" version="5.1.2" profile="17.01">
+<tool id="antismash" name="Antismash" version="@TOOL_VERSION@+galaxy@VERSION_SUFFIX@" profile="17.01">
```
bgruening, line 1: time to increase the profile version?

bgruening (commented):
I still try to build the new version.

tools/antismash/antismash.xml
gallardoalba: Sure!

PR #1211, opened 2022-05-05
Add license

### LINKED ISSUE #1210: License
There doesn't seem to be a license specified, wouldn't that be nice to have? Someone on Gitter asked if the ML tools were open source.

simonbray:
Fix #1210.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,21 @@
+The MIT License (MIT)
+
+Copyright (c) 2014 galaxy-iuc
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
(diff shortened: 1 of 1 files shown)


LICENSE
```diff
@@ -0,0 +1,21 @@
+The MIT License (MIT)
+
+Copyright (c) 2012-present Björn Grüning and other contributors
```
simonbray, line 3: @bgruening you think this line is ok?

bgruening: No idea, is that needed? I mean my name and the authors?

simonbray: No idea either, sorry :/

I just copied from IUC. It seems it is in https://mit-license.org/ and BioConda has that line as well.

bgruening: Then lets write:

> Copyright (c) 2012-present The 'galaxytoools' contributors

bgruening: Thanks

bgruening (approved):

PR #1223, opened 2022-06-30
Add sideloading to AntiSMASH

BartvdHout-TimKuijten:


hexylena:
From some students of mine @bgruening :) (please feel free to squash, there were some git accidents)

hexylena:
Their server went down, test data coming later.

tools/hAMRonize2AntiSMASH/hAMRonize2AntiSMASH.xml
```diff
@@ -0,0 +1,21 @@
+<tool id="hAMRonize2AntiSMASH" name="Convert hAMRonize to AntiSMASH" version="1.0.0">
```
bgruening, line 1: Can we keep the ID in small letters?

bgruening: Maybe you can also add a "profile" to the tools.

tools/hAMRonize2AntiSMASH/hAMRonize2AntiSMASH.xml
```diff
@@ -0,0 +1,21 @@
...
+  <requirements><requirement type="package" version="3.9">python</requirement></requirements>
+  <command detect_errors="exit_code"><![CDATA[
+  ln -s '$input' input.json;
```
bgruening, line 4: please use && instead of ;

bgruening (commented):
Very nice addition @rekado 

I added some comments inline. I guess we also need to change the shed.yml file to be a suite if we really need to stick with the additional tool.

Alternatively, we also could fold the new tool into the main antismash tool. It only has an input and an output, so we could run this as a preprocessing step within the main tool.

tools/hAMRonize2AntiSMASH/hAMRonize2AntiSMASH.xml
hexylena: @BartvdHout-TimKuijten @TimKAvans please add `profile="21.09"` on line 1. This will also solve this issue.

bgruening:
Since it is so specialized my feeling is that it will never be used outside the scope of Antismash, so we can fold it as a parameter into it. 

(If a tool is in reality bound to exactly one other tool the only reason to keep it separate is large runtimes, imho. E.g. pre-processing in a separate tool with large runtime, plotting ontop of the result of that tools very specific outpuf.)

hexylena:
Folded into the tool, got the last two test cases passing. :crossed_fingers:

bgruening:
Can you please bump the tool version?

hexylena:
@bgruening it was done in https://github.com/bgruening/galaxytools/commit/f77c6577c4f90b17d2b8d0d9f916837b4905b0b5

bgruening (approved):

bgruening:
Oh macro, I see. sorry.

Issue #1231, opened 2022-08-08
Format of Search CV parameters


anuprulez:
As reported by one of our students, the tabular dataset returned by the search CV tool is populating the list differently as documented earlier in many ML tutorials. Do we know what needs to be changed in the tool parameter selection to make the search CV tool run? I remember we used to get a list of all hyperparameters and then we could select those whose values need to be optimized. 

![search_cv1](https://user-images.githubusercontent.com/3022518/183456840-512d8fbf-d5a2-45cd-ae07-e9f22d0ab9c3.png)

I tried a few ways to make it work - by changing tool's version, using different options in `Choose a parameter name (with current value)` but the jobs are failing. Do you know what should be the correct setting, then I will update the tutorials accordingly. 

link to the history: https://usegalaxy.eu/u/kumara/h/age-prediction

ping @qiagu 

Thanks a lot!

qiagu:
@anuprulez  Thanks for exposing the issue. My intuition is that changes of Galaxy's tool infrastructure caused the issue. I will look deep into it soon.

qiagu:
@anuprulez  Unfortunately, I couldn't reproduce the issue shown in the shared history. I have correct list of parameters in the dropdown in my local instance. One thing worth mentioning: use only the list in the `Search list`, having parameter name in the front, like `selectkbest__k: [5880, 5890, 5895, 5900]`, would lead to failure.

qiagu:
As wrapped here https://github.com/bgruening/galaxytools/blob/master/tools/sklearn/search_model_validation.xml#L11-L17, lines starting with `@` should be selected to be search parameters, i.e., showing as the dropdown options. The EU instance seems to do the opposite, excluding lines having `@`. I have no idea what causes that, since I can't reproduce the issue elsewhere.

anuprulez:
@qiagu thanks a lot for looking into it. I found a fix here: https://github.com/bgruening/galaxytools/pull/1232

anuprulez:
Adding quotes to @ symbol did not work, unfortunately. But, removing it did and the tests are now passing. Removing it will add a few extra params in the list of estimator params of which the most we already show using @ as a startwith filter. It looks like there is some issue with parsing of this special symbol on EU server in the tool form, not 100% sure though.

anuprulez:
Same issue on Galaxy Main as well. We should keep this issue open until it is fixed

[...]

PR #1232, opened 2022-08-10
Fix issue with param loading and tool tests

anuprulez:
This PR fixes the issue: https://github.com/bgruening/galaxytools/issues/1231

Because of the unquoted @ symbol, most of the parameters of the selected estimator fail to load from a tubular file containing all parameters of the estimator.

params

Without the quoted @, the dropdown fails to load all the params, shown below:

params_fail

EDIT: Removing @ for filtering estimator params fixes the issue

### DIFF
--- a/tools/sklearn/main_macros.xml
+++ b/tools/sklearn/main_macros.xml
@@ -1,5 +1,5 @@
 <macros>
-    <token name="@VERSION@">1.0.8.3</token>
+    <token name="@VERSION@">1.0.8.4</token>
 
     <xml name="python_requirements">
         <requirements>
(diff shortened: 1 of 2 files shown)


anuprulez:
ping @bgruening

bgruening:
Do you want to bump the tool version number?

bgruening (approved):

anuprulez:
updated the version of the tool suite!

qiagu:
This is good workaround. I'm fine with it. Just be aware that there might be lots of untunable options in the dropdown if no filter is in place.

bgruening:
> This is a good workaround. I'm fine with it. Just be aware that there might be lots of untunable options in the dropdown if no filter is in place.

Do you have a better alternative?

qiagu:
> Do you have a better alternative?

It seems this issue only exits on EU ( If it can be reproduced in other place, please let me know). I don't have alternatives.

bgruening:
Why should this be an EU issue only?

anuprulez:
I see the same issue on Galaxy Main

params_galaxy_main

PR #1237, opened 2022-09-26
Add htseqclip tool

michauhl:
Add Galaxy wrapper for htseq-clip, a toolset for the analysis of eCLIP/iCLIP datasets:
https://github.com/EMBL-Hentze-group/htseq-clip

### DIFF
--- a/tools/rna_tools/htseq-clip/htseq_clip.xml
+++ b/tools/rna_tools/htseq-clip/htseq_clip.xml
@@ -163,13 +163,13 @@
         <data name="windows_bed_file" format="bed" from_work_dir="windows.bed" label="${tool.name} on ${on_string}: Sliding windows BED file">
             <filter>action_type["action_type_selector"] == "create_sliding_windows"</filter>
         </data>
-        <data name="windows_txt_file" format="txt" from_work_dir="windows_mapped_to_ids.txt" label="${tool.name} on ${on_string}: Windows annotation table file (DEWSeq input)">
+        <data name="windows_txt_file" format="tabular" from_work_dir="windows_mapped_to_ids.txt" label="${tool.name} on ${on_string}: Windows annotation table file (DEWSeq input)">
             <filter>action_type["action_type_selector"] == "create_sliding_windows"</filter>
         </data>
-        <data name="sample_info_file" format="txt" from_work_dir="sample_info.txt" label="${tool.name} on ${on_string}: Sample information table file (DEWSeq input)">
+        <data name="sample_info_file" format="tabular" from_work_dir="sample_info.txt" label="${tool.name} on ${on_string}: Sample information table file (DEWSeq input)">
             <filter>action_type["action_type_selector"] == "create_count_table"</filter>
         </data>
-        <data name="count_matrix_file" format="txt" from_work_dir="count_matrix.txt" label="${tool.name} on ${on_string}: Count table file (DEWSeq input)">
+        <data name="count_matrix_file" format="tabular" from_work_dir="count_matrix.txt" label="${tool.name} on ${on_string}: Count table file (DEWSeq input)">
             <filter>action_type["action_type_selector"] == "create_count_table"</filter>
         </data>
     </outputs>


michauhl:
Tests are fine locally, but here it seems pysam.index() inside python script cannot create a BAM index file for the input test-data BAM files (probably write-protected ?):

pysam.utils.SamtoolsError: 'samtools returned with error 1: stdout=, stderr=samtools index: failed to create or write index\n'

How to solve this? Normally input BAM files come without index files, but index files need to created before using htseq-clip, otherwise it doesn't work. Or is this not needed inside Galaxy / are they created automatically ?

michauhl:
Oh well, the tool seems to be sensitive to computing environment (maybe tool versions), resulting in different row ordering in the output table files (therefore test file equals tests fail). Need to ask developer why this is the case ...

tools/rna_tools/htseq-clip/htseq_clip.xml
```diff
@@ -35,7 +35,7 @@
                 --hce-g $action_type.hce_options.hce_g
                 --hce-q $action_type.hce_options.hce_q
                 $action_type.hce_options.hce_primary
-                --hce-c 4
+                --hce-c 1
```
bgruening, line 38: this should use GALAXY_SLOTS I think you want to make it configurable, correct?

bgruening:
> Oh well, the tool seems to be sensitive to computing environment (maybe tool versions), resulting in different row ordering in the output table files (therefore test file equals tests fail). Need to ask developer why this is the case ...

You can add sort="true" to test-outputs https://docs.galaxyproject.org/en/latest/dev/schema.html#id106

tools/rna_tools/htseq-clip/htseq_clip.xml
michauhl: Yes, I was going to ask how to adjust this for Galaxy ...  :)

michauhl: How to write this here ? I looked in some wrappers and it is always written like:
` --cpus "\${GALAXY_SLOTS:-12}"`
What does the number mean ? Max CPUs supported ?

bgruening: That is bash syntax, it's means 12 if the variable is not set. A default vakue

michauhl: Then I'll give it a go with `--hce-c \${GALAXY_SLOTS:-1}`

tools/rna_tools/htseq-clip/.shed.yml
```diff
@@ -0,0 +1,18 @@
...
+  htseq-clip is a toolset for the analysis of eCLIP/iCLIP datasets
+homepage_url: https://github.com/EMBL-Hentze-group/htseq-clip
+long_description:
+  htseq-clip is a toolset for the analysis of eCLIP/iCLIP datasets. 
+  It can be used to generate files necessary for data analysis using 
+  the companion R/Bioconductor package DEWSeq.
+remote_repository_url: https://github.com/bgruening/galaxytools/tree/master/tools/rna_tools/htseq-clip
+type: unrestricted
+categories:
+- Sequence Analysis
```
bgruening, line 13: Those needs to be toolshed categories.... You can not add arbritrairy ones.

tools/rna_tools/htseq-clip/htseq_clip.xml
```diff
@@ -0,0 +1,263 @@
...
+                </section>
+
+            </when>
+
+            <when value="create_count_table">
+                <repeat name="exp_bams" min="1" title="CLIP-seq experiment BAM inputs">
+                    <param name="exp_bam" type="data" format="bam" label="BAM files belonging to the CLIP-seq experiment" help="Select BAM file belonging to the CLIP-seq experiment. NOTE that order determines replicate numbering in output tables"/>
+                </repeat>
+                <repeat name="ctr_bams" min="1" title="CLIP-seq control BAM inputs">
+                    <param name="ctr_bam" type="data" format="bam" label="BAM files belonging to the CLIP-seq control" help="Select BAM file belonging to the CLIP-seq control. NOTE that order determines replicate numbering in output tables"/>
```
bgruening, line 94: You can simply use multiple=true and avoid the repeat

tools/rna_tools/htseq-clip/.shed.yml
michauhl: There are no categories for CLIP-seq or RNA-binding proteins. Then I will just keep RNA and Sequence analysis.

tools/rna_tools/htseq-clip/htseq_clip.xml
michauhl: Wolfgang suggested this in the tech tips chat, as I want to keep the order of the input files (rep1, rep2 ... ).

michauhl:
@bgruening  From my side it can now be merged. Regarding RAM usage, the developer suggested >= 32GB, 64GB would be ideal. Regarding toolshed categories, is there a way to add categories (CLIP-seq?) for CLIP-seq related tools? Since there are a number of tools available on Galaxy by now.

bgruening:
> is there a way to add categories (CLIP-seq?) for CLIP-seq related tools? Since there are a number of tools available on Galaxy by now.

I think we can do that.

@martenson can you help here?

martenson:
@bgruening @michauhl Yeah, we can do that. We usually create one on TTS first and once there are actual tools present we create the MTS one. I've created the category on TTS(`CLIP-seq`) already, so please ping me when you have tools there ready for MTS. You can also supply description of such category, I've chosen something generic for now.

bgruening:
If we merge this is gets pushed to TTS and TS. So I guess it makes sense to have one on TS as well?

tools/rna_tools/htseq-clip/.shed.yml
```diff
@@ -0,0 +1,14 @@
...
+description: |
+  htseq-clip is a toolset for the analysis of eCLIP/iCLIP datasets
+homepage_url: https://github.com/EMBL-Hentze-group/htseq-clip
+long_description:
+  htseq-clip is a toolset for the analysis of eCLIP/iCLIP datasets. 
+  It can be used to generate files necessary for data analysis using 
+  the companion R/Bioconductor package DEWSeq.
+remote_repository_url: https://github.com/bgruening/galaxytools/tree/master/tools/rna_tools/htseq-clip
+type: unrestricted
+categories:
```
martenson, line 12: @bgruening do you mean that you want to put `CLIP-seq` here already?

bgruening: @michauhl this is what I understood, correct?

michauhl: @bgruening @martenson if possible, yes.

martenson: Sounds great, done.

bgruening: @michauhl check out https://toolshed.g2.bx.psu.edu/ and add your preferred categories :)

michauhl: @martenson @bgruening perfect, thank you! :)

michauhl:
@bgruening can you merge the pull request? From my side everything looks fine now.

PR #1252, opened 2022-10-24
add hicup2juicer

lldelisle:
Hi,
I finally need a wrapper for one of the converter. Should I add `+galaxy1` to other hicup tools?
Thanks,
Lucille


bgruening (approved):
If we are lucky we only get this one new tool and the others are not touched :)

PR #1255, opened 2022-10-25
add toolshed categories

michauhl:
Add more categories


michauhl:
Well that's great ... 8)

PR #1256, opened 2022-10-25
hifiasm: output noseq.gfa files as collection

abueg:
👋🏼 !

I have edited the hifiasm xml to output the `noseq.gfa` files as a collection. I have added a line at the end of the tool's command to make a noseq folder and put the noseq files in there. They are small files (they're the GFAs minus the actual sequence, so much smaller) so I have them output by default.


bgruening:
LGTM! Do you think you need that output by default?

bgruening:
Then lets get this in!

PR #1276, opened 2022-12-21
gfastats: 1.3.6 version bump

abueg:
hello! version bump for gfastats to version 1.3.6 🧬


abueg:
ah wait!! let me add the new flag to the tool wrapper before merging the PR, please 😅 almost forgot about that

bgruening:
Linting also got stricter and fails now ;-)

tools/gfastats/gfastats.xml
```diff
@@ -188,6 +194,22 @@
                             label="Report actual sequence" help="It reports also the actual sequence"/>
                     </when>
                 </conditional>
+                <conditional name="locale_condition">
+                    <param name="locale_selector" type="select" label="Locale">
```
bgruening, line 198: wow, isn't that too detailed? Should we just pick utf-8 as default and foce users to use it?

abueg: hmm I think it would still be useful to have the non-utf8, because that output would be machine-readable without having to take the commas out... maybe the one with commas should be default then, since it's more human-readable, and then have an option for the comma-less output?

abueg: latest commit is trying it this way, with utf-8 as default and a toggle to turn it off 👍🏼 it does look more straightforward!

abueg:
edited test 1 since the SAK file wasn't actually properly pointing to the contig to be removed, so the output and input ended up effectively the same; this should now actually be checking for the SAK functionality

some weirdness though in that tests 1 & 8 are recurrently failing when testing on my M1 Mac, but the tests work fine on github actions. test 1 on my M1 Mac results in a very malformed FASTA output when converting from FASTQ.GZ, but the same file gunzipped and used as a FASTQ results in an expected FASTA

tools/gfastats/gfastats.xml
```diff
@@ -44,6 +44,15 @@
                     --line-length $mode_condition.output_condition.line_length
                 #end if
             #end if
+            #if $mode_condition.output_condition.out_format == 'gfa'
+                #if $mode_condition.output_condition.terminal_overlaps_condition.terminal_overlaps_select == 'yes':
+                --discover-terminal-overlaps $mode_condition.output_condition.terminal_overlaps_condition.terminal_overlaps_length
+                #end if
+            #else if $mode_condition.output_condition.out_format == 'gfa.gz'
```
bgruening, line 51: It might be the fever, but both the if and the else is the same, is that correct? of the we can remove the if

bgruening: of wite someting like `#if $mode_condition.output_condition.out_format in ['gfa', 'gfa.gz']`

abueg: it's not the fever, you're right! just pushed a commit using that syntax for the `--discover-terminal-overlaps` case and the `--line-length` case, thank you! also feel better soon :(

bgruening (approved):

bgruening:
Thanks a lot!

PR #1281, opened 2023-01-21
hifiasm: ONT ultra-long integration

abueg:
hello! 👋🏼 

I started working on editing the wrapper for ONT UL integration, basing it off of the HiC integration. I think it's mostly working (including a test I'm using the `nanopore.fasta` from the flye test data for) but I have a problem with trying to use multiple ONT inputs... 

image

it seems to think the name of one file is `'input_0.fasta.gz input_1.fasta.gz'` instead of two files: `input_0.fasta.gz` and `input_1.fasta.gz`. 

if there isn't an easy workaround for this, i guess we could just have the user concatenate their inputs?


bgruening:
Input staging problem: Test input file (nanopore.fasta.gz) cannot be found.

bgruening:
grafik

Test 12 was killed, I guess it used to much memory?

abueg:
trying with a subset of the nanopore input instead...

abueg:
OK, disabling the initial bloom filter worked for test 12 to no longer be killed 👍🏼 ! 

I am wondering if you have any suggestions for how to account for multiple read files as input, though? (see error in first post)

tools/hifiasm/hifiasm.xml
```diff
@@ -125,6 +134,16 @@
             --l-msjoin $hic_partition.l_msjoin
         #end if
 
+        #if str($ont_integration.ont_integration_selector) == 'set':
+            --ul '${ ' '.join(["%s" % $x for $x in $ultralong_inputs]) }'
```
bgruening, line 138: For your other question, I think this here is wrong - at least in combination with the outer `'`.

So what is happening here. the join is creating a string like `foo bar baz` and then you are putting `'` around it - yielding to `'foo bar baz'`. But you want `'foo' 'bar' 'baz'`.

Try something like that ... note the double inner quotes and the removed outer quotes.

`  ${ ' '.join(["'%s'" % $x for $x in $ultralong_inputs]) }  `

abueg: image
looks good! should i change the syntax on the HiC partition section to match as well?

bgruening: Yes please!

bgruening (approved):

bgruening:
:heavy_check_mark:  thanks a lot!

abueg:
of course, thank you for your help + patience!! 🙏🏼

PR #1283, opened 2023-01-24
hifiasm: multi-file input syntax fix

abueg:
I am sorry about the third pull request in like a week 😭 

so I looked back again at the help from hifiasm and the syntax for multiple files for ONT/HiC is actually as such in the latest version:
```  
Hi-C-partition:
    --h1 FILEs   file names of Hi-C R1  [r1_1.fq,r1_2.fq,...]
    --h2 FILEs   file names of Hi-C R2  [r2_1.fq,r2_2.fq,...]
  Ultra-Long-integration (beta):
    --ul FILEs   file names of Ultra-Long reads [r1.fq,r2.fq,...]
```

so I just edited the `join` command in the command-building part to delimit with a `,` instead of a space, does this look OK?

here is how the command builds when I tried this out:
image


tools/hifiasm/hifiasm.xml
```diff
@@ -117,8 +117,8 @@
         #end if
 
         #if str($hic_partition.hic_partition_selector) == 'set':
-            --h1 ${ ' '.join(["'%s'" % $x for $x in $hic1_inputs]) }
-            --h2 ${ ' '.join(["'%s'" % $x for $x in $hic2_inputs]) }
+            --h1 ${ ','.join(["%s" % $x for $x in $hic1_inputs]) }
```
bgruening, line 120: why do you remove the quotes? They still need to be quoted, correct?

Maybe you want to add a test, you can simply add the same two files in your latest test I think.

abueg: ah I took the quotes out to it would match with the example in the hifiasm help (`r1.fq,r2.fq` instead of `'r1.fq','r2.fq'`), but I can put them back in if that is better practice?

and yes I'll add a test 👍🏼

bgruening: imaging your data is stored on a path that has spaces in its names, like "/home/stupid bjoern/foo/bar" ... if you do not use single quotes on bash if you specify this path, every tool will interpret it as two paths... that why we need to have everything single quotes ... because we never know if people run those tools on a path with "spaces in its path".

abueg: ohhh okay, I understand now, sorry about that! I'll put them back 👍🏼

abueg:
so I put the single quotes back, and made a test that checks for the multi-input for UL, and the command built looks OK (the file paths are in single quotes), but the command reported in the logfile doesn't have the single quotes, is this OK since the paths are still getting passed properly to the program?
image

bgruening:
`-ul` seems to have quotes, for me this looks good. Can you explain where you don't see quotes?

abueg:
ah, in the logfile output (the top code block) the paths aren't in quotes, though that is probably fine since hifiasm is reading the paths correctly?

bgruening:
Merci!

PR #1292, opened 2023-02-21
hifiasm: added `bin` file output

abueg:
hello! 👋🏼 

i added an option for outputting/keeping the `.bin` files. it's off by default as the files are somewhat large, but i think it'd be good to have the option to keep it, as the `.bin` files are useful for debugging 👍🏼


tools/hifiasm/hifiasm.xml
```diff
@@ -268,20 +268,44 @@
                 </param>
             </when>
         </conditional>
-        <param name="log_out" type="boolean" label="Output log file?" truevalue="yes" falsevalue="no"/> 
+        <conditional name="bins">
```
bgruening, line 271: Should this be a boolean instead of a conditional?

tools/hifiasm/hifiasm.xml
```diff
@@ -268,20 +268,44 @@
...
+                <option value="yes">yes</option>
+            </param>
+            <when value="no" />
+            <when value="yes" />
+        </conditional>
+        <param name="log_out" type="boolean" label="Output log file?" truevalue="yes" falsevalue="no" /> 
     </inputs>
     <outputs>
+        <!-- Yes to bin files -->
+        <data name="ec_bin" format="binary" from_work_dir="output.ec.bin" label="${tool.name} on ${on_string}: ec .bin file">
```
bgruening, line 283: Would a collection make more sense in this case?

abueg: hmm initially i had it as a boolean plus collection, but switched because 1) thought they might be easier to export as individual items, and 2) i was having trouble getting the wrapper to not create an empty collection in the case of the boolean being false... 😓

bgruening: Mh, you can download entire collections and you can dig into them to download single files. Is that not good enough?

bgruening: I can help with the empty-collection hopefully :)

abueg: oh i forgot about downloading the whole collection, that actually would be handy for these haha -- ok i'll change it! pls take a look at the empty collection when you have a chance, thank you 😄

tools/hifiasm/hifiasm.xml
```diff
@@ -522,36 +532,45 @@
                 </assert_contents>
             </output>
         </test>
+        <!-- TEST 14: test bin files -->
+        <test expect_num_outputs="6">
+            <param name="reads" value="hifiasm-in1.fa.gz" ftype="fasta.gz" />
+            <param name="filter_bits" value="0" />
+            <param name="mode_selector" value="standard" />
+            <param name="bins_out" value="yes" />
+            <output name="ovlp_source_bin" file="hifiasm-out13.ovlp.source.bin" ftype="binary" />
```
bgruening, line 541: @abueg what is `ovlp_source_bin`

We have this now:

```
.. ERROR: Test 14: Found output tag with unknown name [ovlp_source_bin], valid names ['raw_unitigs', 'processed_unitigs', 'primary_contig_graph', 'alternate_contig_graph', 'hap1_contigs', 'hap2_contigs', 'raw_unitigs_trio', 'processed_unitigs_trio', 'hic_pcontig_graph', 'hic_acontig_graph', 'hic_balanced_contig_hap1_graph', 'hic_balanced_contig_hap2_graph', 'hic_raw_initig', 'hap1_contigs_hic', 'hap2_contigs_hic', 'log_file', 'noseq_files', 'bin_files']
```

Should this be the new collection `bin_files`?

bgruening: maybe you can just define collection and a `count` https://docs.galaxyproject.org/en/latest/dev/schema.html#id136
? then you don't need to put the binaries in the repo

abueg: yay thank you for pointing me to that, i think this works!

bgruening (approved):

PR #1293, opened 2023-02-26
Bionano: fixes in when generating keys.txt dataset

gallardoalba:


tools/bionano/bionano_scaffold.xml
```diff
@@ -60,7 +60,7 @@
             && python '$__tool_directory__/remove_fake_cut_sites.py' $output_file_NCBI 'SCAFFOLD_NCBI_trimmed.fasta' 'output.log'
             && python '$__tool_directory__/remove_fake_cut_sites.py' $output_file_not_scaffolded 'NOT_SCAFFOLDED_trimmed.fasta' 'output.log'
         #end if
-        && echo >> hybrid_scaffolds/*cut.txt
+        && touch hybrid_scaffolds/*cut.txt
```
bgruening, line 63: will this not create a `*cut.txt` file?

gallardoalba: Yes, it will create if it doesn't exist.

bgruening: but the star is part of the filename ...
And as far as I can see this is what caused the crash I have sent you...

why not `cat hybrid_scaffolds/*cut.txt '$key_file'` ... without touch and and mv at all.

tools/bionano/bionano_scaffold.xml
```diff
@@ -60,8 +60,8 @@
             && python '$__tool_directory__/remove_fake_cut_sites.py' $output_file_NCBI 'SCAFFOLD_NCBI_trimmed.fasta' 'output.log'
             && python '$__tool_directory__/remove_fake_cut_sites.py' $output_file_not_scaffolded 'NOT_SCAFFOLDED_trimmed.fasta' 'output.log'
         #end if
-        && echo >> hybrid_scaffolds/*cut.txt
-        && mv hybrid_scaffolds/*cut.txt hybrid_scaffolds/keys.txt
+        && touch hybrid_scaffolds/temp_cut.txt
+        && cat hybrid_scaffolds/*cut.txt hybrid_scaffolds/keys.txt
```
gallardoalba, line 64:
```suggestion
        && cat hybrid_scaffolds/*cut.txt > hybrid_scaffolds/keys.txt
```

tools/bionano/bionano_scaffold.xml
```diff
@@ -60,8 +60,8 @@
             && python '$__tool_directory__/remove_fake_cut_sites.py' $output_file_NCBI 'SCAFFOLD_NCBI_trimmed.fasta' 'output.log'
             && python '$__tool_directory__/remove_fake_cut_sites.py' $output_file_not_scaffolded 'NOT_SCAFFOLDED_trimmed.fasta' 'output.log'
         #end if
-        && echo >> hybrid_scaffolds/*cut.txt
-        && mv hybrid_scaffolds/*cut.txt hybrid_scaffolds/keys.txt
+        && touch hybrid_scaffolds/temp_cut.txt
```
bgruening, line 63: maybe add a comment why we need to do that

bgruening (approved):

PR #1295, opened 2023-03-14
Update blobtoolkit v4.0.7

gallardoalba:


tools/blobtoolkit/blobtoolkit.xml
```diff
@@ -39,7 +39,7 @@
             #end if
             #if $mode_conditional.cov
                 ln -s '$mode_conditional.cov' './input.bam' &&
-                ln -s '$mode_conditional.cov.metadata.bam_index' './input.bam.bai' &&
+                ln -s '$mode_conditional.cov.metadata.bam_index' './input.bam.csi' &&
```
bgruening, line 42: mh, most of the times it is a bai index I would assume. BAI is not the same as CSI.

gallardoalba: Ah, indeed. In that case blobtool is using csi; I'll include that change.

bgruening (approved):

PR #1296, opened 2023-03-27
Update blobtools

gallardoalba:
This PR removes the plot options. In addition, it avoid to store the taxon database in the Blobtool.tgz directory.


tools/blobtoolkit/blobtoolkit.xml
```diff
@@ -189,6 +174,7 @@
             </when>
             <when value="add">
                 <param name="blobdir" type="data" format="tgz" label="Blobdir.tgz file" help="This file should be generated by Blobtool create" />
+                <param argument="--taxdump" type="data" format="tar,tgz,gz" label="NCBI taxdump directory" help="It should be compressed in tar.gz format"/>
```
bgruening, line 177: Do you think we should add a galaxy data type for it? Maybe even zip? Or should we stick to this for the time being?

gallardoalba: Indeed it seems to be better to use the zip file, because the tar.gz is recognized by Galaxy as gzip, triggering in the tool.

gallardoalba: Unfortunately the container doesn't include zip and unzip...

tools/blobtoolkit/macros.xml
```diff
@@ -42,6 +42,8 @@
             </when>
             <when value="disabled"/>
         </conditional>
-
+    </xml>
+    <xml name="macro_taxdump">
+        <param argument="--taxdump" type="data" format="tar" label="NCBI taxdump directory" help="Please, check that the datatype is set as tar; during the upload process, Galaxy decompress it automatically. You can find more information about it in the help section."/>
```
bgruening, line 47: Is that really tar data type or tar.gz?

gallardoalba: Indeed it also works even if Galaxy recognizes it as gz instead of tar, so I don't think it is a problem after all.

gallardoalba: The main requirement is that it should be the last taxdump version.

PR #1300, opened 2023-04-05
Blobtookit: Undo partially changes regarding folder structure

gallardoalba:
Blobtool view (the interactive tool)  requires a folder structure in order to work.

Test history: https://usegalaxy.eu/u/gallardoalba/h/blobtoolkit-test

The upper interactive tool makes use of the last blobtool version (without internal folder structure); the other makes use of the previous tool version.


bgruening (approved):

tools/blobtoolkit/blobtoolkit.xml
```diff
@@ -156,7 +148,7 @@
                 #end if
                 ./Blobdir
             #end if
-            && tar -C ./Blobdir -zcvf Blobdir.tgz .
+            && tar -zcf './Blobdir.tgz' './Blobdir'
```
bgruening, line 151: why is here blobdir?

What are you trying here? A folder in the tarball or no folder in the tarball?

gallardoalba: Yes, it includes the files inside a folder. In that way the interactive tool seems to work correctly

bgruening: Of course, we need to adopt the interactive tool to NOT take the folder.

The question is what is more user-friendly. What is easier to explain to users and then we adopt both tools.

gallardoalba: Yes, I agree, but from the users point of view I don't think there's a great difference if compressing the files with without the folder.

bgruening: How should the user know that a folder is needed?

gallardoalba: Well, this dataset is generated by the tool itself. If they want to upload a blobdir folder just to use the interactive tool, I think it would be possible to include that information in the help section perhaps?

bgruening: That is the big question here. And yes I think that is a use-case. People with data, that just want to visualise them.

gallardoalba: Ok, I'm modifying the interactive tool in order to avoid those changes, probably it will be more clear in that way. Also I'll include some info there.

gallardoalba: Also it should work https://github.com/usegalaxy-eu/galaxy/pull/158

PR #1310, opened 2023-05-22
Add AGAT tool

gallardoalba:
AGAT has the power to check, fix, pad missing information (features/attributes) of any kind of GTF and GFF to create complete, sorted and standardised gff3 format. Over the years it has been enriched by many many tools to perform just about any tasks that is possible related to GTF/GFF format files (sanitizing, conversions, merging, modifying, filtering, FASTA sequence extraction, adding information, etc). Comparing to other methods AGAT is robust to even the most despicable GTF/GFF files.

Source: https://github.com/NBISweden/AGAT


bgruening (approved):

bgruening (commented):
maybe compress the genome.fasta?

tools/agat/macros.xml
```diff
@@ -0,0 +1,41 @@
...
+    <xml name="citations">
+        <citations>
+            <citation type="doi">10.5281/zenodo.3552717</citation>
+        </citations>
+    </xml>
+    <xml name="ANNOTATION_INPUT" token_format="gff,gtf,gff3,gff3.gz">
+        <param argument="--gff" type="data" format="@FORMAT@" label="Annotation file" help="Input GTF/GFF file" />
+    </xml>
+    <xml name="REFERENCE_FASTA">
+        <param argument="--fasta" type="data" format="fasta" label="Reference FASTA file" help="Input FASTA file" />
```
bgruening, line 24: should that not use the all_fasta loc file as well?

gallardoalba: Sorry, I do not understand what do you mean.

bgruening: Its a reference genome. Should it not be use the Galaxy shipped reference genomes - at least optionally?

gallardoalba: > Its a reference genome. Should it not be use the Galaxy shipped reference genomes - at least optionally?

Ah, yes, I'll include also the references from the databases.

gallardoalba:
Do you think any of the missing functions could be interesting for you @abretaud? I can include them if needed.

PR #1314, opened 2023-05-26
Add fastqsanger acceptance for racon

hexylena:
Discovered by @shiltemann

(I feel like `fastq` and `fastq.gz` should be sufficient to accept both of these but I guess not?)

### DIFF
--- a/tools/racon/racon.xml
+++ b/tools/racon/racon.xml
@@ -53,7 +53,7 @@
             > racon_polished_consensus.fa  
     ]]></command>
     <inputs>
-        <param type="data" name="reads" format="fasta,fasta.gz,fastq,fastq.gz" label="Sequences" help="input file in FASTA/FASTQ format (can be compressed with gzip) containing sequences used for correction"/>
+        <param type="data" name="reads" format="fasta,fasta.gz,fastq,fastq.gz,fastqsanger,fastqsanger.gz" label="Sequences" help="input file in FASTA/FASTQ format (can be compressed with gzip) containing sequences used for correction"/>
         <param type="data" name="overlaps" format="paf,sam,tabular" label="Overlaps" help="input file in MHAP/PAF/SAM format (can be compressed with gzip) containing overlaps between sequences and target sequences"/>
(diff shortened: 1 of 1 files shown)


tools/racon/racon.xml
```diff
@@ -53,7 +53,7 @@
             > racon_polished_consensus.fa  
     ]]></command>
     <inputs>
-        <param type="data" name="reads" format="fasta,fasta.gz,fastq,fastq.gz" label="Sequences" help="input file in FASTA/FASTQ format (can be compressed with gzip) containing sequences used for correction"/>
+        <param type="data" name="reads" format="fasta,fasta.gz,fastq,fastq.gz,fastqsanger,fastqsanger.gz" label="Sequences" help="input file in FASTA/FASTQ format (can be compressed with gzip) containing sequences used for correction"/>
```
mvdbeek, line 56: How did you find that's necessary ? `fastqsanger` and `fastqsanger.gz` are subclassews of fastq / fastq.gz, so this probably has no effect.

shiltemann: We got several reports during Smorg, running into the error

```
[racon::createPolisher] error: file reads.fastqsanger.gz has unsupported format extension (valid extensions: .fasta, .fasta.gz, .fna, .fna.gz, .fa, .fa.gz, .fastq, .fastq.gz, .fq, .fq.gz)!
```

And changing datatype to fastq.gz fixed the problem

shiltemann: I don't think it was a problem of accepting the datasets into the form, but rather how the are passed to racon

bgruening: Wired, https://github.com/bgruening/galaxytools/pull/1314/files#diff-478bcb402461d8d301bf9a40f39ce0f414e45901bb4d111d3f695ace4cb28d1dR9 should take care of that?

Was it an old tool verison?

shiltemann: ahh yes of course, they must've clicked the link in the tutorial in gtn-in-galaxy, thanks @bgruening, should have investigated a bit more

bgruening: Cool, can this be closed?

shiltemann: yes please close, thanks :)

PR #1318, opened 2023-06-28
Adding Squidpy

tcollins2011:
Implementation of squidpy spatialomics for 10x visium data analysis.


tcollins2011:
@astrovsky01 Tagging for squidpy

pavanvidem:
@tcollins2011 Thanks for initiating squidpy! :rocket: 

It would be nice to implement each of the scverse packages similarly to the IUC implementation of Scanpy, i.e.,  group some related functions into a single Galaxy tool of that package. If we have a collection of tools for squidpy functions, then we can also use them for other spatial data like merfish or nanostring.

Moreover, the scverse is currently developing a new datatype called [SpatialData](https://spatialdata.scverse.org/en/latest/index.html) for spatial omics. Maybe it will completely replace anndata for spatial analysis.  In that case, it will become hard to adapt such tools.

tcollins2011:
I would agree that it would be useful to separate Squidpy in a similar manner to Scanpy. Right now this first pass at squidpy is only compatible with 10x visium data, however the tool can take many more inputs and we haven't fully implemented all of its end functionality. I'll continue to try and and separate the tool out as time goes on, but I'll probably need to find a way to tackle the persistent paths necessary for the image container object to work correctly within Squidpy.

bgruening (commented):
please add a .shed.yml ... and probably a suite one so we can add more tools into this directory.

tools/squidpy/squidpy.xml
```diff
@@ -0,0 +1,154 @@
...
+            #if $genome:
+                -g ${genome}
+            #end if
+            #if $library:
+                -l ${library}
+            #else:
+                -l 'None'
+            #end if
+            -s spatial/tissue_hires_image.png 
+            -x "${prefix}"
```
bgruening, line 25: single quotes

tools/squidpy/squidpy.xml
```diff
@@ -0,0 +1,154 @@
+<tool id="squidpy" name="squidpy" version="@TOOL_VERSION@+galaxy@VERSION_SUFFIX@" python_template_version="3.5" profile="21.05">
```
bgruening, line 1:
```suggestion
<tool id="squidpy" name="squidpy" version="@TOOL_VERSION@+galaxy@VERSION_SUFFIX@" profile="21.05">
```

tools/squidpy/squidpy.xml
```diff
@@ -0,0 +1,154 @@
...
+         -a squidpy_input.h5ad
+        #if $library:
+            -l ${library}
+        #else:
+            -l 'None'
+        #end if
+    ]]></command>
+    <inputs>
+        <!-- CHANGE FORMAT LATER -->
+        <param argument="--visium" type="data" format="visium.tar.gz" label="Visium dataset tarball"/>
```
bgruening, line 41: A user probably does not know what a tarball is. Just Visum dataset might be ok.

PR #1322, opened 2023-07-31
Suggested links for the two input formats

egonw:


bgruening:
Thanks, I will care about the stricter lint errors in the coming days.

PR #1324, opened 2023-08-16
Update of label encoder to allow storing of the model

paulzierep:
Something was messed with my rebase. New clean PR.
This update enables the export of the label encoder as h5 model, that can be used to encode new data as well, similarly to the use of pretrained models.


paulzierep:
@bgruening @anuprulez how should the tool version be bumped here ? These tools do not follow the `@TOOL_VERSION@+galaxy@SUFFIX_VERSION@` logic, they all use the `@VERSION@` tag set in the main_macros.xml ?

bgruening:
Hardcode it without macro. We need to shift to IUC standards at some point.

anuprulez:
@paulzierep I followed `version="@VERSION@" profile="@PROFILE@"` pattern while upgrading ML tools to 0.10.0. But, it should not make much difference at the moment if they will be changed according to IUC standards :)

paulzierep:
added the `VERSION_SUFFIX` can this be merged @anuprulez @bgruening

anuprulez:
I think there are a few linting errors that need to be fixed

PR #1327, opened 2023-09-04
Update AGAT tool: v1.2.0

gallardoalba:


tools/agat/agat.xml
```diff
@@ -73,18 +80,74 @@
...
+                --fasta $ref_genome 
+                --config $agat_configfile
+                --output 'output' &&
+                cat 'output' > '${annotation}'
         #else if $tool.selector == 'complement'
             @input_annotation_double@
-            agat_sp_complement_annotations.pl --ref $input1 --add $input2 --size_min $tool.size_min --output 'temp_output' &&
-            cat 'temp_output' > '${annotation_gff}'
+            agat_sp_complement_annotations.pl 
+                --ref $input1
```
bgruening, line 92: single quotes please everywhere

gallardoalba: I think in this case it is not necessary because it is pre-processed here https://github.com/bgruening/galaxytools/blob/61aade39ab9281b579857cf5168d07b0a3a75506/tools/agat/macros.xml#L84

bgruening (approved):

PR #1331, opened 2023-09-05
Add orthologr

gallardoalba:


bgruening:
Why docker and not conda?

gallardoalba:
> Why docker and not conda?

Too many dependencies I think, but you can check:

`if (!requireNamespace("BiocManager", quietly = TRUE))
    install.packages("BiocManager")
BiocManager::install()
# Install package dependencies
BiocManager::install(c("Biostrings", "GenomicRanges", "GenomicFeatures", "Rsamtools", "rtracklayer"))

# install orthologr from GitHub
devtools::install_github("HajkD/metablastr")

# install orthologr from GitHub
devtools::install_github("HajkD/orthologr")`

bgruening:
Only two of them are not in conda yet isn't it?

gallardoalba:
> metablastr

Yes, do you think it is better to create those recipes?

bgruening:
Yes. We will get the a container for free and it provides us more freedom during deployment

PR #1333, opened 2023-09-19
paired_interleaved

TomHarrop:
Hi, salmon quant is failing on GA with paired_interleaved input. I guess it's not used very often but it seems to just be quoting, so I fixed it and added a test.


bgruening (approved):
Cool, thanks Tom!

PR #1334, opened 2023-09-21
Fix ML visualisation tool reg confusion matrix

anuprulez:
I think these changes got missed during the upgrade to Galaxy ML 0.10.0
ping @bgruening


anuprulez:
All tests passed ping @bgruening thanks!

PR #1336, opened 2023-10-12
hifiasm: version bump 0.19.7, add scaff options to wrapper

abueg:
hello! 👋🏼 

version bumping hifiasm to 0.19.7 which adds the `--dual-scaf` and `--scaf-gap` parameters as options. i've added these to the wrapper, too. i'm not sure what to do about tests, i tried running hifiasm on the existing test input files both with and without the `--dual-scaf` option and the results don't change 🤔


bgruening (approved):

bgruening:
Amazing as always!

PR #1316, opened 2023-06-27
SortMeRNA: update version 4.3.6

gallardoalba:
Main changes:
- Log parameter removed (it is not available in the new version)
- Merge/unmerge bash scripts replaced by equivaled functionalities (--out2)
- New option included: --sout (separate paired and singleton aligned reads)
- Indexdb_rna removed (included in the main function)
- Include the possibility of using interleaved reads (-paired option)

PR #1338, opened 2023-10-19
sortmerna: finish update

bernt-matthias:
followup on https://github.com/bgruening/galaxytools/pull/1316 which was not deployed

ping @bgruening 

- use the same chached data for chached and cached_to_index i.e. now they differ only in that the later uses the dbprep macro

  if I get it right previously the datamanager precomputed indexes which could be used. this seems not possible anymore

  I suggest to leave the dm untouched (than the provided data will also work for old sortmerna versions)

- fix usage of cached data (did not work for multiple provided values) and add tests


bgruening:
Tests are failing here now :(

bgruening (approved):

bernt-matthias:
Hrm. toolshed is in bad mood today. 

```
Unexpected HTTP status code: 400: {"err_msg": "You do not have permission to update this repository."}
Unexpected HTTP status code: 403: {"err_msg": "You are not the owner of this repository, so you cannot administer it.", "err_code": 403005}
```

PR #1339, opened 2023-10-20
hifiasm: expose `--trio-dual` parameter and parental read list inputs for trio mode

abueg:
hello!! 👋🏼 

i have made edits to the hifiasm wrapper to expose options that have been available, just not available in the wrapper yet:
1) the `--trio-dual` parameter, which should only be an option when running in trio mode
2) the `-3` and `-4` input options, which allow the user to provide partitioned read lists as parental information, instead of the parental raw data. to add this option i restructured the tool somewhat so that now there is a reads-vs-list conditional inside the trio mode. i've restricted the list inputs to be one file only; primarily because i can't really imagine a scenario where one would end up generating multiple read list files, but also because since these are just plain text files of FASTQ header names, it should be easy for the user to just concatenate them if need be

i've checked out the various options using planemo serve and things look to be in working order... please have a look when you have a chance 🚀


abueg:
EEP did not update the trio test case to select reads, my bad, will fix

bgruening:
@abueg do you have a test (file) for the new list functionality?

bgruening:
We could also argue that if you have this file with header names it should be easy to construct fastq files base on the headers and keep the functionality out of this tool. So just recommend users to construct fastq files externally.

bgruening:
The PR looks great to me, besides the missing test and the theoretical UX question. Thanks!

abueg:
> @abueg do you have a test (file) for the new list functionality?

will make one now 👍🏼 

> We could also argue that if you have this file with header names it should be easy to construct fastq files base on the headers and keep the functionality out of this tool. So just recommend users to construct fastq files externally.

oh wait, sorry, i was unclear in describing the list files -- they're lists of fastq headers for the CHILD reads, saying that abc read should be maternal, xyz read should be paternal, and so forth. so even if you split the child reads into those binned fastq, that doesn't work as an input to hifiasm, you have to give it the whole child reads and then these lists so it knows how to partition them. the list files aren't for editing the parental files, sorry if i made it sound like that!

it's a useful feature for when there's some mis-binning after the usual process (yak count -> hifiasm trio, which does the binning using those yak counts), because it bypasses the yak count step and tells hifiasm explicitly how to bin things.... does that make sense :face_with_spiral_eyes:

bgruening:
Ah I see. Thanks for explaining!!!

bgruening (approved):

abueg:
ahhh S.O.S. @bgruening , sorry i'm not sure what needs to be changed for the test... so i have the test right now set up to check the log to make sure it says "read n reads from the list" to make sure the list input is working, but i think maybe i am specifying the test parameters incorrectly? they look like the way the other trio test is structured, though, so i am a bit confused....

image

tools/hifiasm/hifiasm.xml
```diff
@@ -428,9 +454,12 @@
             <param name="filter_bits" value="0"/>
             <conditional name="mode">
                 <param name="mode_selector" value="trio"/>
-                <param name="reads" value="child.fasta.gz"/>
-                <param name="hap1_reads" value="paternal.fasta.gz"/>
-                <param name="hap2_reads" value="maternal.fasta.gz"/>
+                <conditional name="trioinput">
+                    <param name="reads" value="child.fasta.gz"/>
```
bgruening, line 458: reads needs to be in the outer / one level above conditional?

tools/hifiasm/hifiasm.xml
```diff
@@ -556,6 +585,27 @@
...
             <output_collection name="bin_files" type="list" count="3" />
         </test>
+        <!-- TEST 15: Test trio LIST mode -->
+        <test expect_num_outputs="6">
+            <param name="filter_bits" value="0"/>
+            <param name="log_out" value="yes"/>
+            <conditional name="mode">
+                <param name="mode_selector" value="trio"/>
+                <conditional name="trioinput">
+                    <param name="reads" value="child.fasta.gz"/>
```
bgruening, line 595:
```suggestion
                <param name="mode_selector" value="trio"/>
                <param name="reads" value="child.fasta.gz"/>
                <conditional name="trioinput">
```

tools/hifiasm/hifiasm.xml
```diff
@@ -175,11 +188,24 @@
...
+                        <option value="reads">Parental reads (FASTQ files, gzipped or otherwise)</option>
+                        <option value="lists">Lists of reads assigned by parent (text files)</option>
+                    </param>
+                    <when value="reads">
+                        <param name="hap1_reads" type="data" format="fastq,fastq.gz" multiple="true" label="Haplotype 1 reads" />
+                        <param name="hap2_reads" type="data" format="fastq,fastq.gz" multiple="true" label="Haplotype 2 reads" />
+                    </when>
+                    <when value="lists">
+                        <param name="hap1_list" type="data" format="txt" label="Haplotype 1 read list" />
+                        <param name="hap2_list" type="data" format="txt" label="Haplotype 2 read list" />
```
bgruening, line 202:
```suggestion
                        <param name="hap1_list" type="data" format="txt,tabular" label="Haplotype 1 read list" />
                        <param name="hap2_list" type="data" format="txt,tabular" label="Haplotype 2 read list" />
```

bgruening: Just in case a user specifies this file as a one column tabular file

tools/hifiasm/hifiasm.xml
```diff
@@ -428,9 +454,12 @@
             <param name="filter_bits" value="0"/>
             <conditional name="mode">
                 <param name="mode_selector" value="trio"/>
-                <param name="reads" value="child.fasta.gz"/>
-                <param name="hap1_reads" value="paternal.fasta.gz"/>
-                <param name="hap2_reads" value="maternal.fasta.gz"/>
+                <conditional name="trioinput">
+                    <param name="reads" value="child.fasta.gz"/>
+                    <param name="trio_input_selector" value="reads"/>
```
bgruening, line 459:
```suggestion
                <param name="trio_input_selector" value="reads"/>
                <conditional name="trioinput">
                    <param name="reads" value="child.fasta.gz"/>
```

tools/hifiasm/hifiasm.xml
abueg: yeah, so that's what i thought too when i was initially fixing test 5 LOL but then it didn't work that way, so i made it this way... and now it works inside the conditional in test 5, but not in test 15. but in test 15, it DOES work outside the conditional: 

image

abueg: (both views are the same file just at different positions)

abueg: let me add in your suggestions though!! 👍🏼

bgruening (approved):

bgruening:
Looks like tests are passing! It will be installed during the weekend.

abueg:
YAAAY thank you so much!!

PR #1342, opened 2023-10-31
Add biotools xref to MAFFT

tuncK:


tools/mafft/macros.xml
```diff
@@ -2,6 +2,11 @@
 <macros>
     <token name="@VERSION@">0</token>
```
bebatut, line 3: It would be good to increate the wrapper version there

bebatut (commented):
Thanks @tuncK

tools/mafft/macros.xml
tuncK: Exactly what I was asking Paul. I was in a dilemma given nothing else changes.

tuncK: Thanks, yes. Done.

tuncK:
Results of the pre-existing test #3 passes varies from execution to execution if multi-threading is used. By default it passes during local testing (GALAXY_SLOTS=1), whereas GH testing routing uses 2 threads. Updated the checks to tolerate such differences between different iterations.

bgruening (approved):

PR #1344, opened 2023-10-31
Add bio.tools id for reago

J-Swang:


bgruening:
The linter says:

> .. WARNING: Best practice violation [xrefs] elements should come before [requirements]

Can you change this please?

J-Swang:
I changed both minipolish.xml and macros.xml cause I wasn't which is the right fix for this.

bgruening:
For minipolish you could just move the `    <expand macro="biotools"/>` up to that it comes before the requirements expand. And then revert your last commit

PR #1345, opened 2023-10-31
tool to query ftplinks from bioimage archive

sunyi000:
A tool to query ftplinks for study from bioimage archive

inputs: single study name or multiple separated by comma
output: file with a list of ftplinks
@beatrizserrano


tools/image_processing/bia-ftplinks/ftpLink.txt
```diff
@@ -0,0 +1,2 @@
+null
```
bgruening, line 1: is this correct?

tools/image_processing/bia-ftplinks/biaftplink.xml
```diff
@@ -0,0 +1,41 @@
+<tool id="bia_ftplink" name="FTP Link for Bioimage Archive" version="@VERSION@+galaxy1">
+    <description>Get FTP link for a study from Bioimage Archive</description>
+    <macros>
+        <import>macros.xml</import>
+    </macros>
+    <expand macro="requirements" />
+    <command detect_errors="aggressive">
+	    <![CDATA[
+		#for $study in $study_name.split(","):
```
bgruening, line 9: I guess this tool is a good step forward, but wouldn't it be better to query the BIA API?

tools/image_processing/bia-ftplinks/ftpLink.txt
sunyi000: ah, it shouldn't be there. I removed it.

tools/image_processing/bia-ftplinks/biaftplink.xml
sunyi000: i thought the curl command actually queries the api..I haven't found other api documentation yet..

beatrizserrano: The BIA API seems to be the BioStudies one: https://www.ebi.ac.uk/biostudies/help

bgruening: Yes, its an API. Sorry I was not concrete here. I think its a good first step. I was just wondering if there is no API abstraction library that we should use. But maybe there is none.

sunyi000: thanks both. Yes I found that biostudies API. It seems more complicated than just a tool for getting FTP links.

tools/image_processing/bia-ftplinks/biaftplink.xml
```diff
@@ -0,0 +1,41 @@
...
+        <test>
+            <param name="study_name" value="S-BIAD570" />
+            <output name="ftp_link" ftype="txt" file="ftpLink_single_value.txt" lines_diff="0" />
+        </test>
+
+    </tests>
+    <help>
+        <![CDATA[
+            **What it does**
+	    This tool finds the ftp link for a study in the Bioimage Archive
```
bgruening, line 36:
```suggestion
	    This tool finds all FTP links associated to a study in the Bioimage Archive.
```

bgruening (commented):
Looks good to me @sunyi000

One question, what should a user do with the FTP links? Should this tool offer at least the option to download them and store them in a collection?

tools/image_processing/bia-ftplinks/macros.xml
```diff
@@ -0,0 +1,21 @@
+<macros>
+    <token name="@VERSION@">0.1.0</token>
+    <xml name="requirements">
+        <requirements>
+	    <requirement type="package" version="1.20.3">wget</requirement>
```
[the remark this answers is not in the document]
sunyi000: ah, initially I was thinking of after getting the ftp links then download all the images using wget -r. so the tool will be about download images from bia.

Then I see there is already a lftp tool so I removed the download images part. 

what do you suggest? A tool download images from bia,or a tool like this just output ftp links?

Thanks

sunyi000:
> Looks good to me @sunyi000
> 
> One question, what should a user do with the FTP links? Should this tool offer at least the option to download them and store them in a collection?

Do you mean offering an option to download images with the links? For example,if it is checked,then the output of this tool will be ftp links and images. If it is unchecked,just output links? 
I think this maybe more useful?

bgruening (approved):

bgruening:
Thanks @sunyi000!

Do you want to create a PR to add this tool to the EU instance as well? https://github.com/usegalaxy-eu/usegalaxy-eu-tools

sunyi000:
> Thanks @sunyi000!
> 
> Do you want to create a PR to add this tool to the EU instance as well? https://github.com/usegalaxy-eu/usegalaxy-eu-tools

I got some permission error when trying to open PR on that repo

PR #1356, opened 2023-11-09
tool for wound healing scratch assay image analysis

sunyi000:
a tool for the workflow here
https://git.embl.de/grp-cba/wound-healing-htm-screen

basically it's a wrapper of the groovy script.
@tischi


sunyi000:
@bgruening the test files are larger than 1M in test-data folder, should I use the "location" tag instead? Thank you

bgruening:
Please try the new fancy location tag :) and lets see how good those are working :)

(we might need to update our CI here - copy from IUC)

tools/image_processing/woundhealing/woundhealing.xml
```diff
@@ -0,0 +1,65 @@
+<tool id="woundhealing" name="Wound healing scratch assay image analysis" version="@VERSION@+galaxy1">
```
bgruening, line 1: should we make this tool id a bit more specific? `fiji_woundhealing` or `img_wouldhealing`?

tools/image_processing/woundhealing/woundhealing.xml
```diff
@@ -0,0 +1,65 @@
...
+	    </conditional>
+	    <param name="dataset_id" value="A3ROI2_Slow" />
+	    <output name="tif_output" file="A3ROI2_Slow.tif"/>
+	    <output name="csv_output" file="A3ROI2_Slow.csv"/>
+        </test>
+    </tests>
+    <help>
+        <![CDATA[
+            **What it does**
+            Automated quantification of wound healing in high-throughput time-lapse transmission microscopy scratch assays.
```
bgruening, line 61: Can we elaborate here more?

E.g. the csv is not very self explanatory.

tools/image_processing/woundhealing/woundhealing.xml
```diff
@@ -0,0 +1,65 @@
...
+            <when value="yes">
+                <param name="input_images" type="data" format="tar" label="A tarball of images" />
+            </when>
+        </conditional>
+        <param name="dataset_id" type="text" label="Dataset ID" />
+        <param name="threshold"  type="text" label="CoV threshold (-1: auto)" value="-1"/>
+    </inputs>
+    <outputs>
+        <data name="tif_output" format="tif" from_work_dir="output/*.tif" label="movie output ${on_string}" />
+	<data name="csv_output" format="tabular" from_work_dir="output/*.csv" label="tabular output"/>
```
bgruening, line 45: I prefer this to be a tabular - tab separated - because galaxy has more tools to deal with those.
But the file that is created is a comma-seperated file, isn't it?

You could also add the columns names as metadata: https://github.com/galaxyproject/tools-iuc/blob/3a3dfd676076438bea1518eb731790edfb4da3c0/tools/onto_toolkit/get_term_id_vs_term_name.xml#L16

bgruening (commented):
* you could also move this tools into the imagej folder if you like

Very good @sunyi000!

tools/image_processing/woundhealing/woundhealing.xml
sunyi000: - yes, the output file created by fiji is a comma-separated file.

- I have tried the 'location' tag on `tests->test->param` . planemo lint told me that the `location `tag is not allowed, however, the xml ref seems to allow it. My planemo is up to date, v0.75.14. not sure if it's to do with the ci update

- I also have question about multiple urls as value of the location tag. If the input `multiple=true`, and in the test, I assume I will need to separate multiple urls by comma? The only example I can find its the `mzmine_batch` tool in iuc repo, but they are commented out.
 
thanks, I will work on all your suggestions above, and come back later.

bgruening: You really need to have the latest dependencies of planemo. The location attributes are only in release_23.1.
Try updating your galaxy-tool-utils package.

I assume its comma separated. But haven't tried myself yet.

sunyi000: thanks.. i tried the update of galaxy-toolutils to 23.1.1, but the latest planemo 0.75.17 and the dev 0.75.18.dev0 does not support it yet.. i guess we simply need to wait a few more days, in this case?

bgruening: Yeah, then we need to wait a bit.

bgruening:
>  .. WARNING: Missing shed metadata field [long_description] for repository
.. WARNING: Missing shed metadata field [homepage_url] for repository

Is the shed linting error.

bgruening:
This looks good now, but you need to rebase. Something went wrong in the PR.

sunyi000:
> t R s

I have merged master into wound-healing branch

PR #1357, opened 2023-11-14
Add TAPscan

shiltemann:


shiltemann:
@deeptivarshney just FYI, I moved the tool here, in case you want to review or make changes in the future

bgruening (approved):

Deeptivarshney:
Hi @shiltemann , looks good, thank you. :)

shiltemann:
@bgruening ok let's just merge this, I will create another PR when I get the updated tapscan script from Romy

bgruening (commented):
Do you want to add a creator tag. https://docs.galaxyproject.org/en/latest/dev/schema.html#tool-creator-organization

NFDI4Plants or your name?

Any EDAM terms which we can include?

bgruening:
Thanks!

bgruening:
In a later version if would be nice if [tapscan_domains_v12.txt](https://github.com/bgruening/galaxytools/pull/1357/files#diff-5bbf8bdb8bb735deb6b78ed177a3ff2795ea98f7794457bf194d2a5a1ba80ef7) could be compressed and the tool could load the compressed DB.

shiltemann:
> In a later version it would be nice if [tapscan_domains_v12.txt](https://github.com/bgruening/galaxytools/pull/1357/files#diff-5bbf8bdb8bb735deb6b78ed177a3ff2795ea98f7794457bf194d2a5a1ba80ef7) could be compressed and the tool could load the compressed DB.

It looks like HMMer accepts gzipped out of the box, so can do that now :+1:

shiltemann:
would be nice if we could get this merged and installed on Galaxy before our talk on the 23rd, I will make a follow up PR with the latest version of the script soon then.

PR #1359, opened 2023-11-24
Update Nanopolish to v0.14

tuncK:
I attempted to address the following:

a) Existing version provides incorrect result (0 usable event detected) on a recent in-house project. Allegedly an issue that is nanopolish version-dependent.
b) Current wrapper does not explicitly handle input.fast5.tar.xz as input.
c) Current wrapper fails linting.

### DIFF
--- a/tools/nanopolish/macros.xml
+++ b/tools/nanopolish/macros.xml
@@ -22,6 +22,52 @@
             <output name="output_index_readdb" file="reads.fasta.index.readdb" />
  -->
 
+
+    <token name="@PREPROCESS_INPUTS@"><![CDATA[
+        ln -s '$input_merged' reads.fasta && 
+ 
+        mkdir fast5_files &&
+        #if $input_reads_raw.extension == 'fast5':
+             ln -s '$input_reads_raw' fast5_files/read1.fast5 &&
+
+        #else if $input_reads_raw.extension == 'fast5.tar':
+            ln -s '$input_reads_raw' fast5_files.tar &&
+            tar -xf fast5_files.tar -C fast5_files &&
+
+        #else if $input_reads_raw.extension == 'fast5.tar.bz2':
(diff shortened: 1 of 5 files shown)


bgruening (commented):
Mh, I don't think Galaxy knows about the filetype fast5.tar.xz

https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/config/sample/datatypes_conf.xml.sample#L355

bgruening:
Otherwise this looks very cool!

tuncK:
So what happens if I provide a x.tar.xz? I have such files in my history and no error was raised.

bgruening:
But as you see its not recognized as tar.xz datatype. I guess we would need to add this compression to Galaxy.

tuncK:
OK, so I did a quick test by manually uploading files. gz is auto-decompressed whereas xz is not recognised at all. This is unlike the case I mentioned above, where the contents of the tarball were recognised properly. 
image

So how should I proceed in this case?
a) I should Eliminate .xz from this tool input and you merge as is?
b) We add .xz to allowed filetypes. How?

bgruening:
Ideally, we add a new datatype similar to here: https://github.com/galaxyproject/galaxy/commit/f07c6a98431f1a4c2a2d81f99d3b54a23f0a901e a new class essentially.

Do you know if tools understand this tarball natively?

tuncK:
Is this what you mean https://github.com/tuncK/galaxy/tree/tarxz? Where should I be sending the PR to: usegalaxy-eu/galaxy or galaxyproject/galaxy?

No, I dont think the tool itself supports any compression whatsoever, it only takes paths to fast5 files. That is why this wrapper needs to decompress them manually first.

bgruening:
Yes! Pretty much!

Please sent this to upstream Galaxy and the release_23.1 branch

bgruening (approved):

bgruening:
ofc :)

PR #1360, opened 2023-11-30
Add RBPBench tool wrapper

michauhl:
Hi,
this is a Galaxy wrapper for the tool RBPBench ( https://github.com/michauhl/RBPBench ). Goal is to bring it to usegalaxy.org.
Best,
Michael


michauhl:
Weird, in my local test environment all 8 tests (planemo test .) pass. My environment:
```
conda create -n rbpbench-planemo -c conda-forge -c bioconda python=3.9
conda activate rbpbench-planemo
pip3 install planemo
```

pavanvidem:
is it possible to move the batch_table_wrapper.py to the original tool instead of keeping it in the Galaxy wrapper? It will be easier to maintain on the tool side than on the Galaxy side.

tools/rna_tools/rbpbench/rbpbench.xml
```diff
@@ -0,0 +1,1040 @@
+<tool id="rbpbench" name="RBPBench" version="0.1.0+galaxy0" python_template_version="3.5" profile="21.05">
+
+    <description>- Evaluate CLIP-seq and other genomic region data using a comprehensive collection of RBP binding motifs</description>
+    <requirements>
+        <requirement type="package" version="0.6">rbpbench</requirement>
```
pavanvidem, line 5:
```suggestion
        <requirement type="package" version="0.6">rbpbench</requirement>
        <requirement type="package" version="5.3">meme</requirement>
```

michauhl: How can I see the test output here? Cannot find it ... I could add meme >=5.0 in the bioconda recipe (https://github.com/bioconda/bioconda-recipes/blob/143a88e35c4f8cce93c4c7d335514e3afb4ce42f/recipes/rbpbench/meta.yaml#L28C9-L28C9). If I only update this, I need to increment "number:" to 1 right? (still keeping version 0.6)

michauhl: Ok, I did some update: https://github.com/bioconda/bioconda-recipes/pull/44566 :D

pavanvidem: > How can I see the test output here? Cannot find it ... I could add meme >=5.0 in the bioconda recipe (https://github.com/bioconda/bioconda-recipes/blob/143a88e35c4f8cce93c4c7d335514e3afb4ce42f/recipes/rbpbench/meta.yaml#L28C9-L28C9). If I only update this, I need to increment "number:" to 1 right? (still keeping version 0.6)

Yes, increase the build number.

pavanvidem: @michauhl  wait, I think we have to update it again. Does it work with versions above 5.5.4?

pavanvidem:
Tests are passing. Maybe @michauhl will move the python script into the tool. Then we are ready to go.

pavanvidem:
@michauhl will `rbp_ids.catrapid.omics.v2.1.human.6plus.txt` stay as it is forever or will it change for some reason in the future? Like more proteins? If it is changing based on the tool updates, then it should also go into the tool. It is probably not optimal to maintain that file as tool-data.

michauhl:
> @michauhl will `rbp_ids.catrapid.omics.v2.1.human.6plus.txt` stay as it is forever or will it change for some reason in the future? Like more proteins? If it is changing based on the tool updates, then it should also go into the tool. It is probably not optimal to maintain that file as tool-data.
I don't plan on changing it. If so I would rather add a new file, with new identifier. The user can supply a custom database of motifs anyway, or single motifs. So no problem.

tools/rna_tools/rbpbench/rbpbench.xml
```diff
@@ -0,0 +1,1003 @@
+<tool id="rbpbench" name="RBPBench" version="@TOOL_VERSION@+galaxy@VERSION_SUFFIX@" python_template_version="3.5" profile="@profile@">
+
+    <description>- Evaluate CLIP-seq and other genomic region data using a comprehensive collection of RBP binding motifs</description>
```
bgruening, line 3:
```suggestion
<tool id="rbpbench" name="RBPBench" version="@TOOL_VERSION@+galaxy@VERSION_SUFFIX@" profile="@profile@">
    <description>Evaluate CLIP-seq and other genomic region data using a comprehensive collection of RBP binding motifs</description>
```

bgruening (approved):

michauhl:
@pavanvidem @bgruening Great, thank you for your help!

PR #1362, opened 2023-12-03
update rbpbench tool-data conf

pavanvidem:
Trying to fix the deployment issue with the tool-data config. Also removed the unused `batch_table_wrapper.py`.


pavanvidem:
I did not increment the version

PR #1375, opened 2024-01-19
Add header

Marie59:
Hi @bgruening I want to add the header of the columns when selecting them for this tool. Do you think that's okay ?

### DIFF
--- a/tools/text_processing/text_processing/cut.xml
+++ b/tools/text_processing/text_processing/cut.xml
@@ -40,7 +40,18 @@
                 <option value="-b">bytes</option>
             </param>
             <when value="-f">
-                <param name="list" type="data_column" data_ref="input" use_header_names="true" multiple="true" label="List of Fields" help="(-f)" />
+                <conditional name="colnames_option">
+                    <param name="header" type="select" label="Is there a header for the data's columns ?">
+                        <option value="Y">Yes</option>
+                        <option value="N">No</option>
+                    </param>
+                    <when value="Y">
+                        <param name="list" type="data_column" data_ref="input" use_header_names="true" multiple="true" label="List of Fields" help="(-f)" />
+                    </when>
+                    <when value="N">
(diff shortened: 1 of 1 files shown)


bgruening:
I guess so. But lets ask @nsoranzo and @bernt-matthias as well.

bernt-matthias:
A good idea. Also cool to see this attribute (I never stumbled over it so far).

But I have 1 concern (and a suggestion). 

The main problem is that the input does not always has a header. For input without header the select will probably become unusable. Actually we can not even assume that the input is tabular (input has format `data`).

My suggestion would be to add a conditional that allow the user to specify if the data has a header or not and only use the attribute if the user has selected yes.

Marie59:
Is this what you were thinking of @bernt-matthias ? (I tested it locally it seems to be working fine)

nsoranzo:
> The main problem is that the input does not always has a header. For input without header the select will probably become unusable. Actually we can not even assume that the input is tabular (input has format `data`).

Actually the `input` format is `txt`, which makes me think that we should probably drop the `-b` (bytes) option (try to run `cut -b` on a text file with non-ASCII characters, it makes a mess).

Another issue is that the delimiter (`-d`) can only be specified when using `-f` (fields), e.g.:

```bash
$ cut -b 1,3-5 -d "," test.txt 
cut: an input delimiter may be specified only when operating on fields
```

So the `delimiter` param should be moved inside `<when value="-f">`.

bernt-matthias:
> Is this what you were thinking of @bernt-matthias ? (I tested it locally it seems to be working fine)

Excellent

> Another issue is that the delimiter (-d) can only be specified when using -f (fields)

I just did this. 

> we should probably drop the -b (bytes) option

Feel free to do so.

bernt-matthias (approved):
Thanks @Marie59 and @nsoranzo

Marie59:
And thanks to you @bernt-matthias !

bgruening:
Can you please look at https://github.com/bgruening/galaxytools/pull/1385 we will work on this PR as a general update. 

Thanks!

PR #1376, opened 2024-01-20
add nl

bernt-matthias:


bgruening:
You can do that with https://usegalaxy.eu/root?tool_id=toolshed.g2.bx.psu.edu/repos/devteam/column_maker/Add_a_column1/2.0

@bernt-matthias this entire folder is NOT a suite. Same for Marie's PR yesterday. Its a very old bunch of tools. Should we add new tools into this folder? 
Should we compose those tools into a suite ... even if its breaks tool versions?

bgruening (approved):

bernt-matthias:
> You can do that with https://usegalaxy.eu/root?tool_id=toolshed.g2.bx.psu.edu/repos/devteam/column_maker/Add_a_column1/2.0

I don't see how. We can just compute some formula on each line. How to get a line numbering from this?

> @bernt-matthias this entire folder is NOT a suite. Same for Marie's PR yesterday. Its a very old bunch of tools. Should we add new tools into this folder? Should we compose those tools into a suite ... even if its breaks tool versions?

Would not like to break tool lineage .. and no idea how to ensure not to break it. 

If I install [this](https://toolshed.g2.bx.psu.edu/view/bgruening/text_processing/d698c222f354 ) I just get all tools. This is sufficient for me.

bgruening:
Can you please look at https://github.com/bgruening/galaxytools/pull/1385 we will work on this PR as a general update. 

Thanks!

PR #1381, opened 2024-02-01
Cool to hic

fubar2:
Will this do for a start @bgruening 
Sorry it took so long - I was testing a bad cool file. Once I found a valid one, it worked fine.
uses hictk
needs logic to consider case of mcool 
hictk cl expects things to have informative extensions.
must take a look at the python bindings - probably more convenient..
Output works with JBrowse2

image


fubar2:
gak - made a mess here. will redo

PR #1391, opened 2024-03-01
Add tool for analysing images using Bioimage AI models

anuprulez:
images

Test files: 
https://bioimage.io/#/?id=10.5281%2Fzenodo.5764892
https://zenodo.org/api/records/6647674

Remaining tasks:

- [x] Support for creating the original predicted image matrix
- [x] Add tool tests (imaging models are large (> 50 MB and sometimes as large as 500 MB) and it is unclear how to add sample data for tests)
- [x] Take input asTiff files if possible
- [x] Test with models of different input/output dimensions (3,4,5,...)
- [ ]  Add support for TensorFlow


tools/bioimaging/bioimage_inference.xml
```diff
@@ -0,0 +1,63 @@
...
+<![CDATA[
+        python '$__tool_directory__/main.py'
+
+            --imaging_model '$input_imaging_model'
+            --image_file '$input_image_file'
+]]>
+    </command>
+    <inputs>
+        <param name="input_imaging_model" type="data" format="zip" label="BioImage model" help="Please load a BioImage model from file uploader"/>
+        <param name="input_image_file" type="data" format="npz" label="Image to be analysed" help="Please provide an image"/>
```
bgruening, line 20: NPZ is numpy isn't it? How do people create such image? Should we create an NPZ file from a tiff, png as part of this tool?

tools/bioimaging/bioimage_inference.xml
```diff
@@ -0,0 +1,63 @@
+<tool id="bioimage_inference" name="Load model from BioImage and make inferences" version="0.0.5">
+    <description>using images</description>
```
kostrykin, line 2:
```suggestion
    <description>using images</description>
    <edam_operations>
        <edam_operation>operation_3443</edam_operation>
    </edam_operations>
    <xrefs>
        <xref type="bio.tools">pytorch</xref>
        <xref type="biii">pytorch</xref>
    </xrefs>
```

tools/bioimaging/bioimage_inference.xml
```diff
@@ -0,0 +1,63 @@
+<tool id="bioimage_inference" name="Load model from BioImage and make inferences" version="0.0.5">
```
kostrykin, line 1:
```suggestion
<tool id="bioimage_inference" name="Process image using model from BioImage Model Zoo" version="0.0.5">
```
Does a typical user of this tool know what "inference" is? Just a suggestion, maybe something less technical?

anuprulez: thank you @kostrykin for your suggestions

tools/bioimaging/bioimage_inference.xml
FynnBe: chiming in on this on @beatrizserrano 's request:
(disclaimer: I'm unaware of the context of this PR, but can report from the bioimage.io side: We demand `.npy` test files to be provided alongside every model contributed to bioimage.io. Among the utility functions we provide in bioimageio.core there is
https://github.com/bioimage-io/core-bioimage-io-python/blob/f798344213c179a6c836938aff9e4f6c46f8d23c/bioimageio/core/utils/image_helper.py#L136
 -- a util function using imageio to attempt to load any given image and then attempting to guess it's axes, see 
 https://github.com/bioimage-io/core-bioimage-io-python/blob/f798344213c179a6c836938aff9e4f6c46f8d23c/bioimageio/core/utils/image_helper.py#L42
 
Software using the model zoo (or even specifically bioimageio.core) should try to delegate this guesswork to the end user instead.

note: the links above are from a development branch. These updated functions will be available in the coming bioimageio.core release, currently this functionality is implemented [here](https://github.com/bioimage-io/core-bioimage-io-python/blob/main/bioimageio/core/image_helper.py)

ctr26: Would generating the .npy files/objects at validation time using the spec be sensible/viable?

bgruening:
@anuprulez is this still WIP or can we merge it. @beatrizserrano needs it :)

kostrykin:
Guys, my impression of the comments above is that there has been some confusion. So in order to clear things up, I want to try to summarize the main concerns:

## Concern 1: Support of TIFF/PNG

@bgruening was addressing the **input** file format, which currently is only NPZ. For those people who are the target audience of this tool, this isn't a well-established standard file format such as TIFF or PNG. However, @anuprulez explained the conversion to TIFF or PNG of the **output** files, pointing out that it is not straightforward.

Lets consider the questions of the **input file formats** and the **output file formats** separately.

**1.1) Inputs:** The NPY/NPZ formats are more general than TIFF and PNG, since they can store arbitrary numpy arrays, with arbitrary data types (even mixed) and number of dimensions. Thus, conversion from TIFF/PNG to NPY/NPZ should be straightforward. Given the concerns above, I think the tool wrapper really should also accept TIFF and PNG input files and do this conversion automatically. This should be as simple as something like this:

```python
im = imread('tiff or png image file path')
np.save('input.npy', im)  # to produce an NPY
np.savez('input.npz', im)  # to produce an NPZ
```

**1.2) Outputs:** It was pointed out that the intention of outputting the originally predicted arrays as NPZ was that no information is lost. However, I'm wondering whether converting the data to TIFF is even capable of losing information? My gut feeling is that using TIFF as the output file format should be safe, at least as long as we restrict the wrapper to NPY instead of NPZ (see below).

## Concern 2: NPY or NPZ?

Please keep and mind that NPY is _"the standard binary file format in NumPy for persisting a single arbitrary NumPy array on disk"_, and NPZ is _"the standard format for persisting multiple NumPy arrays on disk"_, which _can_ be compressed, but not necessarily ([docs](https://numpy.org/doc/stable/reference/generated/numpy.lib.format.html#npy-format)). The key difference here is that NPY is for single arrays, NPZ is for multiple arrays.

At this point I'm somewhat confused. I hope that @FynnBe can maybe add something regarding the following two concerns:

**2.1)** What happens if the NPZ contains more than one array? Do the bioimage.io models process them independently and yield another NPZ, which contains predictions for each array in the NPZ? Or do the models bluntly fail to process an NPZ with more than one array?

**2.2.)** As far as I understand the comment made by @FynnBe ([link](https://github.com/bgruening/galaxytools/pull/1391#discussion_r1528212236)), the bioimage.io models are only guaranteed to work with NPY inputs. Doesn't that mean, that the wrapper should actually take/produce NPY instead of NPZ? If this is right, then we should forget about NPZ and restrict our discussions here to PNG, TIFF, and NPY.

---

cc @beatrizserrano

anuprulez:
I was on holidays. I will have a look at it this week. Can you let me know when you require this feature, I can prioritise it @beatrizserrano thanks!

@kostrykin thank you for the clarification.

beatrizserrano:
Thank you @anuprulez! 🤗 

We want to develop tutorials at the Biohackathon in November, so it would be perfect to be able to test the tool well before that. Thanks again!

anuprulez:
@kostrykin thank you for the explanation of NPY and NPZ. 

I would like to elaborate the issue of using TIFF or PNG files as input. When these files are read (using OpenCV), they are represented as 2 dimensional image.

Example from [Neuron Segmentation in EM model](https://bioimage.io/#/?id=10.5281%2Fzenodo.5874741)

TIFF input file shape: `torch.Size([360, 360])` 

However, the NPY file has shape `torch.Size([1, 1, 32, 256, 256])`

When I use the TIFF file as the test input for prediction, I receive the following error:

```
RuntimeError: Expected 5-dimensional input for 5-dimensional weight [64, 1, 3, 3, 3], but got 2-dimensional input of size [360, 360] instead
```

The model expects an input with 5 dimensions but the TIFF file is only 2 dimensional. For other models, the input shape varies - sometimes it is 3,4 or 5. When I use the NPY file as input, it works fine and the model predicts the segmentations.

As suggested by @FynnBe, I tried to use `imageo` and `bioimageo` packages and the suggested converter methods, but TIFF file dimensions don't match that of the respective model.

The issue is to find a way to transform TIFF or PNG images to be represented in the valid dimensions so that the respective AI model can consume the input files and produce desired output. 

Any suggestions here would be really helpful!

Thank you!

@beatrizserrano

kostrykin:
Thanks for the follow-up @anuprulez 

I see the challenge that comes with conversion of PNG/TIFF to NPY. However, by accepting only NPY and dropping support for PNG/TIFF, we would simply shift that challenge to the user of the tool, which IMO we shouldn't do.

The core issue is to determine the _expected_ number of dimensions of the input, the conversion is simply some reshaping, right? I see couple of options here, in the order of preference:

**Option 1.** My suspicion is that the only way to get the dimensions of the TIFF right _automatically_ is to rely on what the model _declares_ as the required input dimension — if it does declare anything. You do

https://github.com/anuprulez/galaxytools/blob/0dfdd4d94650bc58405b46944a929fcc8d66c6ac/tools/bioimaging/main.py#L156

to load the model, is there anything like `model.input_ndim` defined? I'm not familiar with the BIA AI models, but unfortunately, [torch.load](https://pytorch.org/docs/stable/generated/torch.load.html#torch-load) does not specify anything useful for what it returns.

**Option 2.** You quoted the error _"RuntimeError: Expected 5-dimensional input for 5-dimensional weight [64, 1, 3, 3, 3], but got 2-dimensional input of size [360, 360] instead"_, where is that raised from? By looking into the code for where this check is being performed, we might be able to reverse-engineer how the _expected_ dimension of the models is stored.

**Option 3.** In the worst-case (i.e. if Option 2 leads to no solution for some reason), I'd fallback to the documentation, like: Point the user to the documentation of the models, where, hopefully, the required number of dimensions is stated. Add an input field to the tool UI, where the user has to write down the number that was looked up from the model documentation.

anuprulez:
@kostrykin thank you for the interesting pointers. I will explore how to get the input shape of the models dynamically.

anuprulez:
I have updated the tool to take TIF files as input (also taking the nested images inside a TIF file) and produce the predictions as TIF and NPY files. The tool uses three inputs:

- Pytorch models as `weights-torchscript.pt`
- Input image as TIF 
- Input image dimensions. This information is necessary for the model to reshape the input image correctly. This input size is available for each model in its RDF.yaml file (config > test_information > inputs > size). For example, for the Nuclei Segmentation Boundary Model, this value is `256 x 256 x 1 x 1` which should be input as `256,256,1,1`

Outputs:
- TIF file
- NPY file

I have tested this tool with sample TIF images present for each model on https://bioimage.io/#/

Tested models (only Pytorch models):

- Nuclei segmentation boundarymodel
- Livecell segmentation boundarymodel
- Neuron Segmentation EM
- Mitochondria EM Segmentation BoundaryModel
- Platynereisem nuclei segmentation boundarymodel
- Hpa-cell-segmentation-(dpnunet)
- Enhancer Mitochondria EM 2D
- Enhancer Mitochondria EM 3D

It would be great if some of you can test this tool and provide feedback :)

Currently, it does not support TensorFlow models. First I will do a proof of concept (e.g. how to correctly reconstruct them for prediction)  and then maybe add this feature later.

@beatrizserrano @kostrykin @bgruening Thanks a lot!!

kostrykin (commented):
Much thanks @anuprulez!

There are a few minor comments inside.

tools/bioimaging/main.py
```diff
@@ -0,0 +1,78 @@
...
+    pred_data_output = pred_data.detach().numpy()
+
+    # save original image matrix
+    np.save("output_predicted_image_matrix.npy", pred_data_output)
+
+    # post process predicted file to correctly save as TIF file
+    pred_data = torch.squeeze(pred_data)
+    pred_numpy = pred_data.detach().numpy()
+    pred_numpy = pred_numpy * 255
+    pred_numpy = pred_numpy.astype(np.uint8)
```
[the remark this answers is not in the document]
anuprulez: the original predicted values are floats (< 1). When such a matrix is saved as a TIF file, it does not show anything after opening (normalised).

output_empty

This issue I have seen in the `sample_output.tif` files for a few BioImage models.

empty_tiff

To avoid it, I multiply it by 255 and save it as a matrix of Int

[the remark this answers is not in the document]
kostrykin: I'm pretty sure that this is an issue with your image viewer:

> In [1]: import imageio
> 
> In [2]: import numpy as np
> 
> In [3]: im1 = np.random.rand(50,50)
> 
> In [4]: im1.min(), im1.max()
> Out[4]: (0.000287018172377973, 0.999300887698757)
> 
> In [5]: imageio.v3.imwrite('test.tif', im1)
> 
> In [6]: im2 = imageio.v3.imread('test.tif')
> 
> In [7]: np.allclose(im1, im2)
> Out[7]: True

tools/bioimaging/bioimage_inference.xml
```diff
@@ -0,0 +1,74 @@
...
+**Description**
+The tool takes a BioImage model and an image to be analysed. Model makes a prediction and the predicted image becomes available as a TIF file.
+
+**Input files**
+	- BioImage model: Add one of the model from Galaxy file uploader by choosing a "femote" file at "ML Models/bioimaging-models"
+	- Image to be analysed: Pass an image as TIF file
+	- Provide the necessary input shape for the model. This information can be found in the RDF file of each model (RDF file > config > test_information > inputs > size)
+
+**Output file**
+        - Predicted image: Predicted image using the Bioimage model
```
beatrizserrano, line 60:
```suggestion
        - Predicted image: Predicted image using the BioImage.IO model
```

tools/bioimaging/bioimage_inference.xml
```diff
@@ -0,0 +1,74 @@
...
+        - Predicted image matrix: Predicted image matrix in original dimensions
+        ]]>
+    </help>
+    <citations>
+        <citation type="bibtex">
+            @ARTICLE{anuprulez_galaxytools,
+                Author = {Anup Kumar and Beatriz Serrano-Solano and Björn Grüning},
```
beatrizserrano, line 67: @kostrykin you should be there too :)

tools/bioimaging/bioimage_inference.xml
```diff
@@ -0,0 +1,74 @@
...
+            --imaging_model '$input_imaging_model'
+            --image_file '$input_image_file'
+	    --image_size '$input_image_input_size'
+
+]]>
+    </command>
+    <inputs>
+        <param name="input_imaging_model" type="data" format="zip" label="BioImage model" help="Please load a BioImage model from file uploader"/>
+	<param name="input_image_file" type="data" format="tif" label="Image to be analysed" help="Please provide an image"/>
+	<param name="input_image_input_size" type="text" label="Shape of the input image" help="Provide input shape as a tuple. See chosen model's RDF file to find correct input size. For example: Mitochondria EM Segmentation BoundaryModel, the input size is 256 x 256 x 32 x 1. Enter the size as (256,256,32,1)."/>
```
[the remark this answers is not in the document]
anuprulez: I have removed `()` from the `help` section. It is not required while running the tool. A user only needs to provide input image size as '256,256,32,1'  for example

tools/bioimaging/bioimage_inference.xml
```diff
@@ -0,0 +1,74 @@
...
+    </help>
+    <citations>
+        <citation type="bibtex">
+            @ARTICLE{anuprulez_galaxytools,
+                Author = {Anup Kumar and Beatriz Serrano-Solano and Björn Grüning},
+                keywords = {imaging, deep learning},
+                title = {{Inference using BioImage models}},
+                url = {https://github.com/bgruening/galaxytools}
+            }
+        </citation>
```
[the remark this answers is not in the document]
beatrizserrano: Another citation could be https://doi.org/10.1101/2022.06.07.495102

tools/bioimaging/main.py
```diff
@@ -0,0 +1,78 @@
+"""
+Predict images using AI models from BioImage
```
beatrizserrano, line 2:
```suggestion
Predict images using AI models from BioImage.IO
```

tools/bioimaging/bioimage_inference.xml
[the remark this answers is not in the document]
anuprulez: Added this one also :)

tools/bioimaging/bioimage_inference.xml
anuprulez: This format of citation has been replaced by two DOIs based ones.

kostrykin: Thanks, but I think this is what the [creator fields](https://docs.galaxyproject.org/en/latest/dev/schema.html#tool-creator) are actually meant for.

tools/bioimaging/bioimage_inference.xml
```diff
@@ -0,0 +1,74 @@
...
+
+            --imaging_model '$input_imaging_model'
+            --image_file '$input_image_file'
+	    --image_size '$input_image_input_size'
+
+]]>
+    </command>
+    <inputs>
+        <param name="input_imaging_model" type="data" format="zip" label="BioImage model" help="Please load a BioImage model from file uploader"/>
+	<param name="input_image_file" type="data" format="tif" label="Image to be analysed" help="Please provide an image"/>
```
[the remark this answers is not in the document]
anuprulez: Ok, will try that. It shouldn't be that complex

tools/bioimaging/bioimage_inference.xml
```diff
@@ -0,0 +1,74 @@
...
+	<param name="input_image_input_size" type="text" label="Shape of the input image" help="Provide input shape as a tuple. See chosen model's RDF file to find correct input size. For example: Mitochondria EM Segmentation BoundaryModel, the input size is 256 x 256 x 32 x 1. Enter the size as (256,256,32,1)."/>
+    </inputs>
+    <outputs>
+	<data format="tif" name="output_predicted_image" from_work_dir="output_predicted_image.tif" label="Predicted image"></data>
+	<data format="npy" name="output_predicted_image_matrix" from_work_dir="output_predicted_image_matrix.npy" label="Predicted image matrix"></data>
+    </outputs>
+    <tests>
```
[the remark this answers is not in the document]
anuprulez: first will add support for PNG files and then add a test

tools/bioimaging/bioimage_inference.xml
anuprulez: I have added some information inside `<creator>` tag: https://github.com/bgruening/galaxytools/blob/9961ea5c779320b832154afca06432d1eea6a9d9/tools/bioimaging/bioimage_inference.xml#L7

But, I am not sure about a few fields

kostrykin (commented):
Much thanks for adding the PNG support and everything!

Just very few comments left, sorry for being a bit picky maybe.

tools/bioimaging/bioimage_inference.xml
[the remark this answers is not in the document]
anuprulez: done!

tools/bioimaging/bioimage_inference.xml
```diff
@@ -0,0 +1,86 @@
...
+            --imaging_model '$input_imaging_model'
+            --image_file '$input_image_file'
+	        --image_size '$input_image_input_size'
+
+    ]]>
+    </command>
+    <inputs>
+        <param name="input_imaging_model" type="data" format="zip" label="BioImage.IO model" help="Please upload a BioImage.IO model."/>
+	<param name="input_image_file" type="data" format="tiff,png" label="Input image" help="Please provide an input image for the analysis."/>
+	<param name="input_image_input_size" type="text" label="Shape of the input image" help="Provide shape of input image. See chosen model's RDF file to find correct input shape. For example: for the BioImage.IO model MitochondriaEMSegmentationBoundaryModel, the input shape is 256 x 256 x 32 x 1. Enter the shape as '256,256,32,1'."/>
```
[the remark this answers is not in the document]
anuprulez: Ok, I replaced the `x` symbol by `*`. For this sentence `Enter the shape as '256,256,32,1'`, I have left out any quotation or symbols to avoid confusion. It should be entered just like any other text.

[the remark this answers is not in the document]
kostrykin: @anuprulez I just looked into the RDF file because I thought it would be good to use their notation and saw that they use the `x` notation :) much sorry! I didn't know that you sticked to the RDF notation… so actually using the `x` notation which you used originally, then is the best option, I think.

tools/bioimaging/bioimage_inference.xml
```diff
@@ -0,0 +1,79 @@
...
+        python '$__tool_directory__/main.py'
+            --imaging_model '$input_imaging_model'
+            --image_file '$input_image_file'
+            --image_size '$input_image_input_size'
+    ]]>
+    </command>
+    <inputs>
+        <param name="input_imaging_model" type="data" format="zip" label="BioImage.IO model" help="Please upload a BioImage.IO model."/>
+        <param name="input_image_file" type="data" format="tiff,png" label="Input image" help="Please provide an input image for the analysis."/>
+        <param name="input_image_input_size" type="text" label="Shape of the input image" help="Provide shape of input image. See chosen model's RDF file to find correct input shape. For example: for the BioImage.IO model MitochondriaEMSegmentationBoundaryModel, the input shape is 256 * 256 * 32 * 1. Enter the shape as 256,256,32,1."/>
```
kostrykin, line 37: At first I was confused by the fact that you used `x` instead of regular `⨉` multiplication signs. Later I checked the RDF file, which you are referring to here, and saw that they also use the `x` notation. Sorry for the confusion, but taking this into account, I think that using `*` is the worst of all options, because *(i)* the meaning of `*` might be unknown to the users of the this tool and *(ii)* it has no direct relation to the RDF file. Since you seem to dislike `⨉` multiplication signs, I suggest reverting back to `x` as a compromise:
```suggestion
        <param name="input_image_input_size" type="text" label="Shape of the input image" help="Provide shape of input image. See chosen model's RDF file to find correct input shape. For example: for the BioImage.IO model MitochondriaEMSegmentationBoundaryModel, the input shape is 256 x 256 x 32 x 1. Enter the shape as 256,256,32,1."/>
```

tools/bioimaging/bioimage_inference.xml
```diff
@@ -0,0 +1,79 @@
...
+
+**Input files**
+    - BioImage.IO model: Add one of the model from Galaxy file uploader by choosing a "remote" file at "ML Models/bioimaging-models"
+    - Image to be analysed: Provide an image as TIF/PNG file
+    - Provide the necessary input shape for the model. This information can be found in the RDF file of each model (RDF file > config > test_information > inputs > size)
+
+**Output files**
+    - Predicted image: Predicted image using the BioImage.IO model
+    - Predicted image matrix: Predicted image matrix in original dimensions
+        ]]>
```
kostrykin, line 73: This indentation is still out of place. I have fixed BE→AE, and I have also added a sentence to briefly motivate what "prediction" is about (probably unknown by the target audience).
```suggestion
            **What it does**

            The tool takes a BioImage.IO model and an image (as TIF or PNG) to be analyzed. The analysis is performed by the model. The model is used to obtain a prediction of the result of the analysis, and the predicted image becomes available as a TIF file in the Galaxy history.

            **Input files**
            - BioImage.IO model: Add one of the model from Galaxy file uploader by choosing a "remote" file at "ML Models/bioimaging-models"
            - Image to be analyzed: Provide an image as TIF/PNG file
            - Provide the necessary input shape for the model. This information can be found in the RDF file of each model (RDF file > config > test_information > inputs > size)

            **Output files**
            - Predicted image: Predicted image using the BioImage.IO model
            - Predicted image matrix: Predicted image matrix in original dimensions
        ]]>
```

tools/bioimaging/bioimage_inference.xml
anuprulez: Yes, it would be good to follow the RDF.yaml file's notations.

tools/bioimaging/bioimage_inference.xml
```diff
@@ -0,0 +1,79 @@
...
+        python '$__tool_directory__/main.py'
+            --imaging_model '$input_imaging_model'
+            --image_file '$input_image_file'
+            --image_size '$input_image_input_size'
+    ]]>
+    </command>
+    <inputs>
+        <param name="input_imaging_model" type="data" format="zip" label="BioImage.IO model" help="Please upload a BioImage.IO model."/>
+        <param name="input_image_file" type="data" format="tiff,png" label="Input image" help="Please provide an input image for the analysis."/>
+        <param name="input_image_input_size" type="text" label="Shape of the input image" help="Provide shape of input image. See chosen model's RDF file to find correct input shape. For example: for the BioImage.IO model MitochondriaEMSegmentationBoundaryModel, the input shape is 256 x 256 x 32 x 1. Enter the shape as 256,256,32,1."/>
```
kostrykin, line 37: Just one last thing.

What do you think about adding the missing articles?

Besides, I think we're happy here.
```suggestion
        <param name="input_image_input_size" type="text" label="Shape of the input image" help="Provide the shape of the input image. See the chosen model's RDF file to find the correct input shape. For example: for the BioImage.IO model MitochondriaEMSegmentationBoundaryModel, the input shape is 256 x 256 x 32 x 1. Enter the shape as 256,256,32,1."/>
```

kostrykin (approved):
Thanks @anuprulez!

tools/bioimaging/bioimage_inference.xml
```diff
@@ -0,0 +1,79 @@
+<tool id="bioimage_inference" name="Process image using a BioImage.IO model" version="@TOOL_VERSION@+galaxy@VERSION_SUFFIX@">
+    <description>with PyTorch</description>
+    <macros>
+        <token name="@TOOL_VERSION@">2.3.1</token>
+        <token name="@VERSION_SUFFIX@">0</token>
+    </macros>
+    <creator>
+        <organization name="BioImage.IO" url="https://bioimage.io/#/" email="" />
+        <person name="Beatriz Serrano-Solano" email="" />
+        <person name="Leonid Kostrykin" email="leonid.kostrykin@bioquant.uni-heidelberg.de" />
```
bgruening, line 10: @anuprulez you forgot yourself here.

@beatrizserrano your email?

I would probably use `<person givenName="" familyName="" email="" />`

tools/bioimaging/bioimage_inference.xml
```diff
@@ -0,0 +1,79 @@
+<tool id="bioimage_inference" name="Process image using a BioImage.IO model" version="@TOOL_VERSION@+galaxy@VERSION_SUFFIX@">
+    <description>with PyTorch</description>
+    <macros>
+        <token name="@TOOL_VERSION@">2.3.1</token>
+        <token name="@VERSION_SUFFIX@">0</token>
+    </macros>
+    <creator>
+        <organization name="BioImage.IO" url="https://bioimage.io/#/" email="" />
```
bgruening, line 8:
```suggestion
        <organization name="European Galaxy Team" url="https://galaxyproject.org/eu/" />
```

tools/bioimaging/bioimage_inference.xml
```diff
@@ -0,0 +1,79 @@
+<tool id="bioimage_inference" name="Process image using a BioImage.IO model" version="@TOOL_VERSION@+galaxy@VERSION_SUFFIX@">
```
bgruening, line 1:
```suggestion
<tool id="bioimage_inference" name="Process image using a BioImage.IO model" version="@TOOL_VERSION@+galaxy@VERSION_SUFFIX@" profile="23.0">
```

tools/bioimaging/bioimage_inference.xml
```diff
@@ -0,0 +1,79 @@
...
+    <edam_operations>
+        <edam_operation>operation_3443</edam_operation>
+    </edam_operations>
+    <xrefs>
+        <xref type="bio.tools">pytorch</xref>
+        <xref type="biii">pytorch</xref>
+    </xrefs>
+    <requirements>
+        <requirement type="package" version="3.9.12">python</requirement>
+        <requirement type="package" version="2.3.1">pytorch</requirement>
```
bgruening, line 21:
```suggestion
        <requirement type="package" version="@TOOL_VERSION@">pytorch</requirement>
```

was that the intention behind the version number?

tools/bioimaging/.shed.yml
```diff
@@ -0,0 +1,11 @@
+name: bioimage_inference
```
bgruening, line 1: @anuprulez do you plan to add a second tool to this suite? If so this should be a suite based shed.yml file.

If we plan to add the tensorflow functions into this one tool, it can stay like it is.

tools/bioimaging/bioimage_inference.xml
kostrykin: My bad, I actually forgot to add this in my review.

tools/bioimaging/.shed.yml
anuprulez: I think the TensorFlow model rendering should also be included in this tool. But, adding that feature might need some time. It should be one tool.

tools/bioimaging/bioimage_inference.xml
anuprulez: I think tool's version should not replace Pytorch's version. Something else changes in the tool (making the tool's version to change) while Pytorch remains that same - could be problematic in future. Since, this is the first tool, I would keep the tool's version as 0.0.1

kostrykin: This wrapper is rather thin and is mostly "just" a wrapper for `torch.load` (the rest is pre- and post-processing). This suggests that the wrapper should have the same version as the wrapped tool, PyTorch in this case. If the wrapper changes in the future, you would usually increment `@VERSION_SUFFIX`. This is also what I had suggested in https://github.com/bgruening/galaxytools/pull/1391#discussion_r1699899153 and it is more in line with IUC recommendations https://galaxy-iuc-standards.readthedocs.io/en/latest/best_practices/tool_xml.html#tool-versions

anuprulez: sorry, I misinterpreted the previous comment. It's restored based on your comment. Thanks!!

tools/bioimaging/bioimage_inference.xml
```diff
@@ -0,0 +1,80 @@
...
+        python '$__tool_directory__/main.py'
+            --imaging_model '$input_imaging_model'
+            --image_file '$input_image_file'
+            --image_size '$input_image_input_size'
+    ]]>
+    </command>
+    <inputs>
+        <param name="input_imaging_model" type="data" format="zip" label="BioImage.IO model" help="Please upload a BioImage.IO model."/>
+        <param name="input_image_file" type="data" format="tiff,png" label="Input image" help="Please provide an input image for the analysis."/>
+        <param name="input_image_input_size" type="text" label="Shape of the input image" help="Provide the size of the input image. See the chosen model's RDF file to find the correct input size. For example: for the BioImage.IO model MitochondriaEMSegmentationBoundaryModel, the input size is 256 x 256 x 32 x 1. Enter the size as 256,256,32,1."/>
```
kostrykin, line 38: For consistency:
```suggestion
        <param name="input_image_input_size" type="text" label="Size of the input image" help="Provide the size of the input image. See the chosen model's RDF file to find the correct input size. For example: for the BioImage.IO model MitochondriaEMSegmentationBoundaryModel, the input size is 256 x 256 x 32 x 1. Enter the size as 256,256,32,1."/>
```

tools/bioimaging/bioimage_inference.xml
```diff
@@ -0,0 +1,80 @@
...
+    <help>
+        <![CDATA[
+            **What it does**
+
+            The tool takes a BioImage.IO model and an image (as TIF or PNG) to be analyzed. The analysis is performed by the model. The model is used to obtain a prediction of the result of the analysis, and the predicted image becomes available as a TIF file in the Galaxy history.
+
+            **Input files**
+            - BioImage.IO model: Add one of the model from Galaxy file uploader by choosing a "remote" file at "ML Models/bioimaging-models"
+            - Image to be analyzed: Provide an image as TIF/PNG file
+            - Provide the necessary input shape for the model. This information can be found in the RDF file of each model (RDF file > config > test_information > inputs > size)
```
kostrykin, line 69: For consistency:
```suggestion
            - Provide the necessary input size for the model. This information can be found in the RDF file of each model (RDF file > config > test_information > inputs > size)
```

kostrykin (commented):
Just a minor consistency issue

bgruening:
Thanks everyone!

anuprulez:
Thanks @bgruening for merging.

Thank you very much @kostrykin for spending so much time reviewing it

PR #1396, opened 2024-03-05
adjust .shed.yml to fix hictk failure after version update

fubar2:
fix planemo lint failure from missing .shed.yml items...


bgruening:
Thanks Ross!

PR #1397, opened 2024-03-07
Updating cmsearch deoverlap

RZ9082:
Infernal output is now formatted to be Tab-delimited, for better use with galaxy.


tools/rna_tools/cmsearch_deoverlap/cmsearch-deoverlap.xml
```diff
@@ -13,9 +13,11 @@
         #end if
         #if $dirty:
             $dirty &&
+            sed -i 's/ \{1,\}/\t/g' *.sort &&
```
bgruening, line 16:
```suggestion
            sed -i 's/[:blank:]+/\t/g' *.sort &&
```

Maybe?

PR #1399, opened 2024-03-12
Added bio.tools ref to glimmer-hmm

RZ9082:
This PR links glimmer-hmm to its bio.tools [entry](https://bio.tools/glimmer-hmm).


bgruening:
Wow, those are really old tools. @abretaud should we deprecate them?

@RZ9082 test are not starting because this one does not even have a .shed.yml

RZ9082:
> @RZ9082 test are not starting because this one does not even have a .shed.yml

They do not even contain any tests!!

bgruening:
grafik

No comment ;)

PR #1400, opened 2024-03-12
Added bio.tools ref to lighter

RZ9082:
This PR links lighter to its bio.tools [entry](https://bio.tools/lighter).

### DIFF
--- a/tools/lighter/lighter.xml
+++ b/tools/lighter/lighter.xml
@@ -32,7 +32,7 @@
     ]]>
     </command>
     <inputs>
-        <param argument="-r" multiple="True" type="data" label="Input file"/>
+        <param argument="-r" multiple="True" type="data" format="data" label="Input file"/>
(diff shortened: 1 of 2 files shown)


tools/lighter/lighter.xml
```diff
@@ -32,7 +32,7 @@
     ]]>
     </command>
     <inputs>
-        <param argument="-r" multiple="True" type="data" label="Input file"/>
+        <param argument="-r" multiple="True" type="data" format="data" label="Input file"/>
```
bgruening, line 35: The file can be fasta and fastq, and can be gzip'ed with extension *.gz. 
                     When the input file is *.gz, the corresponding output file will also be gzip'ed.

From https://github.com/mourisl/Lighter

RZ9082: Thanks:)

bgruening (approved):

Issue #1412, opened 2024-03-26
New version of grep fails if pattern isn't found


Delphine-L:
Hi !  Version 9.3+galaxy0 of Search in textfiles (grep) (https://usegalaxy.eu/root?tool_id=toolshed.g2.bx.psu.edu/repos/bgruening/text_processing/tp_grep_tool/9.3+galaxy0) fails with the error `Fatal error: Exit code 1 ()` when the query returns no results. When using version 1.1.1, the job is not failing and is simply returning an empty output. 

Example history : https://usegalaxy.eu/u/delphine-l/h/exemple-grep

PR #1413, opened 2024-03-27
Fixes for grep tool

bernt-matthias:
fixes https://github.com/bgruening/galaxytools/issues/1412

### DIFF
--- a/tools/text_processing/text_processing/macros.xml
+++ b/tools/text_processing/text_processing/macros.xml
@@ -6,7 +6,7 @@
         </requirements>
     </xml>
     <token name="@TOOL_VERSION@">9.3</token>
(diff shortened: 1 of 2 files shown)


tools/text_processing/text_processing/macros.xml
```diff
@@ -6,7 +6,7 @@
         </requirements>
     </xml>
     <token name="@TOOL_VERSION@">9.3</token>
-    <token name="@VERSION_SUFFIX@">0</token>
+    <token name="@VERSION_SUFFIX@">1</token>
```
bgruening, line 9: this will bump all tools not just the grep tool

bgruening:
Ok, lets ship it!

PR #1420, opened 2024-04-30
Fix "Plot confusion matrix, precision, recall and ROC and AUC curves of tabular data" tool 

anuprulez:
- Add support for `h5mlm` model format
ping @bgruening thanks!


bgruening:
Let me know if the lints are fixable with a recent amount of time, otherwise we skip them for now.

bgruening:
The HTML files do not need to be included. You can check with asserts maybe?

https://docs.galaxyproject.org/en/latest/dev/schema.html#tool-tests-test-output-assert-contents

For example for the file size with a delta, or some text that you expect to be on the website.

bgruening:
The other linter warning is `.. WARNING: No citations found, consider adding citations to your tool.`

anuprulez:
> The other linter warning is `.. WARNING: No citations found, consider adding citations to your tool.`

yes, the XML linting got fixed locally as well after adding citations.

anuprulez:
> The HTML files do not need to be included. You can check with asserts maybe?
> 
> https://docs.galaxyproject.org/en/latest/dev/schema.html#tool-tests-test-output-assert-contents
> 
> For example for the file size with a delta, or some text that you expect to be on the website.

this I will check.

anuprulez:
```
<output name="output_confusion" file="cnf_binary_sgd.html">
	        <assert_contents>
		    <has_size value="3486809" delta="10000" />
	        </assert_contents>
	    </output>
	    <output name="output_prf" file="prf_binary_sgd.html">
		<assert_contents>
		    <has_size value="3486974" delta="10000" />
	    </assert_contents>
        </output>
```

I tried to add the above changes locally to the first test, but the test did not pass. I will look at it a bit later. Python and tool linting tests pass now. thanks!

bgruening (approved):

tools/plotly_ml_performance_plots/test-data/model_binary_rfc.h5mlm
bgruening, line 1: this file needs to be that large?

anuprulez: maybe not. I am reducing the number of random trees and checking ...

tools/plotly_ml_performance_plots/test-data/prediction_binary_knn.tabular
```diff
@@ -1,801 +1,49 @@
...
-1.64	1.12	1
--1.7	-0.5539999999999999	1
-0.40399999999999997	0.00999	1
-0.7040000000000001	-0.9059999999999999	1
-0.335	1.39	-1
--1.7	-0.569	1
-2.64	1.14	1
-0.769	0.772	-1
--0.255	-0.142	1
+0	58	56	-67	0
```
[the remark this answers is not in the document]
anuprulez: I have changed the (sample) training and test data for the KNN method, which is why it is smaller. The train and test data being used for other methods in this PR do not work with KNN. Therefore, I am using the training and test data used in the original KNN method for writing automatic tests: https://github.com/bgruening/galaxytools/blob/master/tools/sklearn/nn_classifier.xml

training: https://github.com/bgruening/galaxytools/blob/master/tools/sklearn/test-data/train_set.tabular
test: https://github.com/bgruening/galaxytools/blob/master/tools/sklearn/test-data/test_set.tabular

bgruening:
Thanks @anuprulez

PR #1422, opened 2024-05-14
Add MGnify seqprep

RZ9082:
Adding a modified version of SeqPrep. For use with the MGnify pipeline.


tools/mgnify_seqprep/.shed.yml
```diff
@@ -0,0 +1,12 @@
+name: mgnify_seqprep
+owner: iuc
```
bgruening, line 2:
```suggestion
owner: bgruening
```

PR #1423, opened 2024-05-16
Update Cellpose to 3.0.8

pavanvidem:
thanks @bgruening for fixing the container issue!


pavanvidem:
There are some test datasets of more than 1MB. Should we move them to Zenodo or is it fine to keep them?

bgruening (approved):

PR #1426, opened 2024-05-22
Adapt uniprot id mapping tool to new API

bernt-matthias:
- building on the python script given at https://www.uniprot.org/help/id_mapping#example_uniparc
- currently without limits (i.e. submitting to many IDs will lead to an error)
- added code to statically define from and to selects using the API. from and to have been simplified a bit.

Functionality should be the same. Main difference in the UI is the flattened selection of `from` and `to` .. if desired I can also try to keep it as before.


bgruening (approved):
I don't mind, looks good to me. The missing limit might become a problem at some point, but I guess we can fix this later.

Feel free to merge.

PR #1427, opened 2024-05-23
Fix bug in split_file_to_collection

### LINKED ISSUE #1424: `Split file` doesn't include header in last dataset when called on rows
The tool doesn't include the header in the last row, so the resulting dataset is only a single line - see the example dataset here: https://usegalaxy.eu/api/datasets/4838ba20a6d86765d3fbd03294f83c9e/display?to_ext=tabular

This causes the problems in the parameter sweeps tested in this history: https://usegalaxy.eu/u/hechth/h/parameter-sweep-testing-hechth-202405-1-1

https://github.com/bgruening/galaxytools/blob/7e5c9f013821c8905f2c9c452a1f00891e3a0cfd/tools/text_processing/split_file_to_collection/split_file_to_collection.xml#L131

zargham-ahmad:
This PR includes the following:

- added condition to write header in the final record dataset, it was not previously written in the last dataset if the chunk size was 1
- added test

closes #1424

### DIFF
--- a/tools/text_processing/split_file_to_collection/.shed.yml
+++ b/tools/text_processing/split_file_to_collection/.shed.yml
@@ -2,6 +2,10 @@ owner: bgruening
 name: split_file_to_collection
 categories: 
   - Text Manipulation
-description: Split tabular, MGF, FASTA, or FASTQ files to a dataset collection. 
+description: Split tabular, MGF, FASTA, or FASTQ files to a dataset collection.
+long_description: |
+    Split file into a dataset collection.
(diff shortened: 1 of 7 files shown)


hechth:
@bgruening I think this is ready to go :)

bgruening (approved):

bgruening:
Looks good, can you please also fix the linting issues.

hechth:
@bgruening could you rerun the CI please? :)

bgruening:
Sure, can you reply in Matrix please :)

PR #1432, opened 2024-06-05
add graphprot log output to find error in Galaxy version

michauhl:
Hi, 
GraphProt on Galaxy currently does not work. I added the --gp-output to also get GraphProt log outputs, otherwise there is no way telling what the issue is. Offline GraphProt conda version fresh install works fine with the Galaxy python wrapper scripts, just tried.
Best,
Michael


bgruening:
Do you get any error message?

michauhl:
On Galaxy the output just says "output file not found", so not informative. Locally everything works fine. --gp-output gives full output of GraphProt

PR #1445, opened 2024-07-02
Update jbrowse2 conda to 2.12.3 

fubar2:
Latest release still understands our json.

- Big clean up and simplification of default view track style settings.
  - As part of adding working bigwig controls for colour (Yay!) and other style settings
  - Can now do the same more easily for other tracks
- Add bpPerPx to scale default view. Not automatable apparently.
- Add a plugin to colour bed features crudely according to score above/below zero.
  -  for use with https://github.com/galaxyproject/tools-iuc/pull/6121
- Update some help text and prompts


fubar2:
Shall I remove it to get past this wart?

> .. ERROR: Invalid XML found in file: tmpofyb0kul. Errors [/tmp/tmpofyb0kul:1681:0:ERROR:SCHEMASV:SCHEMAV_ELEMENT_CONTENT: Element 'creator': This element is not expected.]

bgruening:
> @bgruening: Shall I remove `<creator>` to get past this planemo wart?
> 
> > .. ERROR: Invalid XML found in file: tmpofyb0kul. Errors [/tmp/tmpofyb0kul:1681:0:ERROR:SCHEMASV:SCHEMAV_ELEMENT_CONTENT: Element 'creator': This element is not expected.]

https://docs.galaxyproject.org/en/latest/dev/schema.html#tool-creator 
Creator is valid ... there needs to be something else.

tools/jbrowse2/jbrowse2.xml
```diff
@@ -8,6 +8,7 @@
         <xref type="bio.tools">jbrowse2</xref>
     </xrefs>
     <expand macro="requirements"/>
+    <expand macro="creators"/>
```
bgruening, line 11:
```suggestion
```

Its already included below citations.

tools/jbrowse2/jbrowse2.xml
```diff
@@ -1129,28 +1146,30 @@ JBrowse2-in-Galaxy
...
-Compared to JBrowse1-in-Galaxy, there is no support for alternative codons for unusual genomes,
-and detailed track styling is not yet implemented. Please contact us if you are missing features.
 
 Use and local viewing
 =====================
 
+Browser tracks need a coordinate system, based on the reference genome, so that must be chosen before adding groups of tracks.
 
-A JBrowse2 history item can be opened by viewing it (the "eye" icon).
+At present, 10 Galaxy datatypes that can be turned into a track for display - in all cases, the selected reference genome must have been
```
bgruening, line 1155: I would not mention "10". Its hard to update this.

tools/jbrowse2/macros.xml
```diff
@@ -108,6 +108,7 @@
             <param name="zipOut" type="boolean" truevalue="true" falsevalue="false" checked="false" label="Create a zip archive for downloading rather than viewing" help="Default is to make an interactive browser appear when the 'eye' icon is activated"/>
             <param name="defaultLocation" type="text" value="" label="Subset to display to new users" help="Initial subset to be shown for users who have never visited the browser before. Example: 'ctgA:1234..5678"/>
             <param name="session_name" type="text" value="New JBrowse2 session" label="Session name" help="Displayed at the top of the window"/>
+            <param name="bpPerPx" type="integer" value="50" label="Base pairs per display pixel for default view" help="Make the default window show more or less BP of sequence. For some reason not automatable in JB2, sorry"/>
```
bgruening, line 111: min=1 ?

tools/jbrowse2/macros.xml
```diff
@@ -239,6 +240,7 @@
                 <option value="LinearPileupDisplay">LinearPileupDisplay - good for dense tracks at scale</option>
                 <option value="LinearArcDisplay">LinearArcDisplay</option>
             </param>
+            <param name="usebedscore" type="boolean" truevalue="true" falsevalue="false" checked="true" label="Use column 5 as a score for the display if bed format"/>
```
bgruening, line 243: Would it be better to let the user select the column?

Maybe that is a stupid idea.
What happens when you disable that?

tools/jbrowse2/jbrowse2.xml
fubar2: Agree. Just happy to have them all :)

tools/jbrowse2/jbrowse2.xml
```diff
@@ -334,25 +347,28 @@ export JBROWSE2_PATH=\$(dirname \$(which jbrowse))/../opt/jbrowse2 &&
                     <param name="genome" type="data" format="fasta" optional="true" label="Select the reference genome"/>
                 </when>
                 <when value="uri">
-                    <param name="uri" type="text" label="URI pointing to tabix compressed fasta"/>
-                    <param name="refname" type="text" label="Reference dbkey for tracks with this genome, such as hg38 or SacCer4"/>
+                    <param name="uri" type="text" label="URI pointing to tabix compressed fasta - must always be on-line" 
+                        help="Warning: Requires an internet connection to view. If this URI is not available for any reason, the track will show an error. Saves disk storage">
+                    </param>
+                    <param name="refname" type="text" label="Reference dbkey for tracks with this genome, such as hg38 or SacCer4">
+                    </param>
```
bgruening, line 354:
```suggestion
                    <param name="refname" type="text" label="Reference dbkey for tracks with this genome, such as hg38 or SacCer4"/>
```

tools/jbrowse2/jbrowse2.py
```diff
@@ -1322,6 +1335,7 @@ def process_annotations(self, track):
             else:
                 logging.warning("Do not know how to handle %s", dataset_ext)
             # Return non-human label for use in other fields
+            logging.debug("### processanno ext=%s trackstoadd = %s" % (dataset_ext, self.tracksToAdd))
```
bgruening, line 1338: remove?

tools/jbrowse2/jbrowse2.py
```diff
@@ -1146,10 +1171,10 @@ def add_paf(self, data, trackData, pafOpts, **kwargs):
             if gname not in self.genome_names:
                 # ignore if already there - eg for duplicates among pafs.
                 asstrack, first_contig = self.make_assembly(gpath, gname, useuri)
-                self.genome_names.append(gname)
+                self.genome_names.append(copy.copy(gname))
```
bgruening, line 1174: why is the hard-copy needed?

tools/jbrowse2/macros.xml
fubar2: Let's not go there yet. It's a minefield and just not worth exploring IMHO.
Bed format is too sinful for that.
This has a very specific use case at present although if useful it could be expanded, but unfortunately JB2 plugins are not trivial or flexible unfortunately.
Disabled should be default - will fix.
Enabled adds a new home made plugin to set the colour of the feature if score < 0 or > 0.

tools/jbrowse2/macros.xml
fubar2: Good point. Thanks.

tools/jbrowse2/jbrowse2.py
fubar2: in that string case it's not - that's a wart I'll fix thanks - I got overenthusiastic making copies of mutables to  appended to generate json. Was seeing strange interactions I think.

tools/jbrowse2/jbrowse2.py
fubar2: Sure. Will do

tools/jbrowse2/macros.xml
```diff
@@ -239,6 +240,7 @@
                 <option value="LinearPileupDisplay">LinearPileupDisplay - good for dense tracks at scale</option>
                 <option value="LinearArcDisplay">LinearArcDisplay</option>
             </param>
+            <param name="usebedscore" type="boolean" truevalue="true" falsevalue="false" checked="false" label="Use column 5 as a score for the display if bed format"/>
```
bgruening, line 243: But the 5th value of a 6-column BED is the score. So why is this not the default? I'm not sure if I understand this parameter and its implications. If I take this off - set to false - which is now the default. Which column will it take instead?

fubar2: If off, nothing changes and it is just another jbrowse2 bed track so tests pass as usual.

There is no current way to use the bed score in JB2, other than turning that switch on.
That's why it was included for the specific use case.

bgruening: What about a help text in the like of "Use column 5 of your BED file as a score to drive different color schemas.""

bgruening:
Good from my side. :tada:

PR #1448, opened 2024-07-08
Fix cut tool

bernt-matthias:
the list of columns must be after `-f` (or `-c` i.e. `cut_element`). I would have liked to add a test, but we can not use an option with the value `","` in a test (it will split it to a list of options `["", ""]`).

Still broken is that `data_column` parameters do not work for anything besides tab and comma separated values. Fixing this properly would require a restructuring of the tool form....

### DIFF
--- a/tools/text_processing/text_processing/macros.xml
+++ b/tools/text_processing/text_processing/macros.xml
@@ -6,7 +6,7 @@
         </requirements>
     </xml>
     <token name="@TOOL_VERSION@">9.3</token>
-    <token name="@VERSION_SUFFIX@">1</token>
+    <token name="@VERSION_SUFFIX@">2</token>
     <token name="@PROFILE@">23.1</token>
     <xml name="stdio">
         <stdio>
--- a/tools/text_processing/text_processing/macros.xml
+++ b/tools/text_processing/text_processing/macros.xml
@@ -6,7 +6,7 @@
         </requirements>
     </xml>
     <token name="@TOOL_VERSION@">9.3</token>
-    <token name="@VERSION_SUFFIX@">2</token>
+    <token name="@VERSION_SUFFIX@">1</token>
     <token name="@PROFILE@">23.1</token>
     <xml name="stdio">
         <stdio>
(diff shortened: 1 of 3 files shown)


bgruening:
Its unfortunately not a suite, so every change will bump all tools. So better to increase all versions?

bernt-matthias:
> Its unfortunately not a suite, so every change will bump all tools. So better to increase all versions?

In this case we might consider fixing cut properly. Wondering how to structure the tool form in the best way. The point is that `data_column` will only work for tab and comma separated inputs (i.e. on the choice of `infile` (format) and `delimiter`).

bgruening:
For which inputs is this a problem? Is the real fix here to make the tool form more flexible and allow an overwrite in those cases?

bernt-matthias:
If we choose anything but tab or comma in the delimiter the tool is broken at the moment, because `data_column` only works for tabular input data (with columns metadata), which is only defined for csv, tsv and tabular.

bgruening:
https://github.com/bgruening/galaxytools/issues/1480

mvdbeek:
Let's merge this ? 

> data_column only works for tabular input data

is not a tool issue so I don't think this should stop the PR.

mvdbeek (approved):

tools/text_processing/text_processing/macros.xml
```diff
@@ -6,7 +6,7 @@
         </requirements>
     </xml>
     <token name="@TOOL_VERSION@">9.3</token>
-    <token name="@VERSION_SUFFIX@">1</token>
+    <token name="@VERSION_SUFFIX@">2</token>
```
mvdbeek, line 9: I would bump just the tool. It'll create a new installable revision but at least there won't be bumps for tools that haven't changed.

bgruening: Is that the case? Ok, then we can do it.

bgruening:
```suggestion
    <token name="@VERSION_SUFFIX@">1</token>
```

bgruening:
Feel free to merge if you think its good to go.

mvdbeek:
Thanks a lot @bernt-matthias @bgruening!

PR #1450, opened 2024-07-11
A data manager for motus

paulzierep:
The data manager worked locally, and the DM could be consumed by the tool.
However, since there is no test DB yet, the test run only checks if the url to the DB is correct and copies a dummy folder to the `GALAXY_DATA_MANAGER_DATA_PATH` location. This dummy DB does not work with the tool !


data_managers/data_manager_motus_db_downloader/data_manager/motus_db_fetcher.xml
```diff
@@ -0,0 +1,68 @@
...
+    <![CDATA[
+        python '$__tool_directory__/data_manager_fetch_motus_db.py' --out '${out_file}' --version '${version}' 
+        $test_data_manager
+    ]]>
+    </command>
+    <inputs>
+    <param name="test_data_manager" type="boolean" truevalue="--test" falsevalue="" checked="False" label="Copy minimal test DB and create mock data table entry." />
+
+    <!-- <param name="test_data_manager" type="text" value=""/> -->
+    <param argument="--version" type="select" multiple="false" label="Database Version" help="Check https://zenodo.org/records/7778108 for current version. The default value is recommended.">
```
bgruening, line 21: You couls use a text param here with options.

```
<param name="foo" type="text" value="foo 1">
    <option value="foo 1">Foo 1 Display</option>
    <option value="foo 2">Foo 2 Display</option>
</param>

File: data_managers/data_manager_motus_db_downloader/data_manager/motus_db_fetcher.xml [hunk 2]
```diff
@@ -0,0 +1,68 @@
...
+    <!-- <expand macro="version" /> -->
+
+    <command detect_errors="exit_code">
+    <![CDATA[
+        python '$__tool_directory__/data_manager_fetch_motus_db.py' --out '${out_file}' --version '${version}' 
+        $test_data_manager
+    ]]>
+    </command>
+    <inputs>
+    <param name="test_data_manager" type="boolean" truevalue="--test" falsevalue="" checked="False" label="Copy minimal test DB and create mock data table entry." />
```
bgruening, line 18: should this param be hidden?

data_managers/data_manager_motus_db_downloader/data_manager/motus_db_fetcher.xml
paulzierep: But this will allow users to add different versions (that cannot be downloaded since the links do not work). So I would prefer to limit the choice using the type="select"

paulzierep:
Can we merge this @bgruening

data_managers/data_manager_motus_db_downloader/data_manager/motus_db_fetcher.xml
bgruening: True, but DMs are not for users, they are for admins - we assume they now what they are doing.

The downside of not allowing free-text inputs is that for every new release we need to update the DM - which is a lot of work.

paulzierep: While I do agree, we also need to make sure that data tables are populated with the same conventions between servers, to make workflows that depend on those DBs federated usable. I also think for tools that can work with many DBs one needs to be more flexible, but for tools that usually have only one DB per version associated, a more strict enforcement of data table entries is needed.

Issue #1458, opened 2024-07-22
CPAT tool options


SaimMomin12:
**Request on behalf of our Galaxy user**

--

Dear Galaxy Europe team,

I recently found out that the usegalaxy.eu <http://usegalaxy.eu/> server
has the tool CPAT (link:
https://usegalaxy.eu/root?tool_id=toolshed.g2.bx.psu.edu/repos/bgruening/cpat/cpat/3.0.5+galaxy0
<https://usegalaxy.eu/root?tool_id=toolshed.g2.bx.psu.edu/repos/bgruening/cpat/cpat/3.0.5+galaxy0>
) available. This is a tool I had previously used online so this is very
helpful for me (the online version only allows <10 MB files, while mine
are larger).
However, it looks like there is a problem with the interface/options for
running this tool.

The first option (Query Nucleotides) allows FASTA files similar to the
online version, but everything else requires files from the history for
things like the reference genome. Is it possible to include an option to
use data already included in Galaxy for this? i.e. to allow using the
included hg19 genome, transcripts, CDS, etc. that I've seen other tools use?

Also, the default version of the CPAT tool published online (link:
https://wlcb.oit.uci.edu/cpat/ <https://wlcb.oit.uci.edu/cpat/> )
includes an option to just upload your transcript sequences as FASTA
files, without any additional files for known CDS. I don't know if their
site has the human reference genome already included, but it would be
nice if I can do something similar in Galaxy.

Thank you for any help you can provide.

Sincerely,
Pavel Vlasov, Columbia University Biological Sciences

SaimMomin12:
- [ ] Add an option to use reference either from history or in built reference

PR #1459, opened 2024-07-22
CPAT: Added option to use reference genome from builtin references

SaimMomin12:
Based on a request from a Galaxy user regarding feature addition in the CPAT tool as described in #1458 this PR adds the functionality of utilizing built in reference genomes from Galaxy as well as retains the functionality of using reference genome from history.


SaimMomin12:
That skipped out of my mind, sure I will add it. Thanks :)

tools/cpat/cpat.xml
```diff
@@ -144,6 +166,41 @@
             </output>
             <output name="no_orf_seqs" file="test02_no_orgs.txt" ftype="txt"/>
         </test>
+        <test expect_num_outputs="4">
+            <param name="gene" value="sequences.fasta.gz"/>
+            <conditional name="ref_source">
+                <param name="source" value="builtin" />
+                <param name="ref_fasta" value="test1.fasta.gz" />
```
bgruening, line 173: ref_fasta_builtin I assume

SaimMomin12: I have defined the value as `builtin` as below

```
<param type="select" label="Reference genome source" name="source">
    <option value="history" selected="true">Use from History</option>
    <option value="builtin">Use Built-in</option>
</param>
```

bgruening: But ref_fasta is not a Param in the built-in case isn't it? On my mobile, no maybe I don't see it

SaimMomin12: Yes I checked it, it should be ref_fasta_builtin I have made the change in recent commit

tools/cpat/test-data/all_fasta.loc
bgruening, line 1: its better no name this one `all_fasta.loc.test`

tools/cpat/tool_data_table_conf.xml.test
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0"?>
+<tables>
+    <!-- Locations of reference genome files in fasta format -->
+    <table name="all_fasta" comment_char="#">
+        <columns>value, name, path</columns>
+        <file path="${__HERE__}/test-data/all_fasta.loc" />
```
bgruening, line 6:
```suggestion
        <file path="${__HERE__}/test-data/all_fasta.loc.test" />
```

bgruening (approved):
This looks all great now. The only thing that worries me is - why did the test pass in the first place (before your last commit), they shouldn't or? I'm just worried the test is not stringent enough, but will trust you if you think everything is good.

SaimMomin12:
> This looks all great now. The only thing that worries me is - why did the test pass in the first place (before your last commit), they shouldn't or? I'm just worried the test is not stringent enough, but will trust you if you think everything is good.

I think that was merely due to the less stringency of the test params. I think everything is good to go. This can be merged.

Issue #1452, opened 2024-07-11
imagej2 enhance contrasts fails


bernt-matthias:
stderr is

```
Available RAM: 345459MB, using 3/4 of that: 259095MB
JRE not found in '/usr/local/share/jre'
JAVA_HOME contains a JRE: '/usr/local/share/java/linux-amd64/jdk1.8.0_66//jre'
Re-executing with correct library lookup path (/.singularity.d/libs:/usr/local/share/lib/linux64:/usr/local/share/mm/linux64)
Available RAM: 345459MB, using 3/4 of that: 259095MB
JRE not found in '/usr/local/share/jre'
JAVA_HOME contains a JRE: '/usr/local/share/java/linux-amd64/jdk1.8.0_66//jre'
Using JAVA_HOME /usr/local/share/java/linux-amd64/jdk1.8.0_66//jre
Opening Java library /usr/local/share/java/linux-amd64/jdk1.8.0_66//jre/lib/amd64/server/libjvm.so
Java HotSpot(TM) 64-Bit Server VM warning: ignoring option PermSize=128m; support was removed in 8.0
Picked up _JAVA_OPTIONS: -Xmx6g -Xms256m
Java HotSpot(TM) 64-Bit Server VM warning: Using incremental CMS is deprecated and will likely be removed in a future release
Adding option: -Djava.home=/usr/local/share/java/linux-amd64/jdk1.8.0_66/
ij1.plugin.dirs: null
$HOME/.plugins: /root/.plugins does not exist
java.lang.IllegalArgumentException: Cannot handle app name in ij.gui.YesNoCancelDialog's public <init>(java.awt.Frame parent, java.lang.String title, java.lang.String msg)
	at net.imagej.patcher.CodeHacker.replaceAppNameInCall(CodeHacker.java:446)
	at net.imagej.patcher.LegacyExtensions.insertAppNameHooks(LegacyExtensions.java:419)
	at net.imagej.patcher.LegacyExtensions.injectHooks(LegacyExtensions.java:291)
	at net.imagej.patcher.LegacyInjector.inject(LegacyInjector.java:308)
	at net.imagej.patcher.LegacyInjector.injectHooks(LegacyInjector.java:109)
	at net.imagej.patcher.LegacyEnvironment.initialize(LegacyEnvironment.java:101)
	at net.imagej.patcher.LegacyEnvironment.applyPatches(LegacyEnvironment.java:495)
[...]

PR #1460, opened 2024-07-23
Fix for imagej relative path handling

bernt-matthias:
fixes https://github.com/bgruening/galaxytools/issues/1452. Unfortunately I could not reproduce the problem in a planemo test, bot the change implemented in https://github.com/bgruening/galaxytools/pull/1460/commits/a2d0e956125e1fb924cf8e139560de122b0cf006 fixed the problem on my instance. 

TODO: 

- [x] error log handling
- [x] profile
- [x] tool version macros

I won't update the tools here, since this does not work out of the box


tools/image_processing/imagej2/imagej2_macros.xml
```diff
@@ -1,8 +1,9 @@
 <macros>
-    <token name="@WRAPPER_VERSION@">3.0</token>
+    <token name="@TOOL_VERSION@">20170530</token>
+    <token name="@VERSION_SUFFIX@">20170530</token>
```
bgruening, line 3: This looks wrong. But the old .1 was also not good, so no idea what is bigger.

bernt-matthias: I just like it if the tool version matches the version of the underlying software

```python
>>> from packaging.version import Version
>>> Version("20170530+galaxy0") > Version("3.0.1")
True
```

bgruening: Bu the suffix is the same as the version?

bernt-matthias, line 3:
```suggestion
    <token name="@VERSION_SUFFIX@">0</token>
```

bgruening:
Merci!

PR #1463, opened 2024-07-26
Update tool - add keep_longest_isoform to agat

tbrown91:
One more program available in agat that we are in need of in order to create our genome reports. Tests were looking a little funky as I'm not sure if there are different output formats given by agat on different devices. For me Test14 was failing - which I don't think I touched.


tools/agat/agat.xml
```diff
@@ -136,12 +136,19 @@
                 #end if
                 -pe $tool.prot_existence
                 -o galaxy_functional_annot
+        #else if $tool.selector == 'keep_longest_isoform'
+            @input_annotation_single@
+            agat_sp_keep_longest_isoform.pl
+                --gff '$input_annotation'
+                --config '$agat_configfile'
+                --output ./output &&
+            cat ./output > '${annotation}'
```
bgruening, line 145:
```suggestion
                --output '${annotation}'
```

We can do that in the other places as well. This extra call of cat is not needed.

tbrown91: ok, I can change these

bgruening:
Nearly perfect! Thanks Tom! One small hint, never work on the master/main branch in git. This will cause you few problems down the road.

bgruening (approved):

tbrown91:
yes Sir

PR #1467, opened 2024-07-31
UniProt: batched access to id mapping

bernt-matthias:
local manual test worked. lets see if CI is happy.

Uniprot limits to 100k entries to map.


bgruening:
Thanks!

PR #1472, opened 2024-08-06
Switch AGAT output to gff3 and update to 1.4.0

### LINKED ISSUE #1471: agat - should the output be GFF3?
Hi, just a question. AGAT seems to output GFF3, e.g. `agat_convert_sp_gxf2gxf.pl`   produces files with the header

```
 ##gff-version 3 
```

The wrapper plain GFF, e.g.

https://github.com/bgruening/galaxytools/blob/d18fae96f65d1b748fbd21b92b75ca3181fdd8bc/tools/agat/agat.xml#L338

Workaround on Galaxy is to `Edit Dataset Atrributes` and save it as gff3.

Could we just change the wrapper to output gff3 or gff1 is really needed by this tool?

Thanks!

TomHarrop:
Switch AGAT gff output to gff3 (fixes #1471).

@bgruening some output has changed between version 1.2.0 and 1.4.0, please see the diff.

Using `gff3` will cause the opposite problem for users who want `gff1` . It could be fixed by rewriting the wrapper to use the full format (e.g. `gff`, `gff3`) as the value in the  `output_format.selector.version` rather than an integer version.

https://github.com/bgruening/galaxytools/blob/902e994cb04db968ce797afa676f5fa6512ab6d3/tools/agat/macros.xml#L28-L33


bgruening:
I'm not sure if gff1 is really useful. But your proposal sounds good. I think we could merge. Your call.

PR #1473, opened 2024-08-07
Add sample numbers to confusion matrix plot

anuprulez:
Old confusion matrix which does not show how many samples for each cell (not that intuitive).

old_plot

Update the confusion matrix showing sample numbers for each cell which is more intuitive and similar to what is used in deep learning tutorials for showing confusion matrix.

test_5_multi_lr

Once this tools is updated, I will make necessary changes in the classification tutorial before recording its video for GTA.

ping @bgruening Thanks!!


bgruening:
Thanks!

PR #1477, opened 2024-08-14
Update ChatGPT tool

arash77:
This pull request makes the question box required and provides a simple error message when the OpenAI API key is not provided in the user preferences.


tools/chatgpt/chatgpt.xml
```diff
@@ -33,7 +33,9 @@ $__user__.extra_preferences.get('chatgpt|api_key', "")
     </configfiles>
     <inputs>
         <param name="context" type="data" multiple="true" optional="false" format="doc,docx,html,json,pdf,txt,jpg,jpeg,png,webp,gif" label="Context" max="500"/>
```
bgruening, line 35: can you add here please an additional help - and again mentioned that its an online tool and data will be uploaded to OpenAI ...

PR #1485, opened 2024-08-20
add collection_type to input collection in JBrowse2

fubar2:
@mvdbeek pointed out that this was needed to lock down the collection input from slurping too much up.


bgruening (approved):

mvdbeek:
Thanks @fubar2!

PR #1494, opened 2024-09-02
Updating tools/cellpose from version 3.0.10 to 3.0.11

gxydevbot:
Hello! This is an automated update of the following tool: **tools/cellpose**. I created this PR because I think the tool's main dependency is out of date, i.e. there is a newer version available through conda.

I have updated tools/cellpose from version 3.0.10 to 3.0.11.

**Project home page:** https://github.com/MouseLand/cellpose/releases

For any comments, queries or criticism about the bot, not related to the tool being updated in this PR, please create an issue [here](https://github.com/planemo-autoupdate/autoupdate/issues/new).

PR #1495, opened 2024-09-02
Updating the test data for tools/cellpose

SaimMomin12:
This pull request updates the test-data for cellpose tool for an existing PR https://github.com/bgruening/galaxytools/pull/1494 and fixes https://github.com/bgruening/galaxytools/pull/1494


SaimMomin12:
For the tests here, should we keep the `delta_frac` high?

kostrykin:
Yes, using `delta_frac` to verify the result image is not reliable.

I'd suggest to use the `img_diff` comparison method https://docs.galaxyproject.org/en/master/dev/schema.html#tool-tests-test-output, because there are composite images involved (i.e. figures with multiple images). The assertions are not suitable for that kind of images.

For the other images, using assertions should be possible, but since you have the expected data generated already, I think using `img_diff` is more reliable and easier.

SaimMomin12:
@kostrykin Thanks for the pointers. I will replace the tests with `image_diff` comparison and check if the tests pass.

SaimMomin12:
Using `image_diff` did not help in this case. We might have to think about some other alternative.

kostrykin:
As I see from the test results https://github.com/bgruening/galaxytools/actions/runs/10681249964?pr=1495, the issue is that the difference between the images generated by Cellpose on the one hand, and the expected images on the other hand, is simply too large. This is not an issue of testing, but of Cellpose generating unexpected results. The question is, why is that happening?

kostrykin:
Ok I managed to generate the results using `--update_test_data`… kind of a workaround, but anyways…

The issue is that the generated results indeed are very different than the expected results.

Here is the expected image for `img02_cp_masks_cyto.tif`:
image

And here is what Cellpose generated for `img02_cp_masks_cyto.tif`:
image

SaimMomin12:
@kostrykin this behaviour is quite strange. Are you using `--biocontainers` in your planemo test command?

kostrykin:
> @kostrykin this behaviour is quite strange. Are you using `--biocontainers` in your planemo test command?

It takes forever to run with `--biocontainers` like several hours, so I omitted it at first.

Interestingly, Test 1 fails when running with `--biocontainers` and `--update_test_data`:

> **Problem:**
> Expecting value: line 1 column 1 (char 0)
> **Command line:**
> export CELLPOSE_LOCAL_MODELS_PATH='cellpose_models' && mkdir -p segmentation && ln -s '/tmp/tmpgzgkjdjl/files/a/e/b/dataset_aebc8958-c288-417a-b3e7-b8b78370474d.dat' ./image.png &&  python '/home/void/Documents/SaimMomin12-galaxytools/tools/cellpose/cp_segmentation.py' --inputs '/tmp/tmpgzgkjdjl/job_working_directory/000/2/configs/tmprb58hfv0' --img_path ./image.png --img_format 'png' --output_dir ./segmentation

The output of Cellpose for `img02_cp_masks_cyto.tif` is also different:

image

So we have 3 different outputs, one for the expected data, one for without `--biocontainers`, and one another for with `--biocontainers`. I have absolutely no idea what is going on here. How did you generate the output for the expected data?

SaimMomin12:
@kostrykin Ideally the tools needs to be tested with `--biocontainers` as this simulates the exact behaviour of Planemo run on GitHub. 

I updated the test data with the following command
`planemo t cellpose.xml --galaxy_root /home/momins/galaxy/ --biocontainers --update_test_data`

kostrykin:
> @kostrykin Ideally the tools needs to be tested with `--biocontainers` as this simulates the exact behaviour of Planemo run on GitHub.
> 
> I updated the test data with the following command `planemo t cellpose.xml --galaxy_root /home/momins/galaxy/ --biocontainers --update_test_data`

I did `planemo test --update_test_data --biocontainers`.

Test 1 fail to run with the above mentioned error. I attached the other generated results. [test-data.tgz](https://github.com/user-attachments/files/16871988/test-data.tgz)

Do you obtain the same results when you re-generate them?

SaimMomin12:
The *.tif files are similar, but the generated PNGs are different from yours.

kostrykin:
I just tested the base branch on my computer.

Using `--biocontainers` still is horribly slow, but faster than with the new version in this PR (like just 1-2 hours instead of 2-3). Moreover, every test passes when using `--update_test_data`.

For the `img02_cp_masks_cyto.tif` file, I again get different results than the expected image in the `test-data` directory. **But, the file has the same file size!** Since the tests used file size comparison before, I suspect the the issue that we are facing here was already present in the base branch version.

In consequence, since the base branch version is faster, does not exhibit failing tests, and to better understand what is going on, I suggest the following strategy:
1. _First_ migrate the tests of the base branch to `img_diff`. In the course of this, we should investigate whether we get persistent results in CI and locally. Fix everything that needs to be fixed.
2. _Then_ re-do this PR against the fixed base branch.

SaimMomin12:
@kostrykin I will first migrate the changes of `image_diff` to the base branch and see if all tests pass and then compare the results of CI and locally. For testing this, we update the tool version or retain the same version?

kostrykin:
> @kostrykin I will first migrate the changes of `image_diff` to the base branch and see if all tests pass and then compare the results of CI and locally. For testing this, we update the tool version or retain the same version?

I think we keep the `TOOL_VERSION`. Possibly it would make sense to increment the `VERSION_SUFFIX`.

SaimMomin12:
@kostrykin I was testing the base branch of cellpose and migrated the `image_diff` changes to it in my last commit. The tests fail here as well due to different images being generated during tests in CI. 

You are correct that all the tests of the tool pass when  `--update_test_data` is used locally, but they do fail here in CI. I am not so sure know why is this happening. Did you find anything similar?

kostrykin:
Nope, unfortunately, I haven't seen similar issues before. Basically there are two options, I guess: Either a dependency of the [Cellpose package](https://github.com/SaimMomin12/galaxytools/blob/cdd5be10848b9050631fa1110adc29c5591911bf/tools/cellpose/cellpose.xml#L14) is resolved to different versions and this induces the diverging behavior. Or it is some internal switch of Cellpose, that causes the divergence. Are the results consistent when running locally with CPU vs. GPU utilization?

Maybe those who implemented/updated the Cellpose wrapper have an idea? @sunyi000 @pavanvidem

bgruening:
Here are the updated test-data with docker: https://github.com/bgruening/galaxytools/pull/1505

pavanvidem:
As we are now using conda requirements, differences maybe due to some dependency. Let me test this.

bgruening:
What should we do here? Do we assume cellpose knows what they are doing and accept that the outputs have changed so dramatically?

We all use biocontainers, hopefully, we should not use conda for testing, locally or remote only biocontainers should be used.

SaimMomin12:
>What should we do here? Do we assume cellpose knows what they are doing and accept that the outputs have changed so dramatically?

>We all use biocontainers, hopefully, we should not use conda for testing, locally or remote only biocontainers should be used.

ping @kostrykin

bgruening:
> @bgruening If I understood https://github.com/bgruening/galaxytools/pull/1495#issuecomment-2360571156 comment correctly, then the problem here is that the Cellpose wrapper for one and the same version of Cellpose produces very different results. This makes me concerned regarding reproducibility. If the wrapper now yields different results when being run locally vs. in CI, it might as well yield inconsistent results in the future when run in Galaxy, as long as we do not understand why the inconsistency happens.

I'm not sure we should assume that. The new tests are stricter and we don't know for sure how it was with the old version + old tests, correct?

To make it easier for all of us, we should always use `--biocontainers`. There is just one problem, also the `--biocontainer` is created on the fly and depending on which time it was created can have different dependencies. There are two workarounds:

1) add more requirements to the tool
2) create a container for this PR already in https://github.com/BioContainers/multi-package-containers/ then this PR will use this rebuild image and we are using all the same image (in contrast to using a on-the-fly build one)

sunyi000:
I could try rework on this tool..  I was trying to use this in one of my testing workflow, it didn't work.
Since the python code was developed with cellpose2, I was thinking of remove it and re-do the tool with cellpose3 headless options

tischi:
Hi, for deploying it in Galaxy, does it have to reproducible in terms of _rebuilding_ the container image, or would be it possible to _build the image once_, and then never touch it again (assuming that at least the results when running from the same image are reproducible)?

bgruening:
@tischi that would be possible. But how do you decide which container is "correct"?

tischi:
Same result as the CellPose in BAND and BARD?
We could also ask Ruman to run the same image in JIPipe.

kostrykin:
> assuming that at least the results when running from the same image are reproducible

As far as I remember from talking to @pavanvidem, the issue is that the results aren't even reproducible when using the same image between the runs (if opencv is installed via Conda, but it works if it is installed via pip). So the assumption is unfortunately wrong. What we need to do here is to build a container for Cellpose where opencv is installed via pip.

bgruening:
ok, since nothing is moving here, I decided to use containers only.

I also updated to the latest version available.

tools/cellpose/cellpose.xml
```diff
@@ -11,7 +11,7 @@
         </xml>
     </macros>
     <requirements>
-        <requirement type="package" version="@TOOL_VERSION@">cellpose</requirement>
+        <container type="docker">quoay.io/biocontainers/cellpose:@TOOL_VERSION@</container>
```
pavanvidem, line 14:
```suggestion
        <container type="docker">quay.io/biocontainers/cellpose:@TOOL_VERSION@</container>
```

pavanvidem:
alternatively, we can also update the cellpose conda recipe with the `opencv-python-headless` instead of `opencv`. I guess that it will fix the reproducibility issue but not 100% sure.

tischi:
Thanks @bgruening for driving this forward!

FWIW, I had a look at the recipe that we use to build the CellPose Docker for our virtual desktop. 

I think the relevant lines are those:

```dockerfile
RUN conda create -y --name cellpose python=3.10 -c conda-forge && \
    conda run --name cellpose python -m pip install torch==2.5.0 torchvision==0.20.0 torchaudio==2.5.0 --extra-index-url https://download.pytorch.org/whl/cu118 && \
    conda run --name cellpose python -m pip install pyqt6==6.6.1 pyqt6-qt6==6.6.1 && \
    conda run --name cellpose python -m pip install cellpose[gui] numpy==1.26.3
```

I _think_ this installs the latest CellPose, which should be `3.1.0`.

Maybe @sunyi000 can add more details and confirm? Note the Yi is currently on holidays and may not answer straight away.

bgruening:
@tischi https://github.com/conda-forge/cellpose-feedstock/pull/23

pavanvidem:
shall we skip 3.0.11 and bump to 3.1.0? It seems 3.0.11 is very similar to 3.0.10.

pavanvidem:
tests are passing but there is still there linting error because of some test data exceeding 1MB

bgruening:
Working on it: https://github.com/BioContainers/multi-package-containers/pull/3430

bgruening:
So they are now way bigger than before?

pavanvidem:
> So they are now way bigger than before?

no, +/- 1kB.

pavanvidem:
one of the params `tile` was removed from `models.Cellpose.eval`, probably it is always used in the latest version as it was recommended to use in the older versions.

PR #1499, opened 2024-09-12
upgrade Fiji, add ROI export to Analyze Particle, add filter tool and crop tool

lldelisle:


lldelisle:
Hi @bgruening ,
The CI is failing because the conda environment does not work in singularity because of, I think, the usage of `$PREFIX` in the build:
https://github.com/bioconda/bioconda-recipes/blob/e33340d37015161965b07fc082854a5c73020588/recipes/fiji/build.sh#L4-L7
Is there a way to tell galaxy to use conda instead of singularity for these tools?

bernt-matthias:
In order to use conda for CI you can add the path to the tool folder to `.tt_biocontainer_skip`. But I think it won't change much, since the container also just contains the conda environment :) 

What is the actual problem/symptom?

lldelisle:
Thanks @bernt-matthias this solved the issue.
The problem was that the conda recipe (which I did) uses a symlink of a bash script that it puts in `$PREFIX/bin`. And this `$PREFIX` does not work when you use it into docker, however, I don't know how to do differently.
The symptom was that all tests were failing with 'ImageJ command not found'.
However all tests were passing locally.

lldelisle:
This PR:
- use the last version of Fiji in conda
- fix adjust threshold binary to not perform a 'Make binary' at first as this was ignoring the method of autothreshold chosen after.
- Allow for manual thresholding method and autothresholding methods in adjust threshold.
- Remove the 'summary' option from analyze particle as this is not possible in headless
- Allow to export ROIs from analyze particle in a format compatible with [omero_roi_import](https://toolshed.g2.bx.psu.edu/repository/browse_repository?id=b91f1833da9239f2).
- Add new show for analyze particles
- Add new tool to apply filter (I have not implemented the 'Convolve' filter option)
- Removes the randomj noise which are not available in ImageJ without plugin installation.
- Change some values to be compatible with the new version of ImageJ.
- Update some outputs
- Provide a bash script to update the outputs of tests.

tools/image_processing/imagej2/imagej2_adjust_threshold_binary_jython_script.py
```diff
@@ -23,28 +21,23 @@
...
-    method_str = method
-IJ.setAutoThreshold(input_image_plus_copy, method_str)
+if method != "Manual":
+    # Set the options.
+    if black_background:
+        method_str = "%s dark" % method
+    else:
+        method_str = method
+    threshold_min = 1
+    threshold_max = float("inf")
```
lldelisle, line 31: I think this is the good values of thresholding to highlight what have been set as mask but I would prefer that someone checks it carefully.

lldelisle:
@rmassei, would you mind to review?

tools/image_processing/imagej2/imagej2_macros.xml
```diff
@@ -4,7 +4,7 @@
     <xml name="fiji_requirements">
         <requirements>
             <requirement type="package" version="@TOOL_VERSION@">fiji</requirement>
-            <requirement type="package" version="3.4">grep</requirement>
+            <requirement type="package" version="3.11">grep</requirement>
```
bgruening, line 7: :+1:

bgruening (commented):
Impressive work. Thanks @lldelisle.

Can you also remove `<xml name="stdio">` and include the `detect_errors` in the command section.

bgruening (approved):

lldelisle:
Let's wait for September 23rd before merging so @rmassei can review.

tools/image_processing/imagej2/imagej2_adjust_threshold_binary.xml
```diff
@@ -24,86 +24,119 @@ ImageJ --ij2 --headless --debug
...
+            <output name="output" file="blobs_threshold_0-8.gif" compare="sim_size"/>
+        </test>
+        <test>
+            <param name="input" value="confocal-series-first-channel.tif"/>
+            <conditional name="method_cond">
+                <param name="method" value="Default"/>
+                <param name="black_background" value="yes"/>
+            </conditional>
+            <param name="display" value="bw"/>
+            <output name="output" file="confocal-series-first-channel_threshold_default.tiff" compare="sim_size"/>
```
bgruening, line 138: It would be nice to use the new fancy imaging asserts from @kostrykin  here.

https://docs.galaxyproject.org/en/latest/dev/schema.html#has-image-center-of-mass

lldelisle:
The error for test is super strange (https://github.com/bgruening/galaxytools/actions/runs/10930578882/job/30343829303?pr=1499#step:9:3694):
```
FileNotFoundError: [Errno 2] No such file or directory: '/home/runner/.planemo/gx_venv_3.11_release_24.1/bin/galaxyctl'
```

bgruening:
I saw this yesterday as well in another repo. Maybe the CI is getting a new Python version not compatible with planemo yet?

bgruening:
Ok, this looks now like legit errors.

tools/image_processing/imagej2/imagej2_adjust_threshold_binary_jython_script.py
dianichj: May I suggest this, is it possible?

```python
# Set threshold to exclude pure black (background) but capture even low-intensity features
threshold_min = 1  # Starting point to exclude pure black

# Set threshold_max based on image bit-depth
# For 8-bit images, use 255; for 16-bit images, use 65535
if bit_depth == 8:
    threshold_max = 255  # Default for 8-bit images
elif bit_depth == 16:
    threshold_max = 65535  # Default for 16-bit images
else:
    threshold_max = float('inf')  # General fallback if bit depth is unknown

# The values can be adjusted by the user based on image characteristics

Comment (lldelisle) [reply tools/image_processing/imagej2/imagej2_adjust_threshold_binary_jython_script.py:37]: Sure, I implement it.

**Comment by lldelisle** [comment]
I don't know why the bunwarpj align test is not passing on the CI, it is passing locally:
```
All 3 test(s) executed passed.
imagej2_bunwarpj_align (Test #1): passed
imagej2_bunwarpj_align (Test #2): passed
imagej2_bunwarpj_align (Test #3): passed
```
I will try if I can reproduce with containers...

**Comment by lldelisle** [comment]
Pfiou... It took me long to understand why the tests were passing locally with conda and failing on the CI with docker.
The tool gives different results when you use the option `--cpus 1` (or anything else than 0) in `docker run` compared to when you run with conda or run with docker with `--cpus 0`. See here on left the result with conda or docker with `--cpus 0` and on the right with `--cpus 1`.
image

**Comment by bernt-matthias** [comment]
Are there docs on `--cpus` and `--cpus 0` in particular?

**Comment by lldelisle** [comment]
`--cpus 0` do not restrain the number of CPU. But no idea why this tool gives **very** different results when you contrain it...

**Comment by lldelisle** [comment]
This tool is not developped since 2019 but I can still try to write an issue... https://github.com/fiji/bUnwarpJ
How can I hardcode the number of CPUs used?

**Comment by bernt-matthias** [comment]
Maybe `-XX:ActiveProcessorCount="1"`? The called shell script seems to [come from bioconda](https://github.com/bioconda/bioconda-recipes/blob/56af0ab4015cce2a355cdc079fd0b8193bed142b/recipes/fiji/bunwarpj.sh#L40) .

**Comment by lldelisle** [comment]
Let's try

**Comment by lldelisle** [comment]
Indeed, it works. Let me push this.

File: tools/image_processing/imagej2/imagej2_macros.xml [hunk 4]
```diff
@@ -1,20 +1,12 @@
 <macros>
-    <token name="@TOOL_VERSION@">20170530</token>
+    <token name="@TOOL_VERSION@">20231211</token>
```
bgruening, line 2:
```suggestion
    <token name="@TOOL_VERSION@">20240614</token>
```

Latest version. Its not a must and will probably require more work. Just wanted to point out that we created a new package. We can do this as a followup if you like.

tools/image_processing/imagej2/imagej2_analyze_particles_binary.xml
```diff
@@ -97,32 +88,55 @@ $circularity_max
             <option value="no" selected="True">No</option>
             <option value="yes">Yes</option>
         </param>
+        <param name="export_rois" type="boolean" truevalue="true" falsevalue="false" checked="false" label="Export ROIs coordinates" help="For example to upload to OMERO" />
```
rmassei, line 91: I would change the label and help and do not use _ROIs_ but _particles outlines_.
A suggestion:

`<param name="export_rois" type="boolean" truevalue="true" falsevalue="false"
 checked="false" label="Export particles outlines coordinates"
 help="Exports particle coordinates as a tabular file, which can be imported as ROIs in OMERO." />`

tools/image_processing/imagej2/imagej2_analyze_particles_binary.xml
```diff
@@ -97,32 +88,55 @@ $circularity_max
...
     </inputs>
     <outputs>
         <data name="output" format_source="input">
             <filter>show != "Nothing"</filter>
        </data>
         <data format="tabular" name="results" label="${tool.name} (results) on ${on_string}">
-            <filter>show == "Nothing" or display_results_cond['display_results'] == "yes"</filter>
+            <filter>show == "Nothing" or display_results == "yes"</filter>
+        </data>
+        <data format="tabular" name="rois" label="${tool.name} (ROI coordinates) on ${on_string}">
```
rmassei, line 100: Same, maybe it is better to use "particles coordinates" instead of ROIs coordinates

rmassei (approved):

tools/image_processing/imagej2/imagej2_analyze_particles_binary.xml
lldelisle: Very good suggestion. I incorporated it.

lldelisle:
I think it is ready to be merged...

bgruening (approved):
Thanks everyone! Very cool!

If you want to move those tools into IUC please feel free to do so, they don't need to stick here.

lldelisle:
Hi everyone,
Something went wrong on the deployment:
- New tools have been published with the 'bgruening' owner.
- Old tools (owned by 'imgteam') has not been updated because 'You are not the owner of this repository, so you cannot administer it.'
 https://github.com/bgruening/galaxytools/actions/runs/11018288035/job/30599171388#step:6:227

lldelisle:
(Unrelated but each time I comment a PR I get a notification that the slash command failed probably because your 'secrets.PAT' variable is not set)

bgruening:
I will fix that.

lldelisle:
Would you mind to leave me a message here when it is done?

bgruening:
Done, I hope.

lldelisle:
Yes, perfect, thank you.

PR #1512, opened 2024-10-05
JBrowse2: add another required file for pulsar; remove label punctuation that can break JBrowse2 paths

fubar2:
- Fix missing plugin for pulsar deployments
- Fallback to using the jbrowse executable clone function if the dependency javascript cannot be found
- Labels become paths. Punctuation can break these so use string.translate to remove everything except `.`, `-`, and `/` from all labels


fubar2:
Test for translate used in sanitising labels for tracks added.
Tricky - cannot fake the entity_identifier string as a test param, so a new test file is named `m:e.r.l .i-n.bw` which is a legal filename and the test  confirms this becomes `"me.r.l.i-n.bw_0"` to confirm that ':' and ' ' have been removed at least...and no "none" added!

bgruening (approved):

PR #1513, opened 2024-10-07
JBrowse2: Speed up paf tracks with new tabix pif adapter.

fubar2:
@bgruening: Dotplots use unindexed PAF format files at present and having spent some time exploring these very interesting artifacts, they can be big, too slow to load and frustratingly slow to refresh after adjusting the view....

JB2 has a new "PairwiseIndexedPAFAdapter" adapter using tabix indexing so have made that the default for local PAF sourced tracks to make it faster for range requests on large PAF files. The remote URI handling is more complicated and assumes it is tabixed index only if it ends in pif.gz or paf.gz - not predictable unfortunately - otherwise it is an unindexed remote paf.

The first test has been adjusted to confirm that the PIF adapter is in play and that the tabix files are present.

Black has reformatted some code again so the diff is annoying - sorry.


bgruening:
Thanks! You should at some point start with a frash branch :-)

fubar2:
oh - yes - this is old. Thanks - will do...

PR #1516, opened 2024-10-08
JBrowse2: fix tabix pif indexing

fubar2:
Using the -n flag for sort was unwise - chromosome does not sort well that way.


fubar2:
close! wrong branch.

PR #1517, opened 2024-10-08
JBrowse2: fix tabix pif indexing

fubar2:
Using the -n flag for sort was unwise - chromosome does not sort well that way.


bgruening:
And what is the problem, the error? If biocontainers do not work, this is usually because of a missing dependency.

PR #1520, opened 2024-10-10
fix LC_ALL in sort tool

bgruening:


bernt-matthias (commented):
Guess this deserves a bump.

PR #1522, opened 2024-10-14
JBrowse2: Revert from pif indexes for paf

fubar2:
Painful.
The pif index is a GMOD thing - not tabix.
The utility works when tested with planemo/conda but fails with this node-js error using biocontainers so cannot be used. Tried making all the paths local - no difference. Something odd about node-js wanting to call **_uv_os_get_passwd_** in the biocontainer. Apparently it happens if a path cannot be found or is RO?

```
DEBUG:jbrowse:cd /tmp/tmpzefftlib/job_working_directory/000/17/outputs/dataset_1e09fe24-141b-439a-ba4d-e72140449478_files && cp /tmp/tmpzefftlib/files/c/c/5/dataset_cc53e7f1-20f8-44df-9069-dd99d8485e0b.dat ./temppif
DEBUG:jbrowse:cd /tmp/tmpzefftlib/job_working_directory/000/17/outputs/dataset_1e09fe24-141b-439a-ba4d-e72140449478_files && jbrowse make-pif ./temppif --out ./peach-grape-map.paf_10.pif.gz
node:os:359
    throw new ERR_SYSTEM_ERROR(ctx);
          ^

SystemError [ERR_SYSTEM_ERROR]: A system error occurred: uv_os_get_passwd returned ENOENT (no such file or directory)
    at userInfo (node:os:359:11)
    at Config._shell (/usr/local/lib/node_modules/@jbrowse/cli/node_modules/@oclif/core/lib/config/config.js:587:67)
    at Config.load (/usr/local/lib/node_modules/@jbrowse/cli/node_modules/@oclif/core/lib/config/config.js:309:27)
    at async Config.load (/usr/local/lib/node_modules/@jbrowse/cli/node_modules/@oclif/core/lib/config/config.js:167:9)
    at async Object.run (/usr/local/lib/node_modules/@jbrowse/cli/node_modules/@oclif/core/lib/main.js:62:20) {
  code: 'ERR_SYSTEM_ERROR',
  info: {
    errno: -2,
    code: 'ENOENT',
    message: 'no such file or directory',
    syscall: 'uv_os_get_passwd'
  },
  errno: [Getter/Setter],
  syscall: [Getter/Setter]
}

Node.js v22.9.0

```


bgruening:
I guess we should try to find out why the container does not work.

https://stackoverflow.com/questions/61961485/a-system-error-occurred-uv-os-get-passwd-returned-enoent-no-such-file-or-direc

Indicates some permission problems. Which is very likely in a container setting.

fubar2:
@bgruening: yes - likely a file permission error and well beyond my skills. 

Meanwhile, synteny tracks do not work right. At least this reverts to working paf and retains the code - just need to set canPif true. Would prefer to merge this while someone figures out the real cause and fixes it.

fubar2:
@bgruening: Until someone (not me!) figures out the fix, the current JBrowse2 is broken and this PR will fix it by reverting.
Please merge, then when you get the biocontainer/nodejs devils exorcised, I can easily enable PIF indexing.

bgruening:
Sorry, not a good week here ...

fubar2:
Sorry you don't have enough monkeys to make every week a good one - sorry to add to your woes - thanks!

scottcain:
One suggestion from @garrettjstevens:

One possible guess, CLI commands might fail if the user running them (meaning OS user) has no homedir or username, which might be true in some containers.

Reasoning: the CLI framework we use probably calls a nodejs api that throws this error if the user running it has no username or homedir (see https://nodejs.org/docs/latest-v20.x/api/os.html#osuserinfooptions)

scottcain:
One suggestion from @garrettjstevens for why pif making might fail in a container:

One possible guess, CLI commands might fail if the user running them (meaning OS user) has no homedir or username, which might be true in some containers.

Reasoning: the CLI framework we use probably calls a nodejs api that throws this error if the user running it has no username or homedir (see https://nodejs.org/docs/latest-v20.x/api/os.html#osuserinfooptions)

fubar2:
Thanks @scottcain - that would explain the failures with other JBrowse command line tools I've seen too. If that's part of a common nodejs idiom, cannot expect every upstream application maintainer like JBrowse2 to remove it so this is likely to annoy many other biocontainer tool herders. 

Creative solutions welcomed!

scottcain:
@fubar2 and @bgruening is there a way to create a toy image that can reproduce this issue? I don't know enough about the galaxy build process to understand what is happening yet.

fubar2:
Posted https://github.com/GMOD/jbrowse-components/issues/4608 as a heads up for the JBrowse2 team.

scottcain:
Hmm, so it is docker. So the problem is not fundamental to docker, since we can build a docker container that does work, so it is how the biocontainers specifically are built. I'm guessing it's due to the stripped down nature of the os in the container. I wonder if writing a PAF indexer in a different but supported language (presumably python and perl would work) would be a faster solution.

fubar2:
@scottcain: suggest we consolidate the ongoing discussion in one place in the interests of minimal redundancy and collective sanity and probably here at https://github.com/GMOD/jbrowse-components/issues/4608 makes sense?

fubar2:
@scottcain: belay that.
Testing on the CL inside the running biocontainer with some test data mounted, jbrowse make-pif seems to work when running as root or adding `--user nobody` to the run command, so this is probably not primarily a JBrowse2/node problem - but something about the way Galaxy and this specific tool call the jbrowse executable. There's a subprocess utility which uses `shell=True` that I am eyeing with suspicion - it's legacy code so will experiment with getting rid of that.

bgruening:
Hi everyone. As Ross pointed out the containers are here https://quay.io/repository/biocontainers/jbrowse2?tab=tags&tag=latest

They are all automatically generated, without Dockerfile. Simply by "putting" conda packages into a minimal busybox containers. Yes, there is no user, no home, no nothing inside this container - its really very minimal.

fubar2:
@scottcain: good news - adding `SHELL="/bin/sh"` to the env for subprocess.run when calling the jbrowse CL seems to do the needful.

Will test and push tomorrow.

PR #1496, opened 2024-09-04
 Integrate black forest labs FLUX.1 models into Galaxy

arash77:
Adding [black forest labs FLUX.1](https://github.com/black-forest-labs/flux) into Galaxy tools

PR #1525, opened 2024-10-16
PTM Dephosphorylation Predictation Tool

haibkhn:
This pull request introduces a new tool for predicting dephosphorylation. Key features include:

- Two modes of operation:
  1. Manual hyperparameter selection
  2. Automated hyperparameter search using Optuna

- Support for 3 protein language model variations:
  - ProtT5-XL-UniRef50
  - ESM
  - ProtT5-XL-BFD

- Additional hyperparameter search option:
  - SMAC functionality included, but note that the latest SMAC version (2.2.0) is not yet updated in Anaconda

@anuprulez Please review these changes and let me know if any modifications are needed.

### DIFF
--- a/tools/prot_tools/optuna_tool.py
+++ b/tools/prot_tools/optuna_tool.py
@@ -39,6 +39,7 @@ from transformers import (
     EsmModel,
     EsmTokenizer,
     PreTrainedModel,
+    set_seed,
     T5EncoderModel,
     T5Tokenizer,
     Trainer,
@@ -46,7 +47,6 @@ from transformers import (
     TrainerControl,
     TrainerState,
     TrainingArguments,
-    set_seed,
 )
 from transformers.modeling_outputs import SequenceClassifierOutput
 from transformers.models.t5.modeling_t5 import T5Config, T5PreTrainedModel, T5Stack


anuprulez:
@haibkhn thanks for the PR. I will have a look and provide my feedback

anuprulez:
Here are a few obvious comments to start with:

- The tool does not have automatic tests. Please see here: https://github.com/bgruening/galaxytools/blob/master/tools/bioimaging/bioimage_inference.xml how to add tool tests
- CI's listing test is failing. Please fix those. All the automatic CI tests should pass.

ping @haibkhn 
Maybe @bgruening also has more comments. I will look into more details probably end of this week or next week.

tools/prot_tools/optuna_tool.xml
```diff
@@ -0,0 +1,339 @@
+<tool id="hyperparameter_finetune" name="Hyperparameter Search for Finetuning model" version="1.0.0">
```
bgruening, line 1:
```suggestion
<tool id="hyperparameter_finetune" name="Hyperparameter Search for Finetuning model" version="1.0.0" profile="23.0">
```

bgruening (commented):
please add a .shed.yml file

anuprulez:
I think while testing, the script tries to download the LLM (ProtT5-XL-UniRef50) having a size > 10 GB. It could be the reason CI/CD throws an error related to the memory

error_protrans

It could be possible to provide a remote link to the model but I think HuggingFace does not allow the model from remote. The models should be hosted at HuggingFace, correct? @haibkhn 

Moving to a container-based tool might help but not sure. Can the HuggingFace table used with the Flux tool might help?
I am looking at https://github.com/bgruening/galaxytools/pull/1496 where we store the names of HF models but not the models themselves

ping @bgruening @arash77 

Thanks!

arash77:
I don't think that it is possible to test the tool completely with the models in GitHub if the model is big. unless there is a smaller version of the model available.
But for production, we can use two methods, one that I have used is to use [HF_HOME](https://huggingface.co/docs/huggingface_hub/en/package_reference/environment_variables#hfhome) environment variable to store all hugging face models or you can use [cache_dir in from_pretrained function](https://huggingface.co/docs/transformers/en/main_classes/model#transformers.PreTrainedModel.from_pretrained.cache_dir)

PR #1529, opened 2024-10-19
jbrowse2: fix for pif; deprecate blastxml 

fubar2:
@bgruening:
Finally fixed the [pif problem](https://github.com/GMOD/jbrowse-components/issues/4608#issuecomment-2421929295)
blastxml is now non-functional and has been deprecated because the old JB1 converter emits gff that 2.15.4 refuses to display with an improved parser. All removed including tests...


fubar2:
@bgruening 
Nothing is really lost. It should never have been there really because the python converter code is still available as the original tool: toolshed.g2.bx.psu.edu/repos/iuc/blastxml_to_gapped_gff3/blastxml_to_gapped_gff3/1.1 so anyone wanting a GFF3 track from a blast xml output can use that and then load the GFF into JBrowse2. 

It was passing tests because everything works _except display_ - the inputs convert, but when loaded as a GFF3 JBrowse2 track, show validation errors with some (not all!) blastxml samples I tried. There have been a number of changes to the GFF parsing noted in recent release notes and it looks like it is now more strict.

image

Point being, the blastxml input option is redundant and at slightly broken so let's leave it out and use the existing converter which can be fixed separately if anyone needs it.

fubar2:
For the record,
https://usegalaxy.eu/u/fubar/h/imported-from-archive-jb2testout contains a GFF3 made using the blastxml converter tool. In this [2.15.4 version](https://usegalaxy.eu/datasets/4838ba20a6d8676540d05128fb06efec/preview) the track shows the error above. In [this 2.14.0 version](https://usegalaxy.eu/datasets/4838ba20a6d86765eeed1c72c66a4ae4/preview), the same track works normally. Something seems to have changed in the JBrowse2 GFF parser.

bgruening:
Ok this makes sense. Thanks for writing this down!

PR #1533, opened 2024-10-22
Add a Galaxy tool for TabPFN package (by Prof. Hutter's group)

anuprulez:
The PR adds a Galaxy tool that uses the TabPFN package (DOI: 10.48550/arXiv.2207.01848) to create a general-purpose classifier. 

- It takes input two files - train and test data along with their corresponding labels 
- TabPFNClassifier trains on the training data and predicts classes using test data
- Additionally, it outputs a precision-recall curve. 
- A conda package has been created using the TabPFN PyPi package (https://github.com/conda-forge/tabpfn-feedstock) because it is required by the Galaxy platform

Many more subsequent analyses can be added as required such as creating workflows with this tool. 

A few challenges:

- One challenge is the export this classifier to standard ONNX format. The standard export of Pytorch models to ONNX does not work with the trained TabPFN model. We may require it, otherwise, we may have to export the trained model as a zipped file.

- Additionally, to move the discussion around integrating the `tabpfn_client` package - it requires a separate login before use, so integrating it to Galaxy would be complicated I think unless there is a way to bypass the login process.

Thank you!

ping @bgruening 

tabpfn1

tabpfn2

tabpfn3

### DIFF
--- a/tools/tabpfn/tabpfn.xml
+++ b/tools/tabpfn/tabpfn.xml
@@ -1,7 +1,7 @@
 <tool id="tabpfn" name="Tabular data prediction using TabPFN" version="@TOOL_VERSION@+galaxy@VERSION_SUFFIX@" profile="23.0">
     <description>with PyTorch</description>
     <macros>
-        <token name="@TOOL_VERSION@">0.1</token>
+        <token name="@TOOL_VERSION@">2.0.3</token>
         <token name="@VERSION_SUFFIX@">0</token>
     </macros>
     <creator>
@@ -10,7 +10,7 @@
         <person givenName="Frank" familyName="Hutter" email="fh@cs.uni-freiburg.de" />
     </creator>
     <requirements>
-	<requirement type="package" version="2.0.3">tabpfn</requirement>
+	<requirement type="package" version="@TOOL_VERSION@">tabpfn</requirement>
 	<requirement type="package" version="2.2.2">pandas</requirement>
 	<requirement type="package" version="3.9.2">matplotlib</requirement>
     </requirements>


tools/tabpfn/main.py
```diff
@@ -0,0 +1,54 @@
...
+import time
+
+import matplotlib.pyplot as plt
+import pandas as pd
+from sklearn.metrics import accuracy_score, average_precision_score, precision_recall_curve
+from tabpfn import TabPFNClassifier
+
+
+def separate_features_labels(data):
+    df = pd.read_csv(data, sep=",")
```
bgruening, line 14: can we read tabular? or both ... but tabular would be nice.

tools/tabpfn/tabpfn.xml
```diff
@@ -0,0 +1,62 @@
...
+            **Input files**
+            - Training data
+            - Test data
+
+            **Output files**
+            - Predicted data along with predicted labels
+        ]]>
+    </help>
+    <citations>
+        <citation type="doi">10.48550/arXiv.2207.01848</citation>
```
bgruening, line 60: we need to update this one

tools/tabpfn/tabpfn.xml
```diff
@@ -0,0 +1,62 @@
...
+    <version_command>echo "@VERSION@"</version_command>
+    <command detect_errors="aggressive">
+    <![CDATA[
+        python '$__tool_directory__/main.py'
+            --train_data '$train_data'
+            --test_data '$test_data'
+    ]]>
+    </command>
+    <inputs>
+	<param name="train_data" type="data" format="csv" label="Train data" help="Please provide training data for training model."/>
```
bgruening, line 26: Some more hints how the data needs to be structured would be good.

bgruening:
The results seem to be a bit random, so we need to work with asserts.

tools/tabpfn/tabpfn.xml
```diff
@@ -0,0 +1,62 @@
+<tool id="tabpfn" name="Tabular data prediction using TabPFN" version="@TOOL_VERSION@+galaxy@VERSION_SUFFIX@" profile="23.0">
+    <description>with PyTorch</description>
+    <macros>
+        <token name="@TOOL_VERSION@">0.1</token>
```
bgruening, line 4:
```suggestion
        <token name="@TOOL_VERSION@">2.0.3</token>
```

bgruening (approved):

PR #1537, opened 2024-10-30
Add --local option to Bismark tool

jonasbucherETH:
This pull request adds the `--local` option to the Bismark alignment tool in Galaxy.

**Changes:**
- Added a boolean parameter `local` to enable the `--local` option.
- Updated `bismark_bowtie2_wrapper.xml` to include the new parameter and updated the command section.
- Modified `bismark_wrapper.py` to handle the `--local` option when constructing the command line.


bgruening (commented):
Thanks a lot @jonasbucherETH!

Github actions do face some problems atm https://www.githubstatus.com/

Do you think we should also bump the requirement versions? e.g. the bowtie2 version and samtools?

jonasbucherETH:
I updated the requirement versions in the bismark_bowtie2_wrapper.xml. The linting seems to fail for other parts of bismark (for example bismark_methylation_extractor), should I update these as well to pass the required statuses and checks?

bgruening:
@jonasbucherETH that would be great! Our linting gets stricter over time, but we are improving the quality with that. Thanks for you help!

Please also bump then the version number of the other tools.

jonasbucherETH:
It seems I accidentally closed the request, is there a way to reopen it? Otherwise I will create a new one.

bgruening:
Can it be that you deleted your branch?

jonasbucherETH:
No, I think it is because I did a force-push, not realizing that this would mess with the pull request.

bgruening:
Upps, sorry! Can I help in any way?

PR #1539, opened 2024-11-01
Add --local option to Bismark Bowtie2 Wrapper (continuation from PR #1537)

jonasbucherETH:
This PR continues from PR #1537:

- Add --local option to Bismark Bowtie2 Wrapper
- update requirement versions
- handle lint warnings


bgruening:
This seems to be in the root folder, but it should be in the tools/bismark folder.

PR #1541, opened 2024-11-04
[BH2024] Update Bioimaging Archive retrieval tool 

B0r1sD:
Related to Project 17: Development of FAIR image analysis workflows & training in Galaxy
More context: https://github.com/beatrizserrano/BH2024-project17/issues/11

### DIFF
--- a/tools/image_processing/bia-ftplinks/biaftplink.xml
+++ b/tools/image_processing/bia-ftplinks/biaftplink.xml
@@ -1,4 +1,4 @@
-<tool id="bia_download" name="FTP Link for Bioimage Archive" version="@VERSION@+galaxy0" profile="22.05">
+<tool id="bia_download" name="FTP Link for Bioimage Archive" version="@VERSION@+galaxy1" profile="23.0">
     <description>Download images from Bioimage Archive</description>
     <macros>
         <import>macros.xml</import>


kostrykin:
By the way if you write

> Fix https://github.com/beatrizserrano/BH2024-project17/issues/11

into the text of your PR, it will be linked automatically by GitHub :)

B0r1sD:
Unfortunately, I discovered even the Bioimage -, Core accession numbers do not follow one clear pattern for their FTP link. See `ftp://ftp.ebi.ac.uk/biostudies/fire/S-BIAD/S-BIAD0-99/S-BIAD24` that is not following the S-BIAD/000/S-BIAD000 pattern that was assumed previously. IMO this tool should just ask in one line for the whole ftp link that users can find in the 'ftp' button in the study they want in their Galaxy history. This way it will be correct all the time and usable for the whole BIA. 
I can quickly prune this wrapper to do exactly that (as the previous version would just fail on these 'exceptions') and then focus on getting a Galaxy button on BIA.

B0r1sD:
The tool also keeps failing on the datatype .lif which is in a lot of BIA studies. The tool fetches whole studies atm so not only the images (hence the name is a bit misleading?). 

> Leia Image Format (LIF) images are JPEG images with metadata 

If we want the Galaxy button in BIA, I think it is a datatype that should be added to Galaxy. But I've heard this is quite cumbersome for the developers and I also have no idea if there are more datatypes in a similar siutation. 

We could introduce a filter to just download .tif/.tiff from studies but that probably drops metadata from the studies. I am leaning towards not spending more time on this refactoring as it exposes too many flaws - the proper solution (the Galaxy button in BIA) has been initatiated but there is no estimation on when it will be there so this tool has to do for now. 
Maybe just updating the help text with the explanation of this is enough for now?... Let me know what you think.

kostrykin:
Thanks so much for all your efforts, @B0r1sD.

Your points are perfectly reasonable. Asking the user for the whole FTP link sounds good. On the other hand, if the user already has the FTP link, couldn't the user then just use Galaxy's "Upload data" button and fetch the data via FTP from there? So would there be any point in using this tool? As far as I remember from talking to @sunyi000, the whole point of the tool was to get those FTP links, but maybe I'm missing something.

Personally, I think, we should generally consider one step after another. It will be a great improvement when the Galaxy button in BIA will be there. With the experiences and feedback we will hopefully gather then, we might consider adding support for further datatypes, LIF or beyond, but first we need the experiences in how far it's actually required.

> Maybe just updating the help text with the explanation of this is enough for now?

Agreed. I'd be interested in the stance of @sunyi000 who originally wrote this tool.

B0r1sD:
Exactly @kostrykin, but I think the original author didn't anticipate the heterogenity of the accession paths as most of them don't follow the expected pattern that would be scriptable. So the best course of action to keep this tool is to clearly state what it can and can not do.
Unfortunately, I've just received some feedback from Teresa at BIA about that button:

> Thanks so much for sending us such a detailed explanation of the implementation. Unfortunately, this would be non-trivial for us because the implementation would need to be done by another EBI service (BioStudies) whose backend we use. We are working on a BIA image-centric API but it’s not quite ready yet. For the moment, it would be easier if those using Galaxy either use [the BioStudies API](https://eur01.safelinks.protection.outlook.com/?url=https%3A%2F%2Fwww.ebi.ac.uk%2Fbiostudies%2Fhelp&data=05%7C02%7Cboris.depoortere%40vib.be%7C833b1669cebf4afa581508dd03cce77f%7C2d714a65b97f41a98ff1ec2cdf7df5cd%7C0%7C0%7C638670900837856294%7CUnknown%7CTWFpbGZsb3d8eyJFbXB0eU1hcGkiOnRydWUsIlYiOiIwLjAuMDAwMCIsIlAiOiJXaW4zMiIsIkFOIjoiTWFpbCIsIldUIjoyfQ%3D%3D%7C0%7C%7C%7C&sdata=81slSOMQq6dos%2B8PgzLXnDrDLAr3Ut9HA4etLr%2BjcTk%3D&reserved=0), FTP, or direct HTTP links.

I believe that was the (alpha) API we discussed was not yet ready to handle data requests, but I could be wrong? As for the metadata, it does seems like it would follow a clear logical of `https://www.ebi.ac.uk/biostudies/api/v1/studies/` + `accession_number` e.g. https://www.ebi.ac.uk/biostudies/api/v1/studies/S-BIAD24/ or https://www.ebi.ac.uk/biostudies/api/v1/studies/S-JCBD-201709074. 

So I guess we wait on their BIA API they are working on, and in the meanwhile address the tool as written above. Input from original author is of course always welcome!

sunyi000:
Hi @B0r1sD @kostrykin 

As I remember the motivation for this was just to get started with BIA in Galaxy, and the plan was to rework it once the BIA's API becomes available.

I am not using it or planning to work on it currently, so please feel free to delete,edit, continue or close it..

Thanks

B0r1sD:
@kostrykin Yes I checked it out and @pavanvidem made it use FTP only by default, which should work and was also one of the recommended ways from Teresa of BIA. I removed all commits but kept the addition of the EDAM topic and operation, that way this PR still serves as a little synergy between both efforts 😄. Thanks all, wishing you a good follow up Biohackathon!

tools/image_processing/bia-ftplinks/biaftplink.xml
```diff
@@ -1,8 +1,14 @@
-<tool id="bia_download" name="FTP Link for Bioimage Archive" version="@VERSION@+galaxy1" profile="22.05">
+<tool id="bia_download" name="FTP Link for Bioimage Archive" version="@VERSION@+galaxy2" profile="22.05">
```
kostrykin: I suppose, to be easy on the toolshed, @bgruening would prefer us to not bump the tool version just for the EDAM annotations, but keep the changes until the next time any actual functionality is added/updated, so that the changes will be deployed to toolshed only then?

B0r1sD: Ok that makes sense. You can close the pull request then. Anything we can do besides keeping an eye on the next round of updates to make sure this gets added?

bgruening: We can merge and not deploy ... if you like.

kostrykin: Lets just merge it with this:
```suggestion
<tool id="bia_download" name="FTP Link for Bioimage Archive" version="@VERSION@+galaxy1" profile="22.05">
```
Then, it won't be deployed, but the changes will be in for the next iteration, so they won't be forgotten.

B0r1sD: As in just merge it on the current version? If that is possible, that seems justified as this doesn't change any functionality of the tool, just adds some helpful metadata if you agree.

B0r1sD: > Lets just merge it with this:
> 
> Then, it won't be deployed, but the changes will be in for the next iteration, so they won't be forgotten.

Ah ok great, yes sounds good!

PR #1549, opened 2024-11-20
Drop datatype repository dependencies for infernal tools

mvdbeek:
AFAICT they are all in Galaxy now and we've deprecated loading TS datatypes a few years ago.


bgruening (approved):
Yes, we integrated them many years ago. Thanks!

PR #1551, opened 2024-11-27
Improve FLUX tool for output label and prompt display

arash77:
Enhance the output label based on the input type and include the prompt in the console output when using FLUX tool.

### DIFF
--- a/tools/flux/flux.xml
+++ b/tools/flux/flux.xml
@@ -2,7 +2,7 @@
     <description>text-to-image model</description>
     <macros>
         <token name="@TOOL_VERSION@">2024</token>
-        <token name="@VERSION_SUFFIX@">3</token>
+        <token name="@VERSION_SUFFIX@">4</token>
     </macros>
     <requirements>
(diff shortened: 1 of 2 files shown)


tools/flux/flux.xml
```diff
@@ -39,7 +39,7 @@ python3 '$__tool_directory__/flux.py'
         </conditional>
     </inputs>
     <outputs>
-        <data name="output" format="png" label="${tool.name} on ${on_string}" from_work_dir="./output.png"/>
+        <data name="output" format="png" label="${tool.name} on ${on_string if $input_type_selector=='file' else 'text prompt'}" from_work_dir="./output.png"/>
```
bgruening, line 42: if that a good idea? We don't know how large the prompt will be

arash77: It doesn't add the prompt it self, it just announces that it is a text prompt.
Before it was empty, if it was not a file.

arash77: Or should it be a short text about the prompt?

bgruening: No, that is ok. Thanks

PR #1552, opened 2024-11-30
Add wordcloud

reytakop:
The wordcloud.xml, macros.xml, and.shed.yml files were added.


bgruening:
The output file seems to be missing.

tools/wordcloud/macros.xml
```diff
@@ -0,0 +1,30 @@
...
+        <requirement type="package" version="1.2.2">sphinx_rtd_theme</requirement>
+        <requirement type="package" version="0.8.2">sphinx_gallery</requirement>
+        <requirement type="package" version="0.4.0">sphinx-argparse</requirement>
+        <requirement type="package" version="3.0.1">sphinx-issues</requirement>
+        <requirement type="package" version="3.8.0">matplotlib</requirement>
+        <requirement type="package" version="6.0.4">multidict</requirement>
+        <requirement type="package" version="0.42.1">jieba</requirement>
+        <requirement type="package" version="1.11.3">scipy</requirement>
+        <requirement type="package" version="2.0.2">python-bidi</requirement>
+        <requirement type="package" version="3.0.0">arabic_reshaper</requirement>
```
bgruening, line 28: Why is this all needed? I would assume that only wordcloud as dependency works?

tools/wordcloud/wordcloud.xml
```diff
@@ -0,0 +1,210 @@
...
+            <option value="RGB">RGB</option>
+            <option value="RGBA">RGBA</option>
+        </param>
+        <param argument="--max_words" value="200" type="integer" label="Maximum number of words"/>
+        <param argument="--width" value="400" type="integer" label="Define output image width"/>
+        <param argument="--height" type="integer" value="200" label="Define output image height"/>
+        <param argument="--random_state" type="integer" label="Random seed"/>
+        <param argument="--no_normalize_plurals" value="True" type="boolean" optional="True" label="Whether to remove trailing s from words"/>
+        <param argument="--repeat" type="boolean" truevalue="True" falsevalue="False" value="False" label="Whether to repeat words and phrases until max_words or min_font_size is reached"/>
+        <param argument="--version" type="boolean" truevalue="True" falsevalue="False" value="False" label="Show program's version number"/>
```
bgruening, line 154: version should not be offered to the user, but you should add a `<version_command>` tag to the tool.

tools/wordcloud/macros.xml
reytakop: > Why is this all needed? I would assume that only wordcloud as dependency works?

Yes, it works with only Wordclous as a requirement.

tools/wordcloud/wordcloud.xml
reytakop: > version should not be offered to the user, but you should add a `<version_command>` tag to the tool.

Thanks, I deleted this line.

bgruening:
I see this error here: 

> wordcloud_cli: error: argument --color: expected one argument

reytakop:
The wordcloud.xml file was updated.

bgruening:
@reytakop It seems that there is a problem in your command section.

> NameMapper.NotFound: cannot find 'color'

tools/wordcloud/wordcloud.xml
```diff
@@ -0,0 +1,182 @@
...
+        <param argument="--fontfile" type="data" format="ttf" label="Font file you wish to use " help=" The font file you want to use"/>
+        <param argument="--min_font_size" value="4" type="integer" label=" Smallest font size to use" optional="False"/>
+        <param argument="--max_font_size" type="integer" label="Maximum font size for the largest word" optional="False"/>
+        <param argument="--font_step" value="1" type="integer" label="Step size for the font" help="Font_step &gt; 1 might speed up computation but give a worse fit"/>
+        <param argument="--margin" value="2" type="integer" label="Spacing to leave around words"/>
+        <param argument="--background" type="color" label="Use given color as background color for the image" value="#000000">
+            <validator type="regex" message="Please select a valid RGB color">[#][0-9A-Fa-f]{6}</validator>
+        </param>
+        <param argument="--mask" type="data" format="png" optional="True" label="Mask to use for the image form"/>
+        <param argument="--colormask" type="data" format="png" optional="True" label="Color mask to use for image coloring"/>
```
bgruening, line 70: here are many "color" options, should they be in your color-conditional?

reytakop:
The wordcloud.xml file was updated, and the related error was fixed.

tools/wordcloud/macros.xml
```diff
@@ -0,0 +1,8 @@
+<macros>
+    <token name="@TOOL_VERSION@">1.9.4</token>
```
bgruening, line 2: If its just one tool, you can move this into the tool if you like - not as a separate file.

tools/wordcloud/wordcloud.xml
```diff
@@ -0,0 +1,217 @@
+<tool id="wordcloud" name="wordcloud" version="@TOOL_VERSION@+galaxy@VERSION_SUFFIX@" profile="@PROFILE@" license="MIT">
+    <description>A little word cloud generator in Python.</description>
```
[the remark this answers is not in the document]
wm75: or maybe even better:
`name="Generate a word cloud"` and `description="with highly customizable appearance."`

or sth similar.

tools/wordcloud/macros.xml
reytakop: Thanks for your comment. The files were updated based on your suggestion.

> If its just one tool, you can move this into the tool if you like - not as a separate file.

Thanks for your comment. The files were updated based on your suggestion.

reytakop: > If its just one tool, you can move this into the tool if you like - not as a separate file.

Thanks for your comment. The files were updated based on your suggestion.

wm75: Ah, you misunderstood: using the tokens instead of hard-coded versions is important for automated update suggestions, but you can set the macros right in the tool xml instead of expanding them from another file.

reytakop: > Ah, you misunderstood: using the tokens instead of hard-coded versions is important for automated update suggestions, but you can set the macros right in the tool xml instead of expanding them from another file.

Thanks for your comment. I updated the file.

PR #1555, opened 2024-12-04
Diff add -y parameter

RZ9082:
We should still consider making the HTML report to be dependent of `-y` and `-unified` in my opinion.


bgruening:
> We should still consider making the HTML report to be dependent of `-y` and `-unified` in my opinion.

Not sure what this means, but please do :)
or in the next version?

bgruening:
Do you know why diff raises a non 0 error code? You can also do `diff ---fooo | true` I think. But its a hack, better to figure put why diff is behaving like that.

RZ9082:
> Do you know why diff raises a non 0 error code? You can also do `diff ---fooo | true` I think. But its a hack, better to figure put why diff is behaving like that.

when files do not differ the diff tool returns exit code 0, when they do differ it exits with code 1. exit code 2 would mean an error.

wm75:
would be good to add a comment like that above the line of code

bgruening:
uha interesting, why did this not fail us before? Thanks @RZ9082 for figuring this out!

bgruening:
Ah, that was the reason the profile did not work. Cool!

Please add this as a comment and we can merge :)

RZ9082:
> Ah, that was the reason the profile did not work. Cool!
> 
> Please add this as a comment and we can merge :)

Exactly.
Comment added. Thanks!

bgruening (approved):

PR #1556, opened 2024-12-04
fix fly datatype

bernt-matthias:
And linter errors that will come with the new TestsCaseValidation linter

xref https://github.com/galaxyproject/tools-iuc/issues/6561


bgruening:
Thanks.

PR #1565, opened 2025-01-14
Add an admin-definable limit to the number of outputs rbc_splitfasta will create

natefoo:
I went with a default of unlimited but could also set some reasonable default if anyone has suggestions.

Additionally the tool will now use the sequence count (if available, although I believe it is [always available](https://github.com/galaxyproject/galaxy/blob/9fc0f0593517612f14c612c1648c2dc1a040f545/lib/galaxy/datatypes/sequence.py#L374)) from metadata rather than calculating it.

### DIFF
--- a/tools/splitfasta/split_fasta.py
+++ b/tools/splitfasta/split_fasta.py
@@ -42,7 +42,6 @@ with open(input_filename) as input_file:
     chunk_record_count = 0  # how many lines have we written to the output file
     records = []
     for record in SeqIO.parse(input_file, "fasta"):
-        print(f"processing record {record}")
         records.append(record)
         if num_chunks == 0 or (
             count < num_chunks and len(records) >= records_per_chunk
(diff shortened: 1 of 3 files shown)


tools/splitfasta/split_fasta.py
```diff
@@ -1,32 +1,41 @@
...
     with open(input_filename) as input_file:
         for line in input_file:
             if line.lstrip().startswith(">"):
                 record_count += 1
 
+if num_chunks != 0:
     records_per_chunk = round(float(record_count) / num_chunks)
 
+if record_limit and record_count > record_limit:
+    exit(f"ERROR: Number of sequences {record_count} exceeds limit {record_limit}")
```
wm75, line 37: @natefoo @bgruening isn't this wrong in the case of chunks?

natefoo: The chunk count is checked above on line 22 I believe?

natefoo: I think you're right, this should be an `elif`.

PR #1566, opened 2025-01-15
Fix PRC curve and add test

anuprulez:


anuprulez:
ping @bgruening thanks!

bgruening (approved):

PR #1248, opened 2022-10-14
Transfer tools from ASaiM repository 

gallardoalba:
Only the tools which are currently installed have been moved. Related to https://github.com/ASaiM/galaxytools/pull/47

PR #1568, opened 2025-01-17
tranfer group_humann2 tool

bernt-matthias:
I would need to get a container for this tool. We could add the ASaiM repo to planmeo-monitor, but the repo seem dead. Also this seems in line with: https://github.com/bgruening/galaxytools/pull/1248

This would then allow to https://github.com/ASaiM/galaxytools/pull/49


bgruening:
@bebatut ok for you?

bernt-matthias:
Seems that the environment can't be created anymore.

bgruening:
Let's bump the versions? Or should I try to create a container with those deps?

bernt-matthias:
> Let's bump the versions? Or should I try to create a container with those deps?

I would be happy if there would be a container where tests are passing.. with or without bumping.

bernt-matthias:
Seems that the conda environment can not be created anymore: https://github.com/bioconda/bioconda-recipes/pull/53474#issuecomment-2613977302

- a link in post-link.sh is broken, i.e. we would need to update the recipe in order to make it installable
- but we can not update since strict repo priority makes it uninstallable (goatools is only available for py3 from conda-forge, but humann2 requires py2)

The only solution that comes to my mind is to create an older version for goatools supporting python2.7.

A bit unfortunate, since the tool is used in GTN: https://github.com/galaxyproject/training-material/blob/b42014f17cb888716aae51705820aa8d9dce5633/topics/microbiome/tutorials/metatranscriptomics/content.md?plain=1#L956

PR #1573, opened 2025-02-04
Add option --no-sequence to the tool gfastats wrapper

Delphine-L:
This option allows to generate GFA files with no sequences.


tools/gfastats/macros.xml
```diff
@@ -44,4 +44,7 @@
             </when>
         </conditional>
     </xml>
+    <xml name="no_sequence_macro">
+        <param argument="--no-sequence" type="boolean" truevalue="--no-sequence" falsevalue="" checked="false" label="Do not include sequences in the GFA output (--no-sequence)" help="Select Yes for a lighter GFA file."/>
```
bgruening, line 48:
```suggestion
        <param argument="--no-sequence" type="boolean" truevalue="--no-sequence" falsevalue="" checked="false" label="Do not include sequences in the GFA output" help="Select Yes for a lighter GFA file."/>
```

No sequence is added by Galaxy to the help text if you use argument

bgruening (approved):

PR #1575, opened 2025-02-07
Adding log option required for multiqc for Sambamba markdup

SaimMomin12:
This PR adds log option required for MultiQC plugin


tools/sambamba/Sambamba_markdup.xml
```diff
@@ -12,6 +12,7 @@
                '${output}'
             --nthreads \${GALAXY_SLOTS:-4}
             --tmpdir '\${TMPDIR:-.}'
+            &> '${log}'
```
mvdbeek, line 15: You'll always want to use tee, otherwise it's quite difficult to debug failures. This is not so much a problem when redirecting the stdout and the tool consistently write error messages to stderr ... but scientific tools can't be trusted.

SaimMomin12: Thanks Marius for the suggestion, I have added the changes in the recent commit

bgruening:
Should the log file be optional?

SaimMomin12:
> Should the log file be optional?

Sure, we can also make it optional.

bgruening (approved):

bgruening:
We needed a build bump here.

SaimMomin12:
@bgruening Sorry, there was some issue in the last CI check. Auto-merge required again! Thanks.

bgruening:
Thanks!

PR #1544, opened 2024-11-07
[BHEU2024] Fix bioimage_inference tool with 3D images

kostrykin:
The tool yields an error when we try to use a 3D image. For example, using the [3DUNetArabidopsisApicalStemCells.pt model](https://bioimage.io/#/?tags=3d&id=emotional-cricket) and an image with shape `256,256,2,1` yields the error:

> Expected weight to be a vector of size equal to the number of channels in input, but got weight of shape [1] and input of shape [2, 256, 256]

Original PR for this tool: https://github.com/bgruening/galaxytools/pull/1391
xref https://github.com/beatrizserrano/BH2024-project17/issues/26

PR #1580, opened 2025-02-24
Fix Bioimaging inference tool

anuprulez:
Fixes https://github.com/bgruening/galaxytools/pull/1544 issue


tools/bioimaging/bioimage_inference.xml
```diff
@@ -30,12 +30,14 @@
...
             --image_file '$input_image_file'
             --image_size '$input_image_input_size'
+            --image_axes '$input_image_input_axes'
     ]]>
     </command>
     <inputs>
         <param name="input_imaging_model" type="data" format="zip" label="BioImage.IO model" help="Please upload a BioImage.IO model."/>
         <param name="input_image_file" type="data" format="tiff,png" label="Input image" help="Please provide an input image for the analysis."/>
         <param name="input_image_input_size" type="text" label="Size of the input image" help="Provide the size of the input image. See the chosen model's RDF file to find the correct input size. For example: for the BioImage.IO model MitochondriaEMSegmentationBoundaryModel, the input size is 256 x 256 x 32 x 1. Enter the size as 256,256,32,1."/>
+        <param name="input_image_input_axes" type="text" label="Axes of the input image" help="Provide the input axes of the input image. See the chosen model's RDF file to find the correct axes. For example: for the BioImage.IO model MitochondriaEMSegmentationBoundaryModel, the input axes is 'bzyx'"/>
```
bgruening, line 40: could this be a select?

bgruening: btw. a text param can also have select options ...

beatrizserrano: can't it be extracted from the rdf so that the user doesn't need to know this?

anuprulez: The parameter `axes` provides the real shape (number of dimensions) of the input image/matrix required by the model to make inferences.  The parameter `input_size` is necessary to slice pixels from different input image dimensions. So, both parameters I think are important to ascertain the target size of the image to be provided to the model. RDF file is not available inside the tool from where we could extract the `axes` information. Also, In a few RDF files, the axes information is not straightforward and written like (from 3d-unet-arabidopsis-apical-stem-cells model): 

```
inputs:
  - axes:
      - type: batch
      - channel_names:
          - channel0
        id: channel
        type: channel
      - id: z
        scale: 1
        size: 100
        type: space
      - id: 'y'
        scale: 1
        size: 128
        type: space
      - id: x
        scale: 1
        size: 128
        type: space
```

Users have to infer the correct axes information from above. From above, it is `bczyx`. Other RDF files have this information directly given as `bczyx`. (bczyx == batch, channel, z, y, x)

I can provide this as a select option.

tools/bioimaging/bioimage_inference.xml
```diff
@@ -46,16 +48,50 @@
...
+            <param name="input_image_input_axes" value="bczyx"/>
+            <output name="output_predicted_image" file="output_3d-unet-arabidopsis-apical-stem-cells.tif" compare="sim_size" delta="100" />
+            <output name="output_predicted_image_matrix" file="output_3d-unet-arabidopsis-apical-stem-cells.npy" compare="sim_size" delta="100" />
+        </test>
+        <test>
+            <param name="input_imaging_model" value="input_imaging_model.zip" location="https://uk1s3.embassy.ebi.ac.uk/public-datasets/bioimage.io/organized-badger/1/files/weights-torchscript.pt"/>
+            <param name="input_image_file" value="input_platynereisemnucleisegmentationboundarymodel.tif" location="https://uk1s3.embassy.ebi.ac.uk/public-datasets/bioimage.io/organized-badger/1/files/sample_input_0.tif"/>
+            <param name="input_image_input_size" value="256,256,32,1"/>
+            <param name="input_image_input_axes" value="bczyx"/>
+            <output name="output_predicted_image" file="output_platynereisemnucleisegmentationboundarymodel.tif" compare="sim_size" delta="100" />
```
bgruening, line 84: @anuprulez you can use the new image_diff here https://docs.galaxyproject.org/en/latest/dev/schema.html#id117

bgruening:
If we store the RDF for every model, next to the model on our server and provide this as a second input ... do you think we can then support most models?

We could add a conditional:
* try to extract model metadata from model RDF file
* set metadata on your own

bgruening:
We might be able to read the RDF file with https://github.com/bioimage-io/spec-bioimage-io

`conda install -c conda-forge bioimageio.spec`

Don't get me wrong, I think we can merge this PR as soon as it is green. Maybe we can add in a next iteration the automatic extraction?

tools/bioimaging/bioimage_inference.xml
beatrizserrano: @FynnBe I understood that the info in the RDF file is standardised, right?

tools/bioimaging/bioimage_inference.xml
```diff
@@ -46,16 +52,50 @@
...
+            <output name="output_predicted_image" file="output_nucleisegboundarymodel.tif" compare="image_diff" />
             <output name="output_predicted_image_matrix" file="output_nucleisegboundarymodel_matrix.npy" compare="sim_size" delta="100" />
         </test>
         <test>
             <param name="input_imaging_model" value="input_imaging_model.zip" location="https://zenodo.org/api/records/6647674/files/weights-torchscript.pt/content"/>
             <param name="input_image_file" value="input_nucleisegboundarymodel.png"/>
             <param name="input_image_input_size" value="256,256,1,1"/>
-            <output name="output_predicted_image" file="output_nucleisegboundarymodel.tif" compare="sim_size" delta="100" />
+            <param name="input_image_input_axes" value="bcyx"/>
+            <output name="output_predicted_image" file="output_nucleisegboundarymodel.tif" compare="image_diff" />
```
bgruening, line 64: If those images are too large we can use asserts:

https://github.com/BMCV/galaxy-image-analysis/blob/07aba15c8f753d650fee9cdd455896a77086a615/tools/superdsm/superdsm.xml#L171

anuprulez: Ok, will try that to check if it works with `TIF` extensions. There are also NPY files which have similar sizes as TIFF files

bgruening: just check there size, but using assert-size

anuprulez:
@bgruening can you rerun the tests? I am not sure why the first two tool tests fail. They show the following error:

```
Job in error state.. tool_id: bioimage_inference, exit_code: 1, stderr: Traceback (most recent call last):
  File "/home/runner/work/galaxytools/galaxytools/tools/bioimaging/main.py", line 7, in <module>
    import imageio
ModuleNotFoundError: No module named 'imageio'
```

If this is the error, the other tests from number 3 to 6 should also fail. Locally, all of the pass. 

Thank you!

bgruening:
This looks good to me!

Should we merge this one, so people can try it.

PR #1584, opened 2025-03-05
Add linter exceptions for 24.2

nilchia:
The skip lint is added based on the code provided in [#6795](https://github.com/galaxyproject/tools-iuc/pull/6795)

```bash
planemo ci_find_repos . > repos.txt
while read line
do
    echo $line;
    touch $line/.lint_skip;
    planemo l $line --skip_file $line/.lint_skip > t;
    egrep "WARN|ERROR" t | cut -d'(' -f2 | cut -d')' -f1 | sort -u >> $line/.lint_skip;
done < repos.txt
```


bgruening:
@nilchia this was good? Are you still working on it?

bgruening:
also ping @bernt-matthias

bernt-matthias:
Looks quite good. I guess a `find . -name .lint_skip -size 0 -delete`  should delete the empty files. Or in git `find . -name .lint_skip -size 0 -exec git rm {} \;`.

nilchia:
@bernt-matthias, does this PR conflict with this? https://github.com/bgruening/galaxytools/pull/1561

I closed it as I thought they might conflict.

bernt-matthias:
> @bernt-matthias, does this PR conflict with this? #1561
> 
> I closed it as I thought they might conflict.

Yours should be more up-to-date. We can close https://github.com/bgruening/galaxytools/pull/1561.

Just remove the empty files and it should be good to go.

nilchia:
I've removed the empty files.
It seems there are other linting problems as well.

bernt-matthias:
> It seems there are other linting problems as well.

All of them should be problems with URLs

The R linter problems could be fixed automatically with https://github.com/bgruening/galaxytools/blob/master/.github/styler.R .. but I would do this in a separate PR

bernt-matthias:
No, this is not the problem. I guess your planemo is not up do date. You need to make sure that galaxy-tool-util 24.2 is used.

t
```diff
@@ -0,0 +1,15 @@
+Linting tool /home/nilchia/Documents/github/bgruening/galaxytools/tools/wtdbg/wtdbg.xml
```
bgruening, line 1: is this needed?

repos.txt
bgruening, line 1: can this be removed

t
nilchia: I don't think. It was a byproduct of the function.
I've removed it.

bernt-matthias: Should have been removed automatically by `egrep "WARN|ERROR"` .. maybe double check everything?

nilchia: I tried the code again, it didn't remove it. :thinking: 
I'm not sure if `egrep "WARN|ERROR" t | cut -d'(' -f2 | cut -d')' -f1 | sort -u >> $line/.lint_skip;` removes the file.

bernt-matthias: If you want you can compare with https://github.com/bernt-matthias/galaxytools/tree/lint-skip .. or just take my branch

PR #1585, opened 2025-03-05
[MAFFT] add --anysymbol parameter

shiltemann:
I had some sequences with a U charachter in the protein sequence, this is not supported without the --anysymbol parameter


bgruening:
Sorry for the pain, the linting is now very strict for test cases, this will take us a while to fix all of those errors now :(

bgruening (approved):

bgruening:
Let me know if you need this update tomorrow.

shiltemann:
@bgruening need is a big word but it would be nice if it's not too much trouble

shiltemann:
awesome, thanks!

Issue #1586, opened 2025-03-05
Find and replace characters in chinese character


dadrasarmin:
Hi,

I was testing a tutorial developed by @Sch-Da and the results were not as expected. I narrowed down the problem to the [tool](https://github.com/bgruening/galaxytools/tree/master/tools/text_processing/text_processing) that is used to perform [find and replace](https://usegalaxy.eu/root?tool_id=toolshed.g2.bx.psu.edu/repos/bgruening/text_processing/tp_replace_in_line/9.3+galaxy1). It is based on `sed` and in the bottom of the help section of this tool it says:

> Note: SED uses extended regular expression syntax, not Perl syntax. \d, \w, \s etc. are not supported.

But what if I want to replace `\n`?  Using command line I am able to use something as follows ([source](https://backreference.org/2009/12/23/how-to-match-newlines-in-sed/index.html)):

```
sed -r --sandbox  -e ':begin;$!N;s/\n//;tbegin' test.txt
```

However, I cannot use this tool to put `begin;$!N;` before `s/\n//;tbegin` at the moment.  Is there a way to find and replace `\n` with this tool?

I found [another tool](https://usegalaxy.eu/root?tool_id=toolshed.g2.bx.psu.edu/repos/bgruening/text_processing/tp_find_and_replace/9.3+galaxy1) (based on Perl regex) on usegalaxy.eu that does the job. Maybe @Sch-Da decides to use that one. But it does not work with Chinese characters. For example, if she wants to find `(.)` and replace it with `$1\n` the output is not Chinese character. I tried `\p{Han}` and Unicode range for Chinese characters and it does not work. Therefore, this tool is not useful by itself.

The only solution at the moment is to mix the tools. Use one (Perl Regex) to replace newline `\n` and use the other one (sed Regex) to work with Chinese Characters. Based on my understanding, this is the correct results (output 49). Could maybe @Sch-Da confirm or reject my theory?

I personally find this approach (workflow) confusing and variable case-by-case. If we want to develop a community over Chinese (or any other non-Latin alphabet) text mining, we need a tool that can handle both cases of find-and-replace.

Cheers.

bgruening:
Is `sed ':a;N;$!ba;s/\n/ /g' file` working?

dadrasarmin:
On a Linux terminal yes. But how can I use this on useGalaxy.eu?
```
sed ':a;N;$!ba;s/\n//g' file
```

Sch-Da:
[...]

PR #1588, opened 2025-03-07
Added extra SED find and replace input option and test file

dadrasarmin:
As discussed [here](https://github.com/bgruening/galaxytools/issues/1586), sometimes there is a need to perform find and replace operation using sed on multiple lines. I added a new input option to answer this need. I also included a new test.


bgruening:
Thanks. The rest are linting errors. The linting get more strict in the last week. To see that locally, you need to run `planemo lint` but you also need to have the latest planemo release and the lastest galaxy-tool-utils package installed.

bgruening:
I added one commit from https://github.com/bgruening/galaxytools/pull/1520

tools/text_processing/text_processing/replace_text_in_line.xml
```diff
@@ -47,10 +48,20 @@
                     <add source="&#39;" target="&#39;&quot;&#39;&quot;&#39;" />
                     <add source="/" target="\/"/>
                 </mapping>
-
             </sanitizer>
-
-        </param>
+         </param>
+         <param name="sed_options" type="text" size="50" label="Additional sed commands before replacement" help="Provide additional sed commands before the replacement (e.g., ':a;N;$!ba;')." >
```
bgruening, line 53:
```suggestion
         <param name="sed_options" type="text" size="20" label="Additional sed commands before replacement" help="Provide additional sed commands before the replacement (e.g., ':a;N;$!ba;')." >
```

?

tools/text_processing/text_processing/replace_text_in_line.xml
```diff
@@ -155,7 +176,7 @@ The select tool searches the data for lines containing or not containing a match
 - **\|** Separates alternate possibilities.
 
 
-**Note**: SED uses extended regular expression syntax, not Perl syntax. **\\d**, **\\w**, **\\s** etc. are **not** supported.
+**Note**: SED uses extended regular expression syntax, not Perl syntax. **\\d**, **\\w**, **\\s** etc. are **not** supported. However, you can use SED FAQ to perform commands using special characters. Please read the SED FAQ here: https://www.pement.org/sed/sedfaq3.html#s3.2
```
bgruening, line 179: Can we add something like this:

More complex options can look like `sed -e '$!N;s/foo/bar/;'`. `$!N;` is an optional part which you only need to set in very special cases .....

dadrasarmin: Done.

tools/text_processing/text_processing/replace_text_in_line.xml
```diff
@@ -155,7 +176,9 @@ The select tool searches the data for lines containing or not containing a match
 - **\|** Separates alternate possibilities.
 
 
-**Note**: SED uses extended regular expression syntax, not Perl syntax. **\\d**, **\\w**, **\\s** etc. are **not** supported.
+**Note**: SED uses extended regular expression syntax, not Perl syntax. **\\d**, **\\w**, **\\s** etc. are **not** supported. However, you can use SED FAQ to perform commands using special characters.
+        More complex options can look like sed -e '$!N;s/foo/bar/;'. Here, $!N; is an optional part which you only need to set in very special cases
```
bgruening, line 180:
```suggestion
        More complex options can look like `sed -e '$!N;s/foo/bar/;'`. Here, `$!N;` is an optional part which you only need to set in very special cases. The `foo` part is the search string, and the `bar` part is the replacement string.
```

dadrasarmin:
I tried to correct all warnings that I could get from `planemo lint` on my local machine. I checked and the `planemo t --biocontainers` passes all tests. There was one error for the `recurring_lines.xml` tool which in my view the provided output was wrong (it assumed the number of lines 10 instead of 9). On my local test, both lint and test have successfully passed.

tools/text_processing/text_processing/macros.xml
bgruening, line 1: @dadrasarmin if you have no joy in bumping the main dependency "coreutils" to 9.5 ... please revert the commits.

But you need to bump the @VERSION_SUFFIX@ then from 1 to 2.
We always need to bump the version for tools.

bgruening:
Its still green, a miricle :) 

Can you maybe still check in each tool that all versions are respecting the macro version and there is no dangling hardcoded `galaxy1` or so...

tools/text_processing/text_processing/macros.xml
dadrasarmin: I did not know this. Thanks. I do not know much about how to properly bump the version and version suffix. Why we go to 9.5 after 9.3 instead of 9.4? Asking for educational purposes.

bgruening: We just go to the latest version. coreutils 9.5 is the latest version available in conda. https://anaconda.org/conda-forge/coreutils

bgruening:
Cool, thanks a lot!

dadrasarmin:
Thanks for your supervision and help.

PR #1597, opened 2025-03-25
recurring_lines.xml add tab as a valid character input

pavanvidem:


tools/text_processing/text_processing/recurring_lines.xml
```diff
@@ -21,7 +21,13 @@
     <inputs>
         <repeat name="token_set" title=" selection" min="1">
             <param name="line" type="text"
-                label="Characters to insert" help="Specify the characters that will be inserted X times in every line"/>
+                label="Characters to insert" help="Specify the characters that will be inserted X times in every line">
+                    <sanitizer invalid_char="">
```
mvdbeek, line 25:
```suggestion
                    <sanitizer>
```

mvdbeek:
Can you also add a test ?

tools/text_processing/text_processing/macros.xml
```diff
@@ -6,7 +6,7 @@
         </requirements>
     </xml>
     <token name="@TOOL_VERSION@">9.5</token>
-    <token name="@VERSION_SUFFIX@">0</token>
+    <token name="@VERSION_SUFFIX@">1</token>
```
mvdbeek, line 9: This will bump the entire suite, and in return will bump all workflows using any of the tools in the suite. Is it not possible to override @VERSION_SUFFIX@ in the tool ?

PR #1601, opened 2025-04-04
Fix `remote_repository_url` in bioimage.io inference tool

kostrykin:


bgruening:
Its a new linter: .. ERROR (ShedVersion): bioimage_inference: version 2.4.1+galaxy1 is less or equal than version of the latest installable revision 2.4.1+galaxy1

I think we can ignore this when you are fine that we do not push it to the TS. So we merge this PR and kill the CI

kostrykin:
Exactly

PR #1603, opened 2025-04-07
bioimage_inference change output format from tif to tiff

pavanvidem:


bgruening (approved):
`tif` is just wrong, we should always use `tiff` ...

PR #1604, opened 2025-04-08
Add LiraSearch tool

simonbray:
Completely WIP, the tool doesn't even have an official name yet.


simonbray:
@bgruening do you have an example of a tool that accesses a database? I don't think I had experience writing such a tool.

simonbray:
> @bgruening do you have an example of a tool that accesses a database? I don't think I had experience writing such a tool.

Hello @bgruening, I was wondering if you could comment on this?

What I meant with 'database' was anything accessed by the tool (could be e.g. a JSON file, but we would ideally like to use RocksDB: https://rocksdb.org/docs/getting-started.html). For a small file, I would put it into `$__tool_directory__`, but I guess this is a bad idea for something which is several GB in size.

bgruening:
The database is something like a reference that the tool needs during runtime? But is read-only?

bgruening:
In that case you can use location files (.loc) files. Those are tabular files that contain a bit of metadata for your "db" and the path to the DB. An admin can then configure this file and downloads the big DB and puts them somewhere locally (assuming the DB is a flat file, and not a server.)

simonbray:
> The database is something like a reference that the tool needs during runtime? But is read-only?

Yes, exactly. Thanks a lot :)

bgruening:
Similar to this tool: https://github.com/bgruening/galaxytools/blob/master/tools/wordcloud/test-data/fonts.loc

With a loc file for fonts.

chemicaltoolbox/lirasearch/lirasearch.xml
```diff
@@ -0,0 +1,63 @@
+<tool id="lirasearch" name="Identify similar conformers" version="0.1.0">
```
bgruening: add a profile please

chemicaltoolbox/lirasearch/lirasearch.xml
```diff
@@ -0,0 +1,63 @@
...
+    <description>using LiraSearch spherical harmonic descriptors</description>
+    <requirements>
+        <container type="docker">sb17/shcoeffs_w_ignite:latest</container>
```
bgruening: specific version?

chemicaltoolbox/lirasearch/lirasearch.xml
```diff
@@ -0,0 +1,63 @@
...
+                ln -s $ligand ligand.pdb &&
+            #else
+                python /opt/ignite/scripts/convert_sdf_to_pdb.py $ligand ligand.pdb &&
+            #end if
+            python -m esp_dnn.predict -m ligand -i . -o . &&
+            esp-surface-generator ligand.pdb.pqr ligand.pdb.tmesh &&
+            sh-coeff-calculator . &&
+            micromamba run -n lirasearch python /opt/ignite/scripts/lira_search_sdf.py --in_file rid_coeffs.csv --out_file output --database /opt/ignite/storage/rid_coeffs.csv --ignite_address 127.0.0.1:10800 --num_results $num_results --out_folder search_results  &&
+            micromamba run -n lirasearch python /opt/ignite/scripts/lira_super.py --target ligand.pdb --folder search_results &&
+            cat search_results/*rotated.sdf >> $output_sdf
```
bgruening:
```suggestion
            cat search_results/*rotated.sdf >> '$output_sdf'
```

simonbray:
@bgruening here is the Dockerfile, if you have any suggestions: https://github.com/simonbray/SHCoeffsDev/blob/main/Dockerfile

The image is 9 GB, but there is a lot in there😬

simonbray:
It is 6.5 GB now.

Still WIP :)

simonbray:
@bgruening I am a bit stuck with the loc file which I tried to implement in the last commit.

Tests are failing with `Parameter 'database_select': requires a value, but no legal values defined`, can you see if there is something I missed?

simonbray:
Oh, I missed the `tool_data_table_conf.xml.sample` file! Let me try adding it

bgruening:
> Caused by: java.io.FileNotFoundException: /opt/ignite/storage/node00-bfcdcb74-02d8-4121-b8f8-af2d7efdb83e/lock (Read-only file system)

bgruening:
Does that path belong to a different user? Does it have world writeable permissions?

simonbray:
> Does that path belong to a different user? Does it have world writeable permissions?

It is the mounted `test-data` directory (or to be precise a symlink to it).

When the tool is installed, will it be possible for the mounted data to be writeable? I remember we talked about it in Freiburg

bgruening:
Depends in the deployment, but I would always assume that most mounted volumes are read-only. Only the job dir is writable.

chemicaltoolbox/lirasearch/tool_data_table_conf.xml.sample
```diff
@@ -0,0 +1,6 @@
+<tables>
+    <table name="lirasearch" comment_char="#">
+        <columns>value, name, path</columns>
```
bgruening: please add a version column and filter by the version inside you tool ...

This will be good for the future, just imagine you are changing the DB layout, then new tools can only show DB that are fittting with the DB version.

simonbray:
@bgruening like so?

Issue #1608, opened 2025-04-14
BMZ inference tool, inconsistency with input data


kostrykin:
For this input image: https://github.com/bgruening/galaxytools/blob/cefdfdc13838de5108e13f54ecd69babb44009a1/tools/bioimaging/test-data/input_nucleisegboundarymodel.png

<img width="423" src="https://github.com/bgruening/galaxytools/blob/cefdfdc13838de5108e13f54ecd69babb44009a1/tools/bioimaging/test-data/input_nucleisegboundarymodel.png?raw=true" />

The [`bioimaging` tool](https://github.com/bgruening/galaxytools/tree/cefdfdc13838de5108e13f54ecd69babb44009a1/tools/bioimaging) produces this output:

<img width="423" alt="Image" src="https://github.com/user-attachments/assets/ac39c6a7-cc17-45dd-8368-4acbc0e9e59e" />

There seems to be an inconsistency with the rows/columns, maybe they are swapped unintentionally either for the input or for the output. It could also be a rotation in conjunction with mirroring, but I think this is rather unlikely.

There also is a test for this,
https://github.com/bgruening/galaxytools/blob/cefdfdc13838de5108e13f54ecd69babb44009a1/tools/bioimaging/bioimage_inference.xml#L68-L82

but the output is not really verified here.

@anuprulez Can you maybe help with this?

anuprulez:
@kostrykin sure, will have a look at it soon

PR #1609, opened 2025-04-15
Fix transposed dimensions

anuprulez:
This PR fixes https://github.com/bgruening/galaxytools/issues/1608

Tested with all the models in https://training.galaxyproject.org/training-material/topics/imaging/tutorials/process-image-bioimageio/tutorial.html

In the above tutorial, "NeuronSegmentationInEM" does not have any Pytorch model available at the mentioned link. I propose to change it to https://bioimage.io/#/?id=10.5281%2Fzenodo.5874741

ping @kostrykin


kostrykin (commented):
Thanks!

Can you maybe please use something like `<output name="output_predicted_image" file="result.tiff" ftype="tiff" compare="image_diff" />` for the tests?
https://github.com/bgruening/galaxytools/blob/973836fb40ecb9c0ac26f675d12b20fc8e5f51f4/tools/bioimaging/bioimage_inference.xml#L67-L82

This will be more reliable.

kostrykin (commented):
You also need to increment the `@VERSION_SUFFIX@`:
https://github.com/bgruening/galaxytools/blob/973836fb40ecb9c0ac26f675d12b20fc8e5f51f4/tools/bioimaging/bioimage_inference.xml#L5

kostrykin:
And another question, this is beyond the scope of this PR, but maybe we can discuss this :)

I tried to wrap my head around the `input_axes ` parameter in this tool, but I just don't get it. Why is this input required?
https://github.com/bgruening/galaxytools/blob/973836fb40ecb9c0ac26f675d12b20fc8e5f51f4/tools/bioimaging/bioimage_inference.xml#L40-L44

The only place that I could find where it is used is in
https://github.com/bgruening/galaxytools/blob/973836fb40ecb9c0ac26f675d12b20fc8e5f51f4/tools/bioimaging/main.py#L123
so it looks like you only need the number of axes?

If this is the case, I have two follow-up questions:

1. Wouldn't only two options be enough then, something like
  ```xml
     <option value="bcyx">Four axes (e.g., bcyx, byxc)</option> 
     <option value="bczyx">Five axes (e.g., bczyx)</option> 
  ```
2. Can't this information be extracted from the input image directly? I mean, if it's just the number of axes, it should be the number of dimensions of the numpy array of the image, shouldn't it?

anuprulez:
> Thanks!
> 
> Can you maybe please use something like `<output name="output_predicted_image" file="result.tiff" ftype="tiff" compare="image_diff" />` for the tests?
> 
> https://github.com/bgruening/galaxytools/blob/973836fb40ecb9c0ac26f675d12b20fc8e5f51f4/tools/bioimaging/bioimage_inference.xml#L67-L82
> 
> This will be more reliable.

I agree but with `file="results.tiff"`, we have to keep `results.tiff` files in the `test_data`, no? These tiff files are big (> 1 MB) and the related tests may not pass. For smaller files (<1MB) we can do it

kostrykin:
> Yes, it should work. I will make this change

Awesome!

> I agree but with `file="results.tiff"`, we have to keep `results.tiff` files in the `test_data`, no? These tiff files are big (> 1 MB) and the related tests may not pass. For smaller files (<1MB) we can do it

At least for this particular test, the file is just about 500 kB: [result.tiff.zip](https://github.com/user-attachments/files/19758631/result.tiff.zip)

tools/bioimaging/bioimage_inference.xml
```diff
@@ -69,7 +68,7 @@
             <param name="input_image_file" value="input_nucleisegboundarymodel.png"/>
             <param name="input_image_input_size" value="256,256,1,1"/>
             <param name="input_image_input_axes" value="bcyx"/>
-            <output name="output_predicted_image" ftype="tiff">
+            <output name="output_predicted_image" ftype="tiff" file="output_nucleisegboundarymodel.tiff" compare="image_diff">
                 <assert_contents>
                     <has_size size="524846" delta="110" />
                 </assert_contents>
```
kostrykin, line 74: With the `image_diff` check in place, I think we can remove this?

The `has_size` test is very heuristical, it might even raise a false-negative check failure.

kostrykin (approved):

tools/bioimaging/bioimage_inference.xml
anuprulez: sure!

kostrykin (approved):

PR #1612, opened 2025-04-22
ImajeJ change from tif to tiff

pavanvidem:


bgruening:
I would not deploy this change, it does not change anything substantial ... isn't it? So maybe revert the bump and we don't deploy to the TS?

pavanvidem:
only test data extension changed. I can revert the version bump

PR #1619, opened 2025-05-09
trim_galore -> 0.6.10 with updated tests

luke-c-sargent:
version bump to 0.6.10 with updated tests; i tried to narrow down the diff checking a bit; each deviation seems to cause 2 hits (maybe because its the unified `diff -u`?), but some things can and should be variable (python version and number of cores, e.g.), and one line seems to randomly reorder itself (these vary in order: `(count smallRNA: 0, count Nextera: 0, count Illumina: 0)`)... but it may be too restrictive (other things might vary that i did not notice in my test runs; also will the underlying version of cutadapt change?)

### DIFF
--- a/tools/trim_galore/macros.xml
+++ b/tools/trim_galore/macros.xml
@@ -1,6 +1,6 @@
 <macros>
-    <token name="@TOOL_VERSION@">0.6.7</token>
-    <token name="@VERSION_SUFFIX@">1</token>
+    <token name="@TOOL_VERSION@">0.6.10</token>
+    <token name="@VERSION_SUFFIX@">0</token>
     <xml name="requirements">
         <requirements>
             <requirement type="package" version="@TOOL_VERSION@">trim-galore</requirement>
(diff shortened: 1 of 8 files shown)


bgruening:
>  each deviation seems to cause 2 hits

This is true. It might help for future updates if you add `diff_lines=X` to the tests. This way we allow for small difference in the version numbers.

luke-c-sargent:
i am not entirely sure why this error is happening:
- https://github.com/bgruening/galaxytools/actions/runs/14936784218/job/41966613282?pr=1619
```
.. ERROR: Error 'HTTPSConnectionPool(host='www.bioinformatics.babraham.ac.ukprojects', port=443): Max retries exceeded with url: /trim_galore/ (Caused by NameResolutionError("<urllib3.connection.HTTPSConnection object at 0x7fb7bae615d0>: Failed to resolve 'www.bioinformatics.babraham.ac.ukprojects' ([Errno -2] Name or service not known)"))' accessing http://www.bioinformatics.babraham.ac.uk/projects/trim_galore/
```
the strings in question are all of the format `http://www.bioinformatics.babraham.ac.uk/projects/trim_galore/` but the `/` between the domain and uri path seems to be getting dropped, possible a parsing error in the CI scripts?

luke-c-sargent:
i am also confused by some of these test results; for example the latest run shows a failure on test 7, which has `16` for `lines_diff` but i only count 13 differences in the output? 🤯 either way, there seems to be enough variability to justify bumping the lines_diff up on other tests

bgruening:
Thanks @luke-c-sargent

PR #1624, opened 2025-05-16
Fix: updated vRhyme XML and macros to correct dependencies and version

Minamehr:
Hey, this is just follow-up to update the vRhyme.
Thanks,
Mina


bgruening:
Thanks!

PR #1630, opened 2025-06-03
Add bio.tools to blobtoolkit

pavanvidem:


tools/blobtoolkit/macros.xml
```diff
@@ -1,7 +1,11 @@
 <macros>
     <token name="@TOOL_VERSION@">4.0.7</token>
     <token name="@VERSION_SUFFIX@">2</token>
```
SaimMomin12, line 3:
```suggestion
    <token name="@VERSION_SUFFIX@">3</token>
```

pavanvidem:
Not sure if it is needed. From the usage perspective, nothing changed.

PR #1633, opened 2025-06-05
Update wordcloud.xml

Sch-Da:
Correct authorship.
According to the licence, https://github.com/amueller/word_cloud?tab=MIT-1-ov-file it is Andreas Mueller

### DIFF
--- a/tools/wordcloud/wordcloud.xml
+++ b/tools/wordcloud/wordcloud.xml
@@ -276,7 +276,7 @@ The bigger and bolder the word appears, the more often it's mentioned within a g
                 title={Word Clouds with Python},
                 author={Mueller, Andreas Christian},
                 year={2018},
-                url={https://amueller.github.io/word_cloud/}
+                url={https://github.com/amueller/word_cloud}
             }
         </citation>
     </citations>


wm75:
Why not also cite the github repo like in https://github.com/amueller/word_cloud/blob/main/CITATION.cff
You're citing the work, not it's documentation.

dadrasarmin:
I know you can more than one citation per tool. So, it theory and technically, it is possible to cite both GitHub and documentation. I am not saying what is correct from Good Scientific Practice point of view.

https://github.com/jmchilton/bwa-mem/blob/0425264039950bfd9ded06997a08cc8b4ee1ad8f/bwa-mem.xml#L348-L358

Sch-Da:
Ah ok, sorry, now I get it. I have no experience with tool citation, as you might have guessed. Therefore, I liked the easy download option in Galaxy and am happy for your insights what's most useful.

tools/wordcloud/wordcloud.xml
```diff
@@ -274,7 +274,7 @@ The bigger and bolder the word appears, the more often it's mentioned within a g
         <citation type="bibtex">
             @misc{amueller2018wordcloud,
                 title={Word Clouds with Python},
-                author={Amueller, Sebastian},
+                author={Mueller, Andreas Christian},
                 year={2018},
                 url={https://amueller.github.io/word_cloud/}
```
SaimMomin12:
```suggestion
                url={https://github.com/amueller/word_cloud}
```

PR #1637, opened 2025-06-15
Add fastspar tool

neo417:
This is one of the tool wrappers I have been working on with @Minamehr as part of my Bachelor's project.

We're pretty happy with this one and think it just needs your feedback / approval before publishing it on the tool shed.

### DIFF
--- a/tools/fastspar/fastspar_reduce.xml
+++ b/tools/fastspar/fastspar_reduce.xml
@@ -24,8 +24,16 @@
         <param argument="--pvalue" type="float" value="0.05" label="P-value threshold"/>
     </inputs>
     <outputs>
-        <data name="correlations" format="tabular" label="${tool.name} on ${on_string}: filtered_correlations.tsv"/>
-        <data name="pvalues" format="tabular" label="${tool.name} on ${on_string}: filtered_pvalues.tsv"/>
+        <data name="correlations" format="tabular" label="${tool.name} on ${on_string}: filtered_correlations.tsv">
+            <actions>
+                <action type="metadata" name="column_names" default="column,row,value" />
+            </actions>
+        </data>
+        <data name="pvalues" format="tabular" label="${tool.name} on ${on_string}: filtered_pvalues.tsv">
+            <actions>
+                <action type="metadata" name="column_names" default="column,row,value" />
+            </actions>
+        </data>
     </outputs>
     <tests>
         <test expect_num_outputs="2">


tools/fastspar/fastspar.xml
```diff
@@ -0,0 +1,107 @@
...
+            --exclude_iterations $exclude_iterations
+            --threshold $threshold
+            --seed $seed
+            --correlation '$correlation'
+            --covariance '$covariance'
+            --threads \${GALAXY_SLOTS:-1}
+            --yes
+    ]]></command>
+    <inputs>
+        <!-- Todo: What is the correct Otu format in galaxy? -->
```
bgruening: remove? Or ask @paulzierep ?

tools/fastspar/fastspar_pvalues.xml
```diff
@@ -0,0 +1,158 @@
...
+                   help="For meaningful p-values, the parameters used during bootstrapped correlation estimation should be identical to those used for the FastSpar run which produced the correlation matrix. &lt;br&gt;For your convenience you can choose to calculate the correlation matrix here. In that case the seed used for the calculation is the same one used for generating the bootstrapped samples.">
+                    <option value="new">Recalculate the correlation matrix</option>
+                    <option value="original">Use an existing correlation matrix</option>
+            </param>
+            <when value="new"/>
+            <when value="original">
+                <param name="input_file" type="data" format="tabular" argument="--correlation" label="Correlation table" help="The correlation matrix generated by the original FastSpar analysis."/>
+            </when>
+        </conditional>
+        <param argument="--number" type="integer" value="1000" min="10" max="10000" label="Number of bootstrap samples" help="Recommended minimum: 1000 bootstrap samples for robust estimation."/>
```
bgruening, line 70: recommended minimum is 1000, but we offer a min=10 is that ok?

tools/fastspar/fastspar_pvalues.xml
```diff
@@ -0,0 +1,158 @@
...
+        <param argument="--pseudo" type="boolean" truevalue="--pseudo" falsevalue="" label="Use pseudo p-values" help="If selected, pseudo p-values are calculated instead of exact p-values. This can provide faster estimates but may be less precise."/>
+    </inputs>
+    <outputs>
+        <data name="output_correlation" format="tabular" label="${tool.name} on ${on_string}: median_correlation.tsv">
+            <filter>correlation['select'] == "new"</filter>
+        </data>
+        <data name="output_covariance" format="tabular" label="${tool.name} on ${on_string}: median_covariance.tsv">
+            <filter>correlation['select'] == "new"</filter>
+        </data>
+        <data name="pvalues" format="tabular" label="${tool.name} on ${on_string}: pvalues.tsv"/>
```
bgruening, line 84: You could add column names to all tabular outputs as metadata if you like 

https://github.com/galaxyproject/tools-iuc/blob/039359757ad615f57028cab966b2b76f8ce8a199/tools/kaiju/kaiju.xml#L109

tools/fastspar/macros.xml
```diff
@@ -0,0 +1,26 @@
+<macros>
+    <token name="@TOOL_VERSION@">1.0.0</token>
+    <token name="@VERSION_SUFFIX@">galaxy0</token>
```
bgruening: the suffix is usually just the 0 ... then out bot can easily bump the integer

bgruening (commented):
Very cool, very high quality? You can also submit this tool to IUC if you like.

tools/fastspar/fastspar_pvalues.xml
neo417: I test the tool with 10 bootstrap samples to make it a little faster. But the test only runs if the parameters are valid.

tools/fastspar/fastspar_pvalues.xml
neo417: Is there a way to dynamically add column names based on the output file?

Otherwise would I have to write something like this for all outputs (as a macro)?
```
<actions>
    <action name="column_names" type="metadata" default="${','.join([line.strip('#').split('\t')[0] for line in open($otu_table.file_name, 'r')])}"/>
</actions>```

bgruening: Are the columns not stable per output? I'm does this specific output here have always different columns that can not be listed as a hardcoded list?

neo417: The order of the output rows/columns depend on the first column of the input otu_table. So all outputs have the same columns, but I'm not aware of a way to share the computation.

Only the sparse matrices of fastspar_reduce are always the same. I will add the metadata there.

bgruening: Ok, then we don't add it here

SaimMomin12:
We can close this, as the tool was merged on [IUC](https://github.com/galaxyproject/tools-iuc/pull/7059)

PR #1639, opened 2025-06-16
Add PhyloPhlAn tool

neo417:
This is the other tool I have been working on with @Minamehr for my Bachelor's project.

Because of the unexpected complexity and time constraints we agreed it would be enough for my Project to just wrap the main phylophlan script, but for reference I have included all my progress on the tool.

The main thing missing from all the tools is a way to access cached reference datasets, as I do not understand how to configure and test the .loc files. Your help with that would be much appreciated. Once I know how to add tool-data manually, I would be willing to at least finish up the parts I have started in my free time.

Relatedly, we decided early on that writing data managers for this tool would be out of scope or the project. PhyloPhlAn provides various scripts to download pre-identified core UniRef90 proteins, reference genomes from the Genbank repository and custom SGB databases. I am not sure if compatible references can already be downloaded by Galaxy or if these data sources are too large to cache with a data manager. I did not continue writing wrappers for them once I realized that even the indices are hundreds of MB in size.

Do you think it would be useful to publish just the main script now and add the remaining tools and data managers later?


SaimMomin12:
@neo417 Great addition! Would you like to move this tool to IUC, as also have MetaPhlAn on that repo?

PR #1649, opened 2025-07-05
Yolo prediction revise

mjoudy:
HI every one.

Here I made some changes on yolo-prediction tool:
- outputs updated: overlayer, annotations added. also for the detect mode mask file which produced a black image is removed.
- The help section is updated to be more informative.


tools/image_processing/yolo/yolo_prediction.xml
```diff
@@ -8,7 +8,7 @@
     <expand macro="requirements" />
     <command detect_errors="aggressive">
     <![CDATA[
-      export YOLO_CONFIG_DIR='\$HOME/.config/ultralytics' &&
+      export YOLO_CONFIG_DIR='HOME/.config/ultralytics' &&
```
bgruening: Any reason for this change?

tools/image_processing/yolo/yolo_prediction.xml
```diff
@@ -19,115 +19,193 @@
...
+            <param name="mode" value="segment" />
+            <param name="image_size" value="640" />
+            <param name="confidence" value="0.25" />
+            <param name="iou" value="0.45" />
+            <param name="max_det" value="300" />
+            
+            <output_collection name="txt_results" type="list">
+                <element name="bus" file="bus.txt" compare="sim_size" />
+            </output_collection>
+            <output_collection name="mask_results" type="list">
```
bgruening: Add count here? And in the other collections?

tools/image_processing/yolo/yolo_prediction.xml
```diff
@@ -19,115 +19,193 @@
...
+
+    **Prediction Mode**
+
+    Choose from two available options using the **mode** parameter:
+
+    - **detect** – Use this if your pre-trained model is a detection model. It outputs bounding boxes, class IDs, and confidence scores for objects detected in each image.
+
+    - **segment** – Use this if your pre-trained model is a segmentation model. It outputs both bounding boxes **and** pixel-level segmentation masks. Additional output files include grayscale TIFF masks.
+
+    ⚠️ **Important:** The selected mode must match the type of the model you upload. For example, if you are using a segmentation model (`*-seg.pt`), you must select `segment` mode. Using a mismatched mode and model will lead to errors or empty results.
```
bgruening, line 180: A general question, can we detect which type a module is from the content of the PT file?

mjoudy: It is possible and I worked on it a bit. but afterwards I thought for some models for any reason this could not be available and in this case it can cause errors.

PR #1650, opened 2025-07-05
fix yolo_training file extension bug

sunyi000:
fixing the file extension bug in yolo training tool and conversion tool


tools/image_processing/yolo/yolo_training.xml
```diff
@@ -13,11 +13,19 @@
       mkdir -p ./input_images ./input_yolo ./training ./runs ./models output_files && 
 
       #for $filename in $input_images:
-          ln -s '$filename' './input_images/${filename.element_identifier}' &&
+          #if $filename.element_identifier.endswith($filename.ext)
```
bgruening, line 16: Is it a problem for yolo when the filename ends with .jgp.jpg?

sunyi000: I just run it locally with file names like this, and they still work..
I wasn't sure initially, so just did a test run.

bgruening: Ok if it works you can simplify this and remove the if condition altogether or?

sunyi000:
> Have you seen this? #1649

yes, it's the other yolo_prediction tool. I haven't test the changes yet

tools/image_processing/yolo/yolo_training.xml
sunyi000: I got a question, i'm a bit confused...
my tests failed after removing the if else..
if input image is` image1.jpg.jpg` then the input yolo txt will need to be `image1.jpg.txt` ..

will this work in a workflow, because the previous tool outputs the yolo txt, before yolo_training, does not have file extension?

or should we just leave the if-else just to be safe?

I will try tomorrow.. thanks

sunyi000: Hi @bgruening 
I did a few more tests, I think we should leave the if-else here.

without the if-else
1. inputs:
      image1.jpg
      image1.txt

2. after creating the symlinks, it becomes
    image1.jpg.jpg - (no problem for the image)
    image1.txt.txt - (this caused error, because yolo expect it to be image1.jpg.txt)

bgruening:
@sunyi000 do you want to fix the '$training_params.model_url' error that you encountered as well? Making it mandatory maybe or move it out of the section? If you got confused many others maybe as well.

Also please sync with the other PR, the lint was also fixed there.


Thanks a lot. I guess we are the. Ready for merge

sunyi000:
> @sunyi000 do you want to fix the '$training_params.model_url' error that you encountered as well? Making it mandatory maybe or move it out of the section? If you got confused many others maybe as well.
> 
> Also please sync with the other PR, the lint was also fixed there.
> 
> Thanks a lot. I guess we are the. Ready for merge

Thank you! @bgruening 

- I have synced with #1649. I modified the `location` attribute of the model param in tests. In the other pr, it uses the file under `v0.0.0` which is the first ultralytics release, but in our suite, we use `v8.3.0` . It's the same file in this case, but I think it's better to use the v8.3.0 version

- i moved the `model_url` out of the section and made it required.

tools/image_processing/yolo/yolo_prediction.xml
```diff
@@ -8,7 +8,7 @@
     <expand macro="requirements" />
     <command detect_errors="aggressive">
     <![CDATA[
-      export YOLO_CONFIG_DIR='\$HOME/.config/ultralytics' &&
+      export YOLO_CONFIG_DIR='HOME/.config/ultralytics' &&
```
bgruening:
```suggestion
      export YOLO_CONFIG_DIR="\$HOME/.config/ultralytics" &&
```

bgruening (approved):

PR #1652, opened 2025-07-07
aggregated updates to gnu-coreutils

nilchia:
This tool sorts rows and then removes duplicates. It is problematic when the first row is the header.
This PR adds the option to exclude first row as header


nilchia:
@wm75 Hi, Can you please review this? Thanks

wm75:
Can you compare this to the `tp_sort_header_tool` please. I think it has all the functionality you need.
We do want to keep `unsorted_uniq` because it presents the same underlying tool from a different user perspective and so makes things more discoverable.
However, please try to simply copy functionality over from the other tool. Makes it easier to understand the logic.
In other words, steal the "number of header lines" functionality and the nice `sed` trick from the other tool :-)

nilchia:
@wm75 Thanks for the comment.
The other tool does not work for me. Because I do not have "duplicate rows" in my data, but "duplicate rownames". So if I only use this tool and set the start and end column to the first column, I can remove duplicate rownames.

I added the functionality of the other tool for the header here.

tools/text_processing/text_processing/unsorted_uniq.xml
```diff
@@ -8,15 +8,20 @@
...
-    sort -u
-        $ignore_case
-        $is_numeric
-        -t '	'
-        #if $adv_opts.adv_opts_selector == "advanced":
-            -k$adv_opts.column_start,$adv_opts.column_end
+    (
+        export LC_ALL=C;
+        #if int($header) > 0:
+            sed '${header}'q &&
```
wm75:
```suggestion
        #if int($adv_opts.header) > 0:
            sed -u '${adv_opts.header}'q &&
```

nilchia:
>The sort tool has start/end columns, too, hasn't it? Anyway, bringing this tool here to about the same functionality is a good thing.

Ahhh, I was looking at the wrong tool :)

Do you mean this tool? https://usegalaxy.eu/root?job_id=11ac94870d0bb33a57d479a5f8efbf60

Yes, this actually works.

Maybe that tool needs a better description or name so the user knows it can return unique values. :thinking:

nilchia:
BTW, why is this tool's id "tp_sorted_uniq" and its XML is named "unsorted_uniq.xml" :thinking:

tools/text_processing/text_processing/unsorted_uniq.xml
```diff
@@ -52,6 +63,24 @@
...
+            <conditional name="adv_opts">
+                <param name="adv_opts_selector" value="advanced"/>
+                <param name="column_start" value="2"/>
+                <param name="column_end" value="3"/>
+                <param name="header" value="1"/>
+            </conditional>
+            <output name="outfile" file="unique_results2.bed"/>
+        </test>
+        <test>
+            <param name="infile" value="1_dup.bed"/>
```
wm75, line 77: This test could be made more conclusive if the test input file didn't start with the smallest value. It would then demonstrate that without the advanced section the first line gets included in the sort.

wm75:
Hmm, I'm noticing this only now, which is bad, sorry: the way you've implemented this, you're tying together the header option with the column selection (because the column selection isn't optional, but on by default as soon as you're opting into advanced options).
So what if a user only wants to ignore a bunch of header lines, but wants to base the sort on the entire line?

I feel a bit bad for asking for this now, but it seems the better way would be to move the "header" option out of the advanced section :face_with_peeking_eye:

tools/text_processing/text_processing/unsorted_uniq.xml
nilchia: True!
I updated the input.

nilchia:
@bgruening  That's true.

nilchia:
Thanks a lot for your kind review :hugs: @wm75

wm75:
@nilchia @RZ9082 stable enough state for a final round of reviewing?

nilchia:
> @nilchia @RZ9082 stable enough state for a final round of reviewing?
> 

Yes

tools/text_processing/text_processing/sort.xml
```diff
@@ -28,36 +36,38 @@
...
+            <param name="style" type="select" display="radio" label="using sort flavor">
                 <option value="n">Fast numeric sort (-n)</option>
                 <option value="g">General numeric sort ( scientific notation -g)</option>
                 <option value="V">Natural/Version sort (-V) </option>
                 <option value="">Alphabetical sort</option>
                 <option value="h">Human-readable numbers (-h)</option>
                 <option value="R">Random order (-R)</option>
             </param>
+            <param name="ignore_case" type="boolean" checked="false" truevalue="f" falsevalue=""
+                   label="ignoring case" help="Turn lowercase symbols to upper case before comparing values in this column. (-f)" />
```
bgruening, line 60: uha, that is interesting

tools/text_processing/text_processing/sort.xml
```diff
@@ -28,36 +36,38 @@
...
-            label="Number of header lines" help="These will be ignored during sort.">
+            label="Number of header lines" help="Header lines will be copied to the output unchanged without operating on them.">
             <validator type="in_range" message="Negative values are not allowed." min="0"/>
         </param>
-
+        
         <repeat name="sortkeys" title="Column selections" min="1">
-            <param name="column" label="on column" type="data_column" data_ref="infile" accept_default="true" />
+            <param name="column" label="Sort on column" type="data_column" data_ref="infile" accept_default="true" />
+            <param name="start_charpos" label="considering its characters from" type="integer" min="1" optional="true" help="Leave empty (or set to 1) to use the column value starting from its first character." />
```
bgruening, line 45: Maybe add an example? The use-case would be to only sort the first 4 letters from `2026-01-05`.

bgruening (approved):

PR #1655, opened 2025-07-14
yolo train/predict bug fix

sunyi000:
removing shutil completly to fix `device or resource busy ` error.

@bgruening could we try this? sorry for the multiple tries..

Thank you


sunyi000:
> I will look at it at the evening. Is the galaxy+1 version working?
> 
Thank you .
I haven't run galaxy+1. I will try that.

When will galaxy+3 be available? I will also test galaxy+3 once it is out.

PR #1669, opened 2025-08-20
Add LLM Hub tool

arash77:
Introduce a new Galaxy tool that provides access to various Large Language Models (LLMs) via a LiteLLM proxy server. Include configuration instructions and a sample data file for model definitions. The tool allows users to interact with multiple LLM APIs using a unified interface.

PR #1670, opened 2025-08-20
Update data table references for LLM models

arash77:
An update that was not applied on https://github.com/bgruening/galaxytools/pull/1669. Change data table references from "litellm" to "llm_models".

### DIFF
--- a/tools/llm_hub/llm_hub.xml
+++ b/tools/llm_hub/llm_hub.xml
@@ -2,7 +2,7 @@
     <description>Call any LLM</description>
     <macros>
         <import>macros.xml</import>
-        <token name="@VERSION_SUFFIX@">0</token>
+        <token name="@VERSION_SUFFIX@">1</token>
         <token name="@PROFILE@">24.0</token>
     </macros>
     <requirements>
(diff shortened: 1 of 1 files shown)


bgruening:
Can you please also do something like that:

```
with open(os.environ.get("LITELLM_CONFIG_FILE"), "r") as f:
    config = yaml.safe_load(f)
    
litellm_api_key = config.get("LITELLM_API_KEY")
litellm_base_url = config.get("LITELLM_BASE_URL")

**Comment by arash77** [comment]
It's ready.

File: tools/llm_hub/llm_hub.xml [hunk 1]
```diff
@@ -38,21 +39,21 @@ python '$__tool_directory__/llm_hub.py'
...
                 <param name="model" type="select" optional="false" label="Model" help="Select the model you want to use.">
-                    <options from_data_table="litellm">
+                    <options from_data_table="llm_models">
                         <filter type="static_value" column="2" value="text"/>
                     </options>
-                    <validator message="No model annotation is available for LiteLLM" type="no_options"/>
+                    <validator message="No model annotation is available for LLM Hub" type="no_options"/>
                 </param>
-                <param name="context" type="data" multiple="true" optional="false" format="docx,html,json,pdf,txt" label="Context" max="500"/>
+                <param name="context" type="data" multiple="true" optional="false" format="html,json,txt" label="Context" max="500"/>
```
anuprulez, line 56: Why is "context" field not optional? We should be able to receive answers without providing additional context and just by writing text in the prompt field?

arash77: It is the same as the ChatGPT tool. I was thinking to keep it that way, but if it’s not necessary to enforce, I could remove the restriction.

bgruening: @anuprulez do you have a use-case for this? Do we want to enable the chat-bot use-case?

anuprulez:
LLM_explain


invoked_wf

anuprulez:
Output of LLM hub tool on the intermediate tabular results of hyperparameter optimization

hyper_results

PR #1672, opened 2025-08-26
Add handling for image models

arash77:
Introduce support for image models.
After this update image-to-text models like this could be add to the loc file:
```
numarkdown-8b-thinking-llmlb	Advanced OCR and document understanding, excels at extracting structured data from scanned images (NuMarkdown-8B-Thinking)	image	uni-freiburg	NuMind
```

### DIFF
--- a/tools/llm_hub/llm_hub.xml
+++ b/tools/llm_hub/llm_hub.xml
@@ -17,10 +17,10 @@
 #set LINK_LIST = []
 #for $input in $context
     #set file_name = os.path.splitext($input.element_identifier)[0]
-    #set ext = $input.ext if $input.ext in ['html', 'json', 'txt', 'jpg', 'jpeg', 'png', 'gif'] else 'txt'
+    #set ext = $input.ext if $input.ext in ['html', 'json', 'txt', 'jpg', 'jpeg', 'png', 'gif', 'tiff'] else 'txt'
     #set LINK = re.sub('[^\w\-]', '_', $file_name)+'.'+$ext
     ln -s '$input' '$LINK' &&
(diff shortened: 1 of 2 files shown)


tools/llm_hub/llm_hub.xml
```diff
@@ -54,11 +61,20 @@ python '$__tool_directory__/llm_hub.py'
...
+                </param>
+            </when>
+            <when value="image">
+                <param name="model" type="select" optional="false" label="Model" help="Select the model you want to use.">
+                    <options from_data_table="llm_models">
+                        <filter type="static_value" column="2" value="image"/>
+                    </options>
+                    <validator message="No model annotation is available for LLM Hub" type="no_options"/>
+                </param>
+                <param name="context" type="data" multiple="true" optional="false" format="jpg,png,gif" label="Context" max="500"/>
```
bgruening: tiff?

bgruening: I think he explicitly said only one image at a time is possible, no?

arash77: Yes, and it seems that for images in all models, we have this limitation.

arash77: Does it make sense to have separate inputs for images and other input types, so that we can limit images to only one?

bgruening (approved):

arash77:
I updated it to now handle images and text contexts separately.
Also, the output is now with the model name (we could also add a short text column to the loc file to use that one for better appearance. Also, if there isn't any context provided, it doesn't show the `on ` part in the output anymore.
image

tools/llm_hub/llm_hub.xml
arash77: > tiff?

I added TIFF and BMP file formats.

PR #1674, opened 2025-08-29
Add  imagemagick tools for image Manipulation

Delphine-L:
This tool is similar to GraphicsMagick for image manipulation, but I couldn't get  GraphicsMagick to work with any text addition without manually editing configuration files. ImageMagick does the same thing as GraphicsMagick, but it is way easier to work with.  

- Montage : Merge images. Addition from the GraphicsMagick tool : can add titles and labels to images, and resize. 
- Convert: Same as GraphicsMagick, I can remove it if it's redundant.
- Compare: different methods of comparison compared to  GraphicsMagick. Also, add the output of statistics and composite images.


bgruening:
Interesting, maybe our bet on GraphicsMagick was the wrong one :)

tools/image_processing/imagemagick/convert.xml
```diff
@@ -0,0 +1,148 @@
+<tool id="graphicsmagick_image_convert" name="Convert image format" version="@TOOL_VERSION@+galaxy@VERSION_SUFFIX@" profile="22.05">
```
bgruening:
```suggestion
<tool id="imagemagick_image_convert" name="Convert image format" version="@TOOL_VERSION@+galaxy@VERSION_SUFFIX@" profile="22.05">
```

PR #1679, opened 2025-09-16
Update flexynesis

nilchia:
This PR updates flexynesis to the newest version.
Features:
1. faster module loading
2. saving model as safetensors

### DIFF
--- a/tools/flexynesis/flexynesis.xml
+++ b/tools/flexynesis/flexynesis.xml
@@ -580,7 +580,7 @@
             </output_collection>
             <output name="model" ftype="safetensors">
                 <assert_contents>
-                    <has_size size="107104"/>
+                    <has_size value="107104"/>
                 </assert_contents>
             </output>
         </test>
@@ -766,7 +766,7 @@
             </output_collection>
             <output name="model" ftype="safetensors">
                 <assert_contents>
-                    <has_size size="198076"/>
+                    <has_size value="198076"/>
                 </assert_contents>
             </output>
         </test>
@@ -826,7 +826,7 @@
             </output_collection>
             <output name="model" ftype="safetensors">
                 <assert_contents>
-                    <has_size size="194476"/>
+                    <has_size value="194476"/>
                 </assert_contents>
             </output>
         </test>


tools/flexynesis/flexynesis.xml
```diff
@@ -494,14 +495,15 @@
         <collection name="results" type="list" label="${tool.name} on ${on_string}: results">
             <discover_datasets pattern="(?P&lt;name&gt;.+)\.tabular$" format="tabular" directory="output"/>
         </collection>
+        <data name="model" format="safetensors" from_work_dir="output/job.final_model.safetensors" label="${tool.name} on ${on_string}: trained_model" />
```
bgruening: Is this always needed? Should we make this optional?

tools/flexynesis/flexynesis.xml
```diff
@@ -1311,16 +1330,17 @@
...
+                        <has_image_center_of_mass center_of_mass="972,737" eps="40"/>
                         <has_image_channels channels="3"/>
-                        <has_image_height height="1461" delta="10"/>
-                        <has_image_width width="1941" delta="10"/>
+                        <has_image_height height="1500" delta="20"/>
+                        <has_image_width width="1980" delta="20"/>
                     </assert_contents>
                 </element>
             </output_collection>
+            <output name="model" ftype="safetensors" file="model_18.safetensors" compare="sim_size"/>
```
bgruening: you don't need to test this output here, I guess one of those outputs is enough? sim_size is not a very reliable test

tools/flexynesis/macros.xml
```diff
@@ -1,6 +1,6 @@
 <macros>
-    <token name="@TOOL_VERSION@">0.2.20</token>
-    <token name="@VERSION_SUFFIX@">3</token>
+    <token name="@TOOL_VERSION@">1.1.1</token>
+    <token name="@VERSION_SUFFIX@">0</token>
     <token name="@PROFILE@">24.1</token>
```
bgruening: 25.0 ... if your datatype gets merged?

nilchia:
Still work in process. The inference mode should be added. Please don't review :)

tools/flexynesis/flexynesis.xml
```diff
@@ -146,7 +146,9 @@
             $use_cv
             $evaluate_baseline_performance
             --feature_importance_method $feature_importance_method
-            \${GALAXY_FLEXYNESIS_EXTRA_ARGUMENTS}
```
bgruening, line 149: why do you remove this? We can now not set the --device to GPU

tools/flexynesis/flexynesis.xml
```diff
@@ -146,7 +146,9 @@
...
+            --safetensors
+            --device 'cpu'
+            --threads \${GALAXY_SLOTS:-4}
```
SaimMomin12:
```suggestion
            --threads "\${GALAXY_SLOTS:-4}"
```

tools/flexynesis/flexynesis.xml
```diff
@@ -572,9 +575,10 @@
                     </assert_contents>
                 </element>
             </output_collection>
+            <output name="model" ftype="safetensors" file="model_1.safetensors" compare="sim_size"/>
```
SaimMomin12: Is there any other way to test this apart from sim_size?

tools/flexynesis/flexynesis.xml
```diff
@@ -882,9 +890,10 @@
                     </assert_contents>
                 </element>
             </output_collection>
+            <output name="model" ftype="safetensors" file="model_6.safetensors" compare="sim_size"/>
```
bgruening: don't use sim-size maybe assert a size, this way we do not need to ship the safetonsor files in the git repo here.

tools/flexynesis/flexynesis.xml
nilchia: I don't think so.

SaimMomin12: maybe some asserts? (Easiest: size?)

tools/flexynesis/flexynesis.xml
nilchia: No we can make it optional.
But I think the optional should be set to True as default.

SaimMomin12: +1 for creating a boolean parameter and making this output optional

tools/flexynesis/flexynesis.xml
nilchia: I don't know of other test for this data.
The size of them are different though.

tools/flexynesis/flexynesis.xml
nilchia: Ok :+1:

tools/flexynesis/flexynesis.xml
SaimMomin12: > No we can make it optional. But I think the optional should be set to True as default.

This would mean your safetensor model will always be a output dataset unless the user unselects it. Do we want that?

nilchia: >This would mean your safetensor model will always be a output dataset unless the user unselects it. Do we want that?
Yes, it will reduce the unnecessary job repetition.  Imagine you are training a big data which takes hours to finish, and then you realize the trained model is not saved.

Especially in the next updates, where it is possible to use the model to integrate new data.

For me, the model is not a byproduct.

tools/flexynesis/flexynesis.xml
```diff
@@ -580,7 +580,7 @@
             </output_collection>
             <output name="model" ftype="safetensors">
                 <assert_contents>
-                    <has_size size="107104"/>
+                    <has_size size="20"/>
```
nilchia: Probably `sim_size` was a better option.
Why is this passing?

bgruening: why?

nilchia: The test passes with both `size="107104"` and `size="20"`

wm75: I don't know for sure, but I assume because of https://github.com/bgruening/galaxytools/actions/runs/18556838587/job/52896358516?pr=1679#step:7:494
basically with the surrounding ftype check not being executable, the inner check won't run?

nilchia: I tried this locally and it passed:
```
                <element name="job.stats">
                    <assert_contents>
                        <has_size size="123456789"/>
                        <has_text_matching expression="DirectPred\tErlotinib\tnumerical\tmse\t"/>
                        <has_text_matching expression="DirectPred\tErlotinib\tnumerical\tr2\t"/>
                        <has_text_matching expression="DirectPred\tErlotinib\tnumerical\tpearson_corr\t"/>
                    </assert_contents>
                </element>
```

bgruening: Interesting. If you think its a bug, please look at one of those Galaxy internal test:

https://github.com/search?q=repo%3Agalaxyproject%2Fgalaxy+assert_contents+path%3A%2F%5Etest%5C%2F%2F+size+language%3AXML&type=code&l=XML

Try to replicated it and create a test that clearly shows its not considerd. If people have a test to work against it, they can fix it.

bgruening: But, now I see it ... its not size ... its value :)

nilchia: https://github.com/galaxyproject/galaxy/pull/21081

bgruening (approved):

PR #1680, opened 2025-09-29
add --no-version-check for salmon quant

jianzuoyi:
The `salmon quant` command will check the latest salmon version, but when the network is poor, the check may fail and write the message  `Version Info Exception: server did not respond before timeout` to logs. This can cause the galaxy tool to fail.

image

Therefore, it's better to add --no-version-check to disable version checking: `salmon quant --no-version-check`

### DIFF
--- a/tools/salmon/macros.xml
+++ b/tools/salmon/macros.xml
@@ -414,6 +414,7 @@
         #end if
 
         salmon quant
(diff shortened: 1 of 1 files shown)


jianzuoyi:
Very well, thank you so much @bernt-matthias .

PR #1684, opened 2025-10-13
Adding 3Dtrees: Tile & Merge tool

kgerb:


tools/3Dtrees_tile_merge/til_mer.xml
```diff
@@ -0,0 +1,116 @@
+<tool id="3dtrees_tile_merge" name="3Dtrees: Tile and Merge" version="1.0.1" profile="24.2">
+    <description>Subsample, tile and merge 3D point cloud data for tree segmentation.
```
bgruening: keep it short and crisp, no sentences if possible.

tools/3Dtrees_tile_merge/til_mer.xml
```diff
@@ -0,0 +1,116 @@
+<tool id="3dtrees_tile_merge" name="3Dtrees: Tile and Merge" version="1.0.1" profile="24.2">
+    <description>Subsample, tile and merge 3D point cloud data for tree segmentation. 
+    </description>
+    <requirements>
+        <container type="docker">ghcr.io/3dtrees-earth/3dtrees_tile_merge:1.0/container>
```
bgruening: if you want to keep the container and tool version in sync, you can use a macro here, let me know if you need more pointers to any of my recommendations.

tools/3Dtrees_tile_merge/til_mer.xml
```diff
@@ -0,0 +1,116 @@
...
+            --min-cluster-size '$operation.min_cluster_size'
+            --initial-radius '$operation.initial_radius'
+            --max-radius '$operation.max_radius'
+            --radius-step '$operation.radius_step'
+            --number-of-threads \${GALAXY_SLOTS:-4}
+        #end if
+    ]]>
+    </command>
+    <inputs>
+        <param name="input" type="data" format="zip,binary" label="Input Point Cloud or ZIP file" help="Input LAS/LAZ point cloud file or ZIP file containing prepared files"/>
```
bgruening: I assume its alway a LAZ file, even if they are in zip, correct?

I think what you want here is `multiple="true"`. This would allow a user to put in multiple single LAZ files, or a collection of LAZ files.

tools/3Dtrees_tile_merge/til_mer.xml
```diff
@@ -0,0 +1,116 @@
...
+        <conditional name="operation">
+            <param name="task" type="select" label="Task">
+                <option value="tile">Tile</option>
+                <option value="merge">Merge</option>
+            </param>
+            <when value="tile">
+                <param name="tile_size" type="integer" value="50" label="Tile Size" help="Size of tiles in meters"/>
```
bgruening, line 42: maybe add a min/max here?

bgruening: and to all other params?

tools/3Dtrees_tile_merge/til_mer.xml
```diff
@@ -0,0 +1,116 @@
...
+                <param name="max_radius" type="float" value="5.0" label="Maximum Search Radius (m)" help="Maximum radius for point reassignment search (default: 5.0m)"/>
+                <param name="radius_step" type="float" value="1.0" label="Radius Step (m)" help="Radius increment step for point reassignment (default: 1.0m)"/>
+            </when>
+        </conditional>
+    </inputs>
+    <outputs>
+        <data name="output_tile" format="zip" label="Prepared Files" from_work_dir="prepared_files.zip">
+            <filter>operation['task'] == "tile"</filter>
+        </data>
+        <data name="output_merge" format="binary" label="Merged Point Cloud" from_work_dir="final_pc.laz">
```
bgruening: I think we should add the LAZ file as proper Galaxy datatype, do we?

Can you give me some information about LAZ or LAS?

tools/3Dtrees_tile_merge/til_mer.xml
```diff
@@ -0,0 +1,116 @@
...
+            <param name="input" value="test_pc_nopred.laz"/>
+            <param name="operation|task" value="tile"/>
+            <param name="operation|tile_size" value="50"/>
+            <param name="operation|overlap" value="20"/>
+            <param name="operation|tiling_threshold" value="3"/>
+            <param name="operation|points_threshold" value="1000"/>
+            <param name="operation|subsampling_resolution" value="10"/>
+            <output name="output_tile" file="prepared_files.zip">
+                <assert_contents>
+                    <has_size value="0" delta="0" negate="true"/>
```
bgruening: That test is not testing much :)

You can for example test of members in the ZIP file: https://docs.galaxyproject.org/en/latest/dev/schema.html#id255

But I guess we want to get rid of ZIP files (maybe?), see above

tools/3Dtrees_tile_merge/til_mer.xml
```diff
@@ -0,0 +1,116 @@
...
+                </assert_contents>
+            </output>
+        </test>
+        <test expect_num_outputs="1">
+            <param name="input" value="processed_files.zip"/>
+            <param name="operation|task" value="merge"/>
+            <output name="output_merge" file="final_pc.laz">
+                <assert_contents>
+                    <has_size value="0" delta="0" negate="true"/>
```
bgruening: depends on how fancy we want to be in the future, but we could implement asserts that work on LAZ file:

https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/tool_util/verify/asserts/hdf5.py

tools/3Dtrees_tile_merge/til_mer.xml
```diff
@@ -0,0 +1,116 @@
...
+                </assert_contents>
+            </output>
+        </test>
+    </tests>
+    <help>
```
bgruening, line 92: by default a help is written in rst - restructured text, you can set the format="markdown"

PR #1689, opened 2025-10-16
update bigwig_to_bedgraph dependency

martenson:
mostly because there is no biocontainer for v312

https://biocontainers.pro/tools/ucsc_tools


bgruening:
But its here: https://quay.io/repository/biocontainers/ucsc_tools?tab=tags

bgruening (approved):

Issue #1688, opened 2025-10-16
Problem with the cellpose headless tool and TIFF input


rmassei:
Hi, 
I noticed some problems when processing TIFFs using the Cellpose headless tool ([here](https://github.com/FAIR-imaging/galaxy-image-community/issues/15#issuecomment-3404738044) some discussion already going on with @kostrykin).

**Problem:**
In short: Since TIFFs are very flexible, axis can be ordered arbitrarily. The Cellpose wrapper currently just assumes that TIFF channels are the third axis which is often leading to the tool to fail. Moreover, even if a channel axis is present, and it is the third axis, there is no guarantee that there are three channels included. It might as well be just a single one, which, again, would make the wrapper fail. 

**Potential solution:**
Considering this, a solution is to just simplify the tool by accepting as an input just single channel images. In case of, multichannel TIFFs, single channel can be extracted using another tool such as [this](https://usegalaxy.eu/root?tool_id=toolshed.g2.bx.psu.edu/repos/imgteam/split_image/ip_split_image/2.2.3+galaxy1)) and then processed

rmassei:
**Update**:
[new PR](https://github.com/bgruening/galaxytools/pull/1695#issuecomment-3474905700) to also add Cellpose v.4 (Cellpose-SAM) which might solve the issue of the channel. 

According to the [documentation](https://cellpose.readthedocs.io/en/latest/settings.html):

> Channels are no longer an input to Cellpose-SAM - Cellpose-SAM has been trained to be invariant to the order of the channels in your image. Cellpose-SAM will use the first 3 channels of your image, truncating the rest. It has been trained with three different channels for H&E images, and for cellular images it has been trained with the cytoplasm and nuclear channels in any order, with the other channel set to zero

Might make sense to close [this PR ](https://github.com/bgruening/galaxytools/pull/1692)without update and just add the new Cellpose-SAM

PR #1692, opened 2025-10-22
fix: cellpose headless tool

rmassei:
Hi,
based on the issues https://github.com/issues/assigned?issue=FAIR-imaging%7Cgalaxy-image-community%7C15 and https://github.com/bgruening/galaxytools/issues/1688 this PR:

- Modify the Cellpose tool to accept just one channel as input. This avoid to have problems with multi-channel TIFFs.
- Add some modification in the tool xml by clarifying the Cellpose version (3.1) and improve the Help section

closes https://github.com/bgruening/galaxytools/issues/1688

### DIFF
--- a/tools/cellpose/cellpose.xml
+++ b/tools/cellpose/cellpose.xml
@@ -1,5 +1,5 @@
 <tool id="cellpose" name="Run generalist cell and nucleus segmentation" version="@TOOL_VERSION@+galaxy@VERSION_SUFFIX@" profile="23.02">
-    <description>with Cellpose</description>
+    <description>with Cellpose 3</description>
     <macros>
         <token name="@TOOL_VERSION@">3.1.0</token>
         <token name="@VERSION_SUFFIX@">2</token>


tools/cellpose/cellpose.xml
```diff
@@ -1,14 +1,8 @@
 <tool id="cellpose" name="Run generalist cell and nucleus segmentation" version="@TOOL_VERSION@+galaxy@VERSION_SUFFIX@" profile="23.02">
-    <description>with Cellpose</description>
+    <description>with Cellpose 3.1</description>
```
bgruening: that is very unusual for Galaxy tools

tools/cellpose/cellpose.xml
```diff
@@ -126,23 +100,23 @@
...
-            <param name="chan" value="2"/>
-            <param name="chan2" value="3"/>
             <param name="show_segmentation" value="false"/>
             <output name="cp_mask" file="img02_cp_masks_cyto.tif" compare="image_diff"/>
         </test>
     </tests>
     <help>
         <![CDATA[
-        Cellpose: A generalist algorithm for cell and nucleus segmentation.
+        cCellpose is an anatomical segmentation algorithm written in Python 3 by Carsen Stringer and Marius Pachitariu.
```
bgruening:
```suggestion
        Cellpose is an anatomical segmentation algorithm written in Python 3 by Carsen Stringer and Marius Pachitariu.
```

are those changes relevant for the end user?

tools/cellpose/cellpose.xml
```diff
@@ -27,7 +21,6 @@
         python '$__tool_directory__/cp_segmentation.py'
             --inputs '$inputs'
             --img_path ./image.${img_in.ext}
-            --img_format '${img_in.ext}'
```
bgruening, line 30: is that not needed anymore?

tools/cellpose/cellpose.xml
rmassei: Ok, we can go back to the previous description :)

tools/cellpose/cellpose.xml
rmassei: Not anymore since before it was just checking if the format was TIFF and, in case, accepting different channels.
The tool now is accepting just single channel images,

tools/cellpose/cellpose.xml
```diff
@@ -74,48 +60,36 @@
...
-            <param name="chan2" value="3"/>
             <output name="cp_mask" file="img02_cp_masks_cyto2.tif" compare="image_diff"/>
             <output name="cp_segm" file="img02_cp_segm_cyto2.png" compare="image_diff"/>
         </test>
         <test expect_num_outputs="2">
             <param name="img_in" value="img02.png"/>
-            <param name="model_type" value="cyto3"/>
-            <param name="chan" value="2"/>
-            <param name="chan2" value="3"/>
+            <param name="model_type" value="cyto3"/>/>
```
kostrykin: This is what the linter complains about:
https://github.com/bgruening/galaxytools/actions/runs/19647211770/job/56265663207?pr=1692#step:7:431

```suggestion
            <param name="model_type" value="cyto3"/>
```

kostrykin:
> I think the issue might be related to the test datasets, all images have 3 channels....

Can you replace them?

tools/cellpose/cellpose.xml
```diff
@@ -72,77 +58,70 @@
...
+            <param name="img_in" value="img04_XYC_nuclei.tiff"/>
+            <param name="model_type" value="nuclei"/>
+            <output name="cp_mask" file="img04_XYC_nuclei_masks.tif" compare="image_diff"/>
+            <output name="cp_segm" file="img04_XYC_nuclei_segm.png" compare="image_diff"/>
         </test>
     </tests>
     <help>
         <![CDATA[
         Cellpose: A generalist algorithm for cell and nucleus segmentation.
+        **Important**
```
kostrykin: Fix punctuation:
```suggestion
        **Important:**
```

Image

tools/cellpose/cellpose.xml
```diff
@@ -72,77 +58,70 @@
...
+        This tool accepts single-channel images with coordinates XYC for segmentation. In case of multi-channel images, it is necessary to
```
kostrykin: > with coordinates XYC

I don't think that a user will understand this. I'm not sure myself what you mean by this :) I think, making the reference to single-channel images should be enough.

Actually, what other implicit constraints do we have, that we should mention? Single-frame images, as in opposite of multi-frame image sequences? 2-D images?

tools/cellpose/cellpose.xml
```diff
@@ -72,77 +58,70 @@
...
+            <output name="cp_mask" file="img04_XYC_nuclei_masks.tif" compare="image_diff"/>
+            <output name="cp_segm" file="img04_XYC_nuclei_segm.png" compare="image_diff"/>
         </test>
     </tests>
     <help>
         <![CDATA[
         Cellpose: A generalist algorithm for cell and nucleus segmentation.
+        **Important**
+        This tool accepts single-channel images with coordinates XYC for segmentation. In case of multi-channel images, it is necessary to 
+        get the channel to process using the tool "Split image along axes" tool.
```
kostrykin: I think, this should work?

```suggestion
        get the channel to process using the `Split image along axes`_ tool.
        
        .. _Split image along axes: ?tool_id=toolshed.g2.bx.psu.edu%2Frepos%2Fimgteam%2Fsplit_image%2Fip_split_image
```

This creates a hyperlink to the tool, but the link only works if that tool is installed on the server.

tools/cellpose/cellpose.xml
kostrykin: We should add a validator to enforce single-channel images:

```suggestion
        <param name="img_in" type="data" format="ome.tiff,tiff,jpg,png" label="Choose the image file for segmention" help="Single-channel image for segmentation">
            <validator type="dataset_metadata_in_range" metadata_name="channels" min="0" max="1"/>
        </param>
```

However, this doesn't seem to work as expected. I tried it with `planemo serve`, but it still also accepts multichannel images (Galaxy correctly recognizes the number of channels of the images in the metadata, I checked that interactively). Here are some example data: [images.zip](https://github.com/user-attachments/files/23860020/images.zip) @bgruening Can you maybe have a look, what we are doing wrong with the validator here?

tools/cellpose/cellpose.xml
kostrykin: I agree that it is rather unusual, but there are some precedential cases where this is done, for example:
https://github.com/bgruening/galaxytools/blob/6915fa8ff00c5dc999f348ff34530243b2cdf545/tools/cellprofiler_v4/cellprofiler.xml#L2

Especially given that we are going to have `Cellpose SAM` with https://github.com/bgruening/galaxytools/pull/1695, I think it actually makes sense to be more specific here, to avoid confusion (Cellpose SAM also is Cellpose, right?). Hence, I'd go with:

```suggestion
    <description>with Cellpose 3</description>
```

tools/cellpose/cellpose.xml
rmassei: Ok, maybe I was a bit cryptic :D and it also not correct.

The tool will work for sure with 2D images, I actually did not test with multi-frame single channel images... Should we add an additional test to cover this?

tools/cellpose/cellpose.xml
rmassei: image

Actually it is working, when the validator is there it is not possible to run the tool @kostrykin 

I will also add a more informative the error message

tools/cellpose/cellpose.xml
kostrykin: IMIO, in a perfect world, we would either add a test, or forbid using what we have not tested (i.e. either clearly state what is supported or what is not supported, to avoid negative experiences, also add a validator).

rmassei: I added the validator and clearly stated in the Help Section which input is accepted

tools/cellpose/test-data/img01_XY_cyto2_masks.tif
kostrykin, line 1: Can you please rename TIF &rarr; TIFF? At least the new TIFF files.
https://github.com/FAIR-imaging/galaxy-image-community/issues/43

tools/cellpose/cellpose.xml
```diff
@@ -2,13 +2,7 @@
...
     <macros>
         <token name="@TOOL_VERSION@">3.1.0</token>
-        <token name="@VERSION_SUFFIX@">1</token>
-        <xml name="channel">
-            <option value="0" selected="true">grayscale/None</option>
-            <option value="1">red</option>
-            <option value="2">green</option>
-            <option value="3">blue</option>
-        </xml>
+        <token name="@VERSION_SUFFIX@">3</token>
```
kostrykin: I think you have incremented twice, haven't you?

```suggestion
        <token name="@VERSION_SUFFIX@">2</token>
```

tools/cellpose/cellpose.xml
```diff
@@ -27,28 +21,22 @@
...
             --output_dir ./segmentation
         ]]>
     </command>
     <configfiles>
         <inputs name="inputs" />
     </configfiles>
     <inputs>
-        <param name="img_in" type="data" format="ome.tiff,tiff,jpg,png" label="Choose the image file for segmention (usually after registration)"/>
+        <param name="img_in" type="data" format="ome.tiff,tiff,jpg,png" label="Choose the image file for segmentation (usually after registration)">
+            <validator type="dataset_metadata_in_range" metadata_name="channels" min="0" max="1" message="Input image is not single channel"/>
```
kostrykin: I suppose that this tool doesn't work with time-series images?
```suggestion
            <validator type="dataset_metadata_in_range" metadata_name="channels" min="0" max="1" message="Input image is not single channel."/>
            <validator type="dataset_metadata_in_range" metadata_name="frames" min="0" max="1" message="Input image is a multi-frame image sequence, must be a single-frame image."/>
```

tools/cellpose/cellpose.xml
```diff
@@ -72,77 +60,72 @@
...
+            <param name="model_type" value="nuclei"/>
+            <output name="cp_mask" file="img04_XYC_nuclei_masks.tif" compare="image_diff"/>
+            <output name="cp_segm" file="img04_XYC_nuclei_segm.png" compare="image_diff"/>
         </test>
     </tests>
     <help>
         <![CDATA[
         Cellpose: A generalist algorithm for cell and nucleus segmentation.
+        **Important:**
+        Only single channel 3D images with shapes XYC or CXY are supported (channel axis must be either first or last) by this tool. Other layouts (e.g. CXYZ, XYCZ etc.) are not accepted.
```
kostrykin: > Only single channel 3D images with shapes XYC or CXY are supported (channel axis must be either first or last)

This is kind of contradictory, isn't it? In the most simple case, a single-channel image has no C axis (e.g., YX, XY). Of course, there also are images with singleton axes like XYC or CXY. Is a singleton channel axis really mandatory? Otherwise, I'd phrase it like this:

```suggestion
        Only single channel 3-D images with XY, XYC, and CXY axes layout are supported (channel axis must be either first or last) by this tool. Other layouts (e.g. CXYZ, XYCZ etc.) are not accepted.
```

Also, what about 2-D images? I'd assume they are supported, but the text suggests that they are not?

```suggestion
        Only single channel 2-D and 3-D images are supported. If a channel axis is present, it must be a singleton axis and it must be either the first or the last axis within the axes layout (e.g., XY, XYC, or CXY). Other layouts are not accepted (e.g. CXYZ, XYCZ).
```

For 3-D images, the z-axis can be at arbitrary position in the axes layout, as long as C is first or last?

Actually, are you sure that layouts with YX are not supported?

kostrykin (commented):
Thanks a lot @rmassei, I know this is very tedious!

tools/cellpose/cellpose.xml
rmassei: Hi @kostrykin, sorry for the big confusion about definition and axis...!

This is definitely the most precise and concise explanation:

> Only single channel 2-D and 3-D images are supported. If a channel axis is present, it must be a singleton axis and it must be either the first or the last axis within the axes layout (e.g., XY, XYC, or CXY). Other layouts are not accepted (e.g. CXYZ, XYCZ).

kostrykin: Thanks @rmassei! I think this is the last issue that we need to figure out here…

I know I wrote that suggestion, but I'm still a bit confused. Maybe we can figure it out.

> If a channel axis is present, it must be a singleton axis and it must be either the first or the last axis within the axes layout (e.g., XY, XYC, or CXY). Other layouts are not accepted (e.g. CXYZ, XYCZ).

The `CXYZ` axes layout agrees with the requirement that _"if a channel axis is present, it must be a singleton axis and it must be […] the first […] axis within the axes layout"_. So why is the `CXYZ` layout not allowed?

rmassei: I am thinking of dimensions as axes, maybe this is the misunderstanding? 
A CXYY is a 4-D image, so it goes against the first statement 

> Only single channel 2-D and 3-D images are supported.

kostrykin: Ah yes! When people say 2-D/3-D/4-D data, they usually mean *spatial* or *spatiotemporal* axes (of course, technically, you're right, that YXC also is 3-D). So lets clear that up:

```suggestion
Only single-channel 2-D images are supported. If a channel axis is present, it must be a singleton axis and it must be either the first or the last axis within the axes layout (e.g., XY, XYC, or CXY). Other layouts are not accepted (e.g. CXYZ, XYCZ).
```

BTW, as far as I can see, all your test files that allow custom axes layouts (i.e. TIFF input files, that are `img03_CXY_nuclei.tiff` and `img04_XYC_nuclei.tif`) have YX axes order, not XY:
- `img03_CXY_nuclei.tiff` has CYX axes layout[^1]
- `img04_XYC_nuclei.tif` has YXC axes layout[^1]

I'm not sure whether `skimage.io.imread` can cope with `XY` axes order for TIFF files. Unless you have tested this, it might be better to be specific on this in the help text:

```suggestion
Only single-channel 2-D images are supported. If a channel axis is present, it must be a singleton axis and it must be either the first or the last axis within the axes layout (e.g., YX, YXC, or CYX). Other layouts are not accepted (e.g. CYXZ, XYCZ).
```

[^1]: In fact it's even something different, it's QYX and YXQ, respectively, but this is a different story. You can use `python -c "import tifffile; print(tifffile.TiffFile('img03_CXY_nuclei.tiff').series[0].axes)"` to check the axes layout of a TIFF file.

rmassei: ok, `img03_CXY_nuclei.tiff `and `img04_XYC_nuclei.tif` have now the axes layout XYC and CXY. Since the test went trough also in this case, the tool is working with CYX, YXC, XYC and CXY

tools/cellpose/cellpose.xml
```diff
@@ -72,77 +61,72 @@
...
+            <output name="cp_mask" file="img04_XYC_nuclei_masks.tiff" compare="image_diff"/>
+            <output name="cp_segm" file="img04_XYC_nuclei_segm.png" compare="image_diff"/>
         </test>
     </tests>
     <help>
         <![CDATA[
         Cellpose: A generalist algorithm for cell and nucleus segmentation.
+        **Important:**
+        Only single channel 2-D and 3-D images are supported. If a channel axis is present, it must be a singleton axis and it must be either the first or the last axis within the axes layout (e.g., XY, XYC, or CXY). Other layouts are not accepted (e.g. CXYZ, XYCZ).
+        In case of multichannel images, it is necessary to get the channel to process using the `Split image along axes`_ tool.
```
kostrykin: Lets fix the 3-D thing in the help and the consistency of writing `single channel` vs `multichannel`:
```suggestion
        Only single-channel 2-D images are supported. If a channel axis is present, it must be a singleton axis (axis of length 1) and it must be either the first or the last axis within the axes layout (e.g., YX, YXC, or CYX). Other layouts are not accepted (e.g. CYXZ, XYCZ).
        In case of multi-channel images, it is necessary to priorly extract the channel for segmentation using the `Split image along axes`_ tool.
```

Or if you prefer without the hyphens:
```suggestion
        Only singlechannel 2-D images are supported. If a channel axis is present, it must be a singleton axis (axis of length 1) and it must be either the first or the last axis within the axes layout (e.g., YX, YXC, or CYX). Other layouts are not accepted (e.g. CYXZ, XYCZ).
        In case of multichannel images, it is necessary to priorly extract the channel for segmentation using the `Split image along axes`_ tool.
```

kostrykin (approved):
Thanks!

If possible, could you please rename the remaining `.tif` files in this PR to `.tiff`?

tools/cellpose/cellpose.xml
kostrykin: @rmassei What you think about this?

rmassei: I guess it make sense to specify the version. `Cellpose 3` for this and `Cellpose SAM` for the upcoming one. These therms are also used in the [documentation](https://cellpose.readthedocs.io/en/v3.1.1.1/).

PR #1694, opened 2025-10-22
Coreutils sort - Add end column and character range functionality

RZ9082:


tools/text_processing/text_processing/test-data/sorted3.tabular
```diff
@@ -0,0 +1,4 @@
+Bob	28.01.2024
+Alex	25.06.2025
+Charlie	13.06.2025
```
bgruening, line 3: Those two are sorted wrong, isn't it?

RZ9082: No, I am testing for years and months in ascending order, and days in descending order

RZ9082:
What about the other linting errors? should the version suffix be bumped for all tools in the package?

PR #1695, opened 2025-10-23
add cellposeSAM (v4)

sunyi000:
this is related to #1692 
ping @rmassei 

- The v4.0.7 container was already merged into biocontainers, but has not become available in quay.io yet.
- you will need to uncomment the `<requirement>` once it becomes available
- I tested it locally, with the same container, it seems to work.although took a bit long
- it currently runs cpsam model
- feel free to continue working on it, removing/adding/modifying things.

I will be away for the next 2 weeks, and work on it again after Im back.

thanks.


tools/cellpose_v4/.shed.yml
```diff
@@ -0,0 +1,14 @@
+name: cellposesam
```
bgruening, line 1: is cellposeSAM = cellpose4

sunyi000: yes. it's the same

btw, i think the test will take a bit of time.. even test locally without GPU, it takes long time to finish.

bgruening:
That is an interesting one.

The error comes from Pillow: https://github.com/python-pillow/Pillow/blob/ec40c546d7d38efebcb228745291bd3ba6233196/src/PIL/Image.py#L3559

bgruening:
Look also at this command: https://github.com/python-pillow/Pillow/blob/ec40c546d7d38efebcb228745291bd3ba6233196/Tests/test_file_libtiff.py#L1149

It seems the produced files do have some problems.

sunyi000:
> Look also at this command: https://github.com/python-pillow/Pillow/blob/ec40c546d7d38efebcb228745291bd3ba6233196/Tests/test_file_libtiff.py#L1149
> 
> It seems the produced files do have some problems.

Thanks a lot @bgruening 

Does this mean that the options we have now are
either lower the Pillow version in the current cellpose v4 container, or we wait until Pillow fixes this in the next version?

sunyi000:
> > Look also at this command: https://github.com/python-pillow/Pillow/blob/ec40c546d7d38efebcb228745291bd3ba6233196/Tests/test_file_libtiff.py#L1149
> > It seems the produced files do have some problems.
> 
> This is about Pillow v6. Galaxy is [at v11 oder v12 now](https://github.com/search?q=repo%3Agalaxyproject%2Fgalaxy%20pillow&type=code), but I don't what version is installed in the Conda environment. I strongly guess that it is something >>6, can you please verify that, @sunyi000?
> 
> Can you maybe also share the output images you obtain, so I can check whether there is something suspicious?

the pillow version in the environment is 11.0.0
I used the input test image here, called `image.png`

image

cellpose version=4.0.7

the command I use to run 
` cellpose --image_path ./image.png --no_npy  --save_png  --save_tif --flow_threshold 0.4 --cellprob_threshold 0.0 --niter 0 --anisotropy 1.0 `

the command creates 2 images, 1 png mask 1 tif mask here

image_cp_masks

[image_cp_masks.tif](https://github.com/user-attachments/files/23881158/image_cp_masks.tif)

sunyi000:
@kostrykin I fixed the issue. 
The tool is actually working now. 

However
The tests will pass only if we set `eps` to a large number. Because every time cellpose runs, it creates masks, the objects in the masks maybe the same but the labels on the masks are not always the same. 

So I guess the options are
1.set a large `eps` value OR, 
2. use `sim_size` instead of `image_diff`
3. write a custom python script in the tool to compare masks' stats, not pixels, e.g. extract the total no. object and put it in json file, then in the tests, compare this json file. But, this is just an idea and may require a lot more work

@bgruening what would you suggest?

Thank you

kostrykin:
> @kostrykin I fixed the issue. The tool is actually working now.

Nice. What was the issue?

> However The tests will pass only if we set `eps` to a large number. Because every time cellpose runs, it creates masks, the objects in the masks maybe the same but the labels on the masks are not always the same.

This is what we have `metric="iou"` for. https://docs.galaxyproject.org/en/latest/dev/schema.html#id118

Is it all deterministic/reproducible now, aside from the labels?

tools/cellpose_v4/cellpose.xml
```diff
@@ -0,0 +1,165 @@
+<tool id="cellposesam" name="Run superhuman generalization for cellular segmentation" version="@TOOL_VERSION@+galaxy@VERSION_SUFFIX@" profile="24.2">
```
kostrykin: I know that this is the paper title… but, frankly speaking, this is quite a sensational wording, that I'd suggest to avoid in the name of the tool. I'd stick to the same name as the v3 tool, just change the description (like you did).

tools/cellpose_v4/cellpose.xml
```diff
@@ -0,0 +1,165 @@
...
+    <description>with Cellpose SAM</description>
+    <macros>
+        <token name="@TOOL_VERSION@">4.0.7</token>
+        <token name="@VERSION_SUFFIX@">1</token>
```
kostrykin:
```suggestion
        <token name="@VERSION_SUFFIX@">0</token>
```

tools/cellpose_v4/cellpose.xml
```diff
@@ -0,0 +1,165 @@
+<tool id="cellposesam" name="Run superhuman generalization for cellular segmentation" version="@TOOL_VERSION@+galaxy@VERSION_SUFFIX@" profile="24.2">
+    <description>with Cellpose SAM</description>
```
kostrykin: I think the precise wording is
```suggestion
    <description>with Cellpose-SAM</description>
```
according to the paper.

tools/cellpose_v4/test-data/image_cp_masks.tiff
kostrykin, line 1: Can you please rename the files `tif` &rarr; `tiff`? https://github.com/FAIR-imaging/galaxy-image-community/issues/43

tools/cellpose_v4/.shed.yml
```diff
@@ -0,0 +1,14 @@
...
+owner: bgruening
+categories:
+  - Imaging
+description: CellPose SAM
+remote_repository_url: https://github.com/bgruening/galaxytools/tree/master/tools
+type: unrestricted
+homepage_url: https://cellpose.org
+long_description: |
+    Cellpose-SAM: cell and nucleus segmentation with superhuman generalization. It can be optimized for your own data, applied in 3D, works on images with shot noise, (an)isotropic blur, undersampling, contrast inversions, regardless of channel order and object sizes.
```
kostrykin, line 10: > It can be optimized for your own data, applied in 3D

Can you please add a test for a 3D image?

tools/cellpose_v4/cellpose.xml
```diff
@@ -0,0 +1,165 @@
...
+	#end if
+        > ./cp.log 2>&1
+
+        ]]>
+    </command>
+    <configfiles>
+        <inputs name="inputs" />
+    </configfiles>
+    <inputs>
+        <param name="img_in" type="data" format="ome.tiff,tiff,jpg,png" label="Choose the image file for segmention (usually after registration)"/>
```
kostrykin: According to the `long_description`, this seems not to support multi-frame images, right? What about images with only 2 or more than 3 channels? Can you please add corresponding validators to prevent negative user experiences?

```suggestion
        <param name="img_in" type="data" format="ome.tiff,tiff,jpg,png" label="Choose the image file for segmention (usually after registration)">
            <validator type="dataset_metadata_in_range" metadata_name="frames" min="0" max="1" message="Input image is multi-frame, must be single-frame"/>
        </param>
```

A corresponding validator for the `channels` would also be nice.

tools/cellpose_v4/cellpose.xml
```diff
@@ -0,0 +1,165 @@
...
+        </section>
+        <param name="detailed_output" type="boolean" label="Detailed logging file?" help="If set, a CellPose log file will be generated." />
+    </inputs>
+    <outputs>
+        <data format="tiff" name="cp_tiff" from_work_dir="output/image_cp_masks.tif" label="Cellpose SAM masks in TIFF">
+	    <filter>save_tif</filter>
+	</data>
+	<data format="png" name="cp_png" from_work_dir="output/image_cp_masks.png" label="Cellpose SAM masks in PNG">
+	    <filter>save_png</filter>
+	</data>
```
kostrykin: Using RGB images for labels is somewhat unconventional. Most tools (I'm aware of) require single-channel images for labels. In terms of tool interoperability, it would be advisable to change this output to a uint16 oder uint32 single-channel TIFF.

Because this is what this output is for, right, for subsequent analysis using other tools?

sunyi000: These are just e everything from the cellpose cli, I have no idea how to change these .

sunyi000: not sure if i understand correctly. these are just arguments for the cellpose cli, we just wrap them in the xml.
it has a `--save_png` arg, and others like `--save_tiff, --save_flows` etc. 
do we really need to care about what format cellpose generate?

I know the flows, outlines are for visualization, but since cellpose create the files, we need to test these output files in the xml. I'm not sure what are the best metric to test these files.

The `image_diff` use `mae,mse, iou,rms, fro`, currently the test will pass if we just set a big number to eps, but is this really the correct way to test, that's what I'm trying to find out..
or, we simply use `sim_size` for these two files instead..or we just don't wrap these two arguments in cellpose..

tools/cellpose_v4/cellpose.xml
kostrykin:
```suggestion
        <param name="img_in" type="data" format="ome.tiff,tiff,jpg,png" label="Choose the image file for segmentation (usually after registration)"/>
```

tools/cellpose_v4/cellpose.xml
kostrykin: > do we really need to care about what format cellpose generate?

Depends. If we care about interoperability of the tools that we have in Galaxy, then we should care about this. This is up to you. If you care about it, then you might want to create a Python wrapper, that does some conversion. But I understand if you feel like this is a whole new barrel. In that case, I'd appreciate if you could create an issue on [our community repo](https://github.com/FAIR-imaging/galaxy-image-community/issues) after finishing this PR, so we will remember to fix this at some later point.

> The image_diff use mae,mse, iou,rms, fro, currently the test will pass if we just set a big number to eps, but is this really the correct way to test, that's what I'm trying to find out..

No, there shouldn't be any difference between the runs. I tested it locally ([link](https://github.com/bgruening/galaxytools/pull/1695#issuecomment-3636458708)) and I cannot spot any differences. Where are the deviations coming from?

sunyi000: thanks, did you test the two runs within planemo, or just the cellpose runs?
If I just run the same commands in cellpose env(the same used by planemo, but without planemo), it generates the same output, however, if I run twice the planemo test command, it reports failure in the test..

tools/cellpose_v4/cellpose.xml
```diff
@@ -0,0 +1,165 @@
...
+	    <param name="no_resample" type="boolean" truevalue="--no_resample" falsevalue="" checked="false" label="Disables flows/cellprob resampling to original image size before computing masks."
+                help="Using this flag will make more masks more jagged with larger diameter settings."/>
+            <param argument="flow_threshold" type="float" min="0" value="0.4" label="Flow error threshold (all cells with errors below threshold are kept) (not used for 3D)"/>
+            <param argument="cellprob_threshold" type="float" value="0.0" label="Cell probability threshold (all pixels with prob above threshold kept for masks)"/>
+            <param argument="niter" type="integer" min="0" value="0" label="Number of iterations"
+		help="By defalut, sets the number of iterations to be proportional to the ROI diameter. For longer ROIs, more iterations might be needed."/>
+            <param argument="anisotropy" type="float" min="0" value="1.0" label="Anisotropy of volume in 3D"
+                help="By defalut, sets the number of iterations to be proportional to the ROI diameter. For longer ROIs, more iterations might be needed."/>
```
kostrykin: FIx typo, …
```suggestion
                help="By default, sets the number of iterations to be proportional to the ROI diameter. For longer ROIs, more iterations might be needed."/>
```

… plus, it's not clear:
- What is the ROI that is referred to here? Or you mean the image?
- What a "long" ROI is? Or you mean "larger"?
- What does "by default" mean? The value of 1.0?
- How can the "number of iterations" be fractional? Or, is this the factor of the proportionality?

If so, maybe be more clear on this:

```suggestion
                help="Governs the number of iterations. For larger images, more iterations might be needed."/>
```

tools/cellpose_v4/cellpose.xml
kostrykin: > thanks, did you test the two runs within planemo, or just the cellpose runs?

just cellpose runs

> If I just run the same commands in cellpose env(the same used by planemo, but without planemo), it generates the same output, however, if I run twice the planemo test command, it reports failure in the test..

then it must be something with the wrapper.

sunyi000:
> > I ran this twice and I cannot confirm any deviations, at least for the output of `cp_flows` that is `image_flows_cp_masks.tif` the outputs of the two runs are identical, even on a binary level.
> > > a question: is it correct to say if iou is 0.5, it means perfect match of two object?
> > 
> > 
> > No. The `iou` metric is defined as `1 - intersection(X, Y) / union(X, Y)`, and if the two objects `X` and `Y` are identical, then `iou` will be 0. If you experience weird results with the `iou` metric, then this might be due to the fact that you are using it for RGB images, which is kind of something that it wasn't meant for (but still should work, I think).
> 
> thanks a lot for the explanation, which is really helpful

@kostrykin sorry, on the `iou` again..
1. from [here](https://veradp-ai.com/iou-vs-dice/), `iou` seems to mean Jaccard index, at least in imaging?

2. if 1 is true, then in the galaxy implementation [here](https://github.com/usegalaxy-eu/galaxy/blob/37a7ea0d748b0fad38958ec7467c0083577c23b6/lib/galaxy/tool_util/verify/__init__.py#L477) is wrong.
Looks like the galaxy implemented the half-dice index, which is not a `iou`.

If i understand correctly, under the galaxy implementation, `iou=0.5` is perfect match

3. The `iou` is a similarity measure, so the greater the better, `1` is the perfect match. However, in Galaxy documentation [here](https://docs.galaxyproject.org/en/latest/dev/schema.html#tool-tests-test-output), (the `eps` explanation in the middle of the page), says it is the max distance allowed. 

This is confusing, Distance should be the inverse of similarity.  if we set `metric=iou`, then the  `eps` should be set bigger, the closer to 1 the better.

I think maybe we should ask in the upstream Galaxy..

tools/cellpose_v4/cellpose.xml
sunyi000: instead of the suggested change, should we just directly copy over the cellpose explanation of the params? this maybe more user friendly as they see the same `help` in Galaxy tool and Cellpose Documentation.

sunyi000:
it's quicker with cellpose3, just slow with v4

I;m going to leave this here for now, and come back to it later.

kostrykin:
Thanks.

Which output is that?

In case it's different on your local branch then on the remote: What is the exact `<output...>` definition that you are using for verification?

sunyi000:
@kostrykin this is green now..can we merge?
thanks

kostrykin (approved):
https://github.com/bgruening/galaxytools/actions/runs/20269527079/job/58201757613?pr=1695#step:9:1186

> All 4 test(s) executed passed.
> cellposesam (Test #1): passed
> cellposesam (Test #2): passed
> cellposesam (Test #3): passed
> cellposesam (Test #4): passed

Yes, it seems to be working now… although I'm wondering what is going on with the [`Combine chunked test results` check](https://github.com/bgruening/galaxytools/actions/runs/20269527079/job/58206402539)

**Two last requests from my side:**

*(i)* Can you please fix the indentation of the XML file? It's super random and really hard to follow for people who are going to maintain this in the future.

*(ii)* You have used quite large `eps` for your tests, which indicates some amount of non-determinism, which is not to be expected here. This might also be gone with the Biocontainer now. It would be awesome if you could try lowering the `eps` values to the defaults (or at least not higher than 0.1), if possible.

kostrykin (approved):
Thanks!

tools/cellpose_v4/cellpose.xml
```diff
@@ -0,0 +1,200 @@
...
+	<param name="channel_axis" type="integer" optional="true" min="0" label="Channel axis" help="The axis (0-based) of the image which corresponds to the image channels and to the z axis" />
+	<param name="use_gpu" type="boolean" checked="false" truevalue="true" falsevalue="false" label="Whether to use GPU?" />
+	<section name="options" title="Advanced Options" expanded="False">
+        <param name="no_norm" type="boolean" label="Do not normalize images" truevalue="true" falsevalue="false" checked="false"/>
+        <param name="diameter" type="float" optional="true" label="Cell or nuclei diameter in pixels" help="Use to resize cells to the training diameter. Leave blank for automated estimation."/>
+	    <param name="no_resample" type="boolean" truevalue="true" falsevalue="false" checked="false" label="Disables flows/cellprob resampling to original image size before computing masks."
+		    help="Using this flag will make more masks more jagged with larger diameter settings."/>
+        <param name="no_interp" type="boolean" truevalue="true" falsevalue="false" checked="true" label="Interpolate when running dynamics."/>
```
kostrykin, line 100: Two questions here: @sunyi000 

**a)** From `--help`, the `--no_interp` sounds like it _prevents_ interpolation when running dynamics:
> --no_interp        do not interpolate when running dynamics (was default)

Thus, the label should actually be _"**Do not** interpolate when running dynamics."_, no?
```suggestion
        <param name="no_interp" type="boolean" truevalue="true" falsevalue="false" checked="true" label="Do not interpolate when running dynamics."/>
```

**b)** I'm not sure what this switch _actually_ does, but since it is disabled by default, what was the intention to enable it by default in the Galaxy wrapper? Or is this a "mistake" and it should rather be…?

```suggestion
        <param name="no_interp" type="boolean" truevalue="true" falsevalue="false" checked="false" label="Interpolate when running dynamics."/>
```

sunyi000: i think you are right, maybe chaning a) is enough?

kostrykin: My preference would be to also change b) because, unless there is some reason to do otherwise (is there, in this case?), one should stick to the defaults of the wrapped tool to avoid confusion by users (e.g., when they are used to the command line tool already).

PR #1706, opened 2025-10-30
Yolo training : Add the ability to choose your own model

TuturBaba:
Updated yolo_training.xml to support input from a pre-trained model and added new test files. 
The first two tests don’t work on my machine, maybe they are outdated?

### DIFF
--- a/tools/image_processing/yolo/yolo_training.xml
+++ b/tools/image_processing/yolo/yolo_training.xml
@@ -99,7 +99,7 @@
 	    </section>
     </inputs>
     <outputs>
-        <data name="best_model" auto_format="true" from_work_dir="output_files/best.pt" label="Best Model"/>
+        <data name="best_model" auto_format="true" from_work_dir="output_files/best.pt" label="Best model"/>
         <data name="last_model" auto_format="true" from_work_dir="output_files/last.pt" label="Last Model"/>
         <data name="metrics_csv" format="csv" from_work_dir="output_files/results_metrics.csv" label="Training Metrics"/>
         <data name="metrics_plot" format="png" from_work_dir="output_files/results_plot.png" label="Training Plot"/>


tools/image_processing/yolo/yolo_training.xml
```diff
@@ -61,16 +65,30 @@
...
+                <option value="https://github.com/ultralytics/assets/releases/download/v8.3.0/yolo11n-seg.pt">YOLO11n-seg</option>
+                <option value="https://github.com/ultralytics/assets/releases/download/v8.3.0/yolo11n.pt">YOLO11n</option>
+                <option value="https://github.com/ultralytics/assets/releases/download/v8.3.0/yolov8n.pt">YOLOv8</option>
+                <option value="https://github.com/ultralytics/assets/releases/download/v8.3.0/sam_b.pt">SAM base</option>
+                <option value="https://github.com/ultralytics/assets/releases/download/v8.3.0/sam_l.pt">SAM large</option>
+                <option value="https://github.com/ultralytics/assets/releases/download/v8.3.0/sam2_t.pt">SAM2 tiny</option>
+                <option value="https://github.com/ultralytics/assets/releases/download/v8.3.0/sam2.1_t.pt">SAM2.1 tiny</option>
+            </param>
+            <when value="personal">
+                    <param name="model_input" type="data" format="data" label="Select your pretrained YOLO model from history" />
```
bgruening: Do you know if we can have those models in the "safetensors" format?

TuturBaba: No, normally it only supports YOLO or SAM, and your question triggered me — so I checked the code and tested it myself. SAM doesn’t even work, and when I looked at the code it made sense, because it only has `model = YOLO()` and never `model=SAM()`, but since it uses ultralytics I don’t think safetensor works but maybe some accept

TuturBaba: SAM is a zero-shot model, not intended for training, but for instant segmentation and data auto-annotation. I think I will look into it later to see if it wouldn't be interesting to make a tool out of it.

tools/image_processing/yolo/yolo_training.xml
```diff
@@ -61,20 +67,26 @@
...
-        <param name="model_url" type="select" optional="false">
-            <option value="https://github.com/ultralytics/assets/releases/download/v8.3.0/yolo11n-seg.pt">YOLO11n-seg</option>
-            <option value="https://github.com/ultralytics/assets/releases/download/v8.3.0/yolo11n.pt">YOLO11n</option>
-            <option value="https://github.com/ultralytics/assets/releases/download/v8.3.0/yolov8n.pt">YOLOv8</option>
-            <option value="https://github.com/ultralytics/assets/releases/download/v8.3.0/sam_b.pt">SAM base</option>
-            <option value="https://github.com/ultralytics/assets/releases/download/v8.3.0/sam_l.pt">SAM large</option>
-            <option value="https://github.com/ultralytics/assets/releases/download/v8.3.0/sam2_t.pt">SAM2 tiny</option>
-            <option value="https://github.com/ultralytics/assets/releases/download/v8.3.0/sam2.1_t.pt">SAM2.1 tiny</option>
-        </param>
+	    <param name="input_yolo" type="data" format="txt" multiple="true"  label="Input YOLO txt files" help="The YOLO text files, each text file must correspond to one input image (same name different extension)." />
```
bgruening:
```suggestion
	    <param name="input_yolo" type="data" format="txt" multiple="true"  label="Input YOLO txt files" help="The YOLO text files, each text file must correspond to one input image (same name, different extension)." />
```

tools/image_processing/yolo/yolo_training.xml
```diff
@@ -95,67 +107,103 @@
...
+                    <param name="num_train" value="70" />
+                    <param name="epochs" value="10" />
+                    <param name="image_size" value="500" />
+                    <param name="scale" value="0.8" />
+                    <param name="degree" value="10.0" />
+                    <param name="hsv_v" value="0.5"/>
+                    <param name="init_lr" value="0.02" />
+                    <param name="weight_decay" value="0.001" />
+            </section>
+                <output name="best_model">
```
bgruening, line 124: indentation is here wrong.

The Galaxy language server has a reformat option that you can use.

tools/image_processing/yolo/yolo_training.xml
```diff
@@ -95,67 +107,103 @@
...
+                        <has_n_lines n="11"/>
+                    </assert_contents>
             </output>
-            <output name="metrics_csv">
-                <assert_contents>
-                    <has_n_lines n="11"/>
-                </assert_contents>
-	    </output>
-	    <output name="metrics_plot" file="results_plot.png" compare="image_diff" />
+            <output name="metrics_plot" file="results_plot.png" compare="sim_size" />
```
bgruening: image_diff is a bit better than sim_size, any reason why you have changed that?

tools/image_processing/yolo/yolo_training.xml
```diff
@@ -95,67 +107,103 @@
...
+                <output name="metrics_csv">
+                    <assert_contents>
+                        <has_n_lines n="11"/>
+                    </assert_contents>
             </output>
-            <output name="last_model">
-                <assert_contents>
-                    <has_size value="5960100" delta="100"/>
-                </assert_contents>
+            <output name="metrics_plot" file="results_plot1.png" compare="sim_size" />
```
bgruening: image_diff

bgruening:
@kostrykin @sunyi000 can you look at this one. Looks good from my side.

tools/image_processing/yolo/yolo_training.xml
```diff
@@ -95,67 +106,103 @@
...
+				<param name="image_size" value="500" />
+				<param name="scale" value="0.8" />
+				<param name="degree" value="10.0" />
+				<param name="hsv_v" value="0.5"/>
+				<param name="init_lr" value="0.02" />
+				<param name="weight_decay" value="0.001" />
+            </section>
+                <output name="best_model">
+                    <assert_contents>
+                        <has_size value="5960100" delta="100"/>
```
kostrykin, line 166: A bit of off topic, but would it actually make sense to include the option to run custom commands for output validation in a future release of Galaxy? Like, run some Python script to validate the produced output?

tools/image_processing/yolo/yolo_training.xml
```diff
@@ -95,67 +106,103 @@
...
+            <section name="training_params">
+				<param name="num_train" value="70" />
+				<param name="epochs" value="10" />
+				<param name="image_size" value="640" />
+				<param name="scale" value="0.5" />
+				<param name="hsv_v" value="0.5"/>
+				<param name="init_lr" value="0.02" />
+				<param name="weight_decay" value="0.001" />
+            </section>
+                <output name="best_model">
```
kostrykin, line 204: Indentation seems wrong here and below

kostrykin (commented):
I strongly second that `image_diff` should be used for validation of output images

tools/image_processing/yolo/yolo_training.xml
TuturBaba: "Yes, that's because the metrics are never exactly the same from one run to another. The sizes remain very close to each other, at least if I remember correctly

bgruening: you can add "eps" to increaes the difference I think

bgruening:
> Output metrics_plot:  different than expected, difference (using image_diff):
( /tmp/tmpxxaliaynresults_plot1.png v. /tmp/tmpsvjut5mhresults_plot1.png )
Image difference 0.03798912037037037 exceeds eps=0.01.

sunyi000:
I did some changes, and local tests passed. 
i opened a PR in your forked repo @TuturBaba can you merge and try again?
Thanks

kostrykin:
> It's not on my computer, it's on the test GitHub But with eps=1 , test passed. It's not ideal, but it works

I get the motivation to make the CI pass, and sometimes relaxing tolerances is the right short-term fix. That said, one of the reasons we keep "strict" defaults is to raise early attention to possible reproducibility issues. Of course there are cases, when tools are non-deterministic, but I think these are very rare exceptions.

If tests fail on CI but not locally, it would be good to understand whether this is _(i)_ genuinely non-deterministic behavior in the underlying tool (rather unlikely, but possible), or _(ii)_ a problem with reproducibility in the wrapper or the test setup.

Sometimes, even in case _(i)_ it is possible to establish reproducibility by tweaking environment variables and alike (like we did for the Cellpose wrapper, for example).

tools/image_processing/yolo/yolo_training.xml
```diff
@@ -169,7 +175,13 @@
                         <has_n_lines n="11"/>
                     </assert_contents>
 	            </output>
-	            <output name="metrics_plot" file="results_plot1.png" compare="image_diff" eps="0.02" />
+	            <output name="metrics_plot" >
+                    <assert_contents>
+                        <has_size min="10000"/>
+                        <has_image_channels channels="4"/>
+                        <has_image_height height="1200"/>
```
kostrykin, line 183: Can we also employ `has_image_width` here and below?

tools/image_processing/yolo/yolo_training.xml
```diff
@@ -61,31 +67,36 @@
     </command>
     <inputs>
         <param name="input_images" type="data" format="jpg" multiple="true"  label="Input images"/>
```
kostrykin, line 69: Any specific reason why only JPG is allowed, and not PNG or TIFF, for example?

(JPG is rather discouraged for computed-based image analysis due to compression artifacts)

tools/image_processing/yolo/yolo_training.xml
```diff
@@ -61,31 +67,36 @@
...
+	        <param name="hsv_h" type="float" value="0.015" min="0.0" max="1.0" label="Image HSV-Hue" help="Image HSV-Hue augmentation (fraction)." />
+		    <param name="init_lr" type="float" value="0.01" min="0.00001" max="0.1" label="Learning rate" help="Initial learning rate for training." />
+	        <param name="weight_decay" type="float" value="0.0005" min="0.0001" max="0.1" label="Weight decay" />
+	 	    <param name="confidence" type="float" value="0.5" min="0.0" max="1.0" label="Confidence" help="Confidence value (0-1) for each detected bounding box." />
+		    <param name="iou" type="float" value="0.7" label="IoU" min="0.1" max="1.0" help="Intersection over Union threshold for non-maximum suppression." />
+		    <param name="max_det" type="integer" value="300" min="100" max="1000" label="Max. number of detections" help="Maximum number of detections allowed per image. Limits the total number of objects the model can detect in a single inference, preventing excessive outputs in dense scenes." />
+	    </section>
     </inputs>
     <outputs>
         <data name="best_model" auto_format="true" from_work_dir="output_files/best.pt" label="Best Model"/>
```
kostrykin: Consistent capitalization:
```suggestion
        <data name="best_model" auto_format="true" from_work_dir="output_files/best.pt" label="Best model"/>
```

bgruening:
Thanks @TuturBaba

PR #1713, opened 2025-11-04
bia-ftp make accession parameter mandatory in XML

pavanvidem:
I skipped bumping the gx version

Fix https://github.com/FAIR-imaging/galaxy-image-community/issues/11

### DIFF
--- a/tools/image_processing/bia-ftplinks/biaftplink.xml
+++ b/tools/image_processing/bia-ftplinks/biaftplink.xml
@@ -22,7 +22,7 @@
 ]]>
     </command>
     <inputs>
-        <param name="accession" type="text" label="The accession ID of BioImages-Core or BioStudies-JCB" help="for eg. S-BIAD570, S-JCBD-201309038"/>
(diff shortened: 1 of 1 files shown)


tools/image_processing/bia-ftplinks/biaftplink.xml
```diff
@@ -22,7 +22,7 @@
 ]]>
     </command>
     <inputs>
-        <param name="accession" type="text" label="The accession ID of BioImages-Core or BioStudies-JCB" help="for eg. S-BIAD570, S-JCBD-201309038"/>
+        <param name="accession" type="text" optional="false" label="The accession ID of BioImages-Core or BioStudies-JCB" help="for eg. S-BIAD570, S-JCBD-201309038"/>
```
kostrykin, line 25: You sure that this fixes it? Because, actually [`optional` should be `false` by default](https://docs.galaxyproject.org/en/latest/dev/schema.html#id45).

My suspicion was that an empty string is not considered as "no input", and I assumed that we actually need to use an [`empty_field` validator](https://docs.galaxyproject.org/en/latest/dev/schema.html#validators-for-textual-inputs-text-select).

pavanvidem: it should fix it

PR #1716, opened 2025-11-06
Add BioStudies upload tool

davelopez:
This pull request adds a new Galaxy tool `biostudies_upload`, which enables users to upload data to the BioStudies database at EMBL-EBI using Aspera high-speed transfer. 

The tool uses the new `credentials` requirement available in Galaxy since version 25.1 to handle secrets in a user-friendly way.

Tests are expected to fail since there is no current way of providing credentials in the testing environment for now.

### DIFF
--- a/tools/biostudies_upload/biostudies_upload.xml
+++ b/tools/biostudies_upload/biostudies_upload.xml
@@ -11,8 +11,18 @@
         </credentials>
     </requirements>
     <command><![CDATA[
+## Determine what to upload based on the dataset format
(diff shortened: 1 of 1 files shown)


hechth:
@davelopez check this for adding credentials to tests :)
https://github.com/RECETOX/galaxytools/blob/master/tools/aoptk/aoptk_chemical_identifier.xml

Adding an empty environment variable or default value - of course this will not work if you need real credentials for testing which you can't / won't expose

PR #1717, opened 2025-11-07
Update CI

arash77:
Upgrade the GitHub Actions workflow from tools_iuc to use the latest versions of actions and update the branch reference for consistency.

### DIFF
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -15,7 +15,7 @@ on:
       - '*'
 env:
   GALAXY_FORK: galaxyproject
-  GALAXY_BRANCH: release_25.0
+  GALAXY_BRANCH: release_25.1
   MAX_CHUNKS: 4
   MAX_FILE_SIZE: 1M
 concurrency:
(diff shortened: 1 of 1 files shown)


.github/workflows/pr.yaml
```diff
@@ -403,7 +403,7 @@ jobs:
   deploy-report:
     name: Report deploy status
     needs: [deploy]
-    if: ${{ always() && needs.deploy.result != 'success' && (github.ref == 'refs/heads/master' || github.ref == 'refs/heads/main' ) && github.repository_owner == 'bgruening' }}
+    if: ${{ always() && (github.ref == 'refs/heads/master' || github.ref == 'refs/heads/main' ) && github.repository_owner == 'bgruening' }}
```
bgruening, line 406: @arash77 do you have any idea if we can put this "bgruening" into a VAR or anything, so that we can actually just update all YAML without any local changes?

arash77: So, do we want to add, for example, `vars.REPOSITORY_OWNER` instead of the owner username?
I think using [environment](https://docs.github.com/en/actions/how-tos/write-workflows/choose-what-workflows-do/deploy-to-environment) option would be better.

bgruening: Yes, for example.

arash77: https://github.com/bgruening/galaxytools/pull/1720

PR #1720, opened 2025-11-13
Update GitHub Actions deployment conditions

arash77:
Refine [deployment conditions](https://docs.github.com/en/actions/how-tos/write-workflows/choose-what-workflows-do/deploy-to-environment) in the GitHub Actions workflow by removing the repository owner check and adding an [environment](https://docs.github.com/en/actions/how-tos/deploy/configure-and-manage-deployments/manage-environments) specification for deployment.

### DIFF
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -366,8 +366,9 @@ jobs:
   deploy:
     name: Deploy
     needs: [setup, lint, combine_outputs]
-    if: ${{ (github.ref == 'refs/heads/master' || github.ref == 'refs/heads/main' ) && github.repository_owner == 'bgruening' }}
+    if: ${{ github.ref == 'refs/heads/master' || github.ref == 'refs/heads/main' }}
     runs-on: ubuntu-latest
+    environment: toolshed-deployment
     strategy:
       matrix:
         python-version: ['3.11']
@@ -403,8 +404,9 @@ jobs:
   deploy-report:
     name: Report deploy status
     needs: [deploy]
-    if: ${{ always() && (github.ref == 'refs/heads/master' || github.ref == 'refs/heads/main' ) && github.repository_owner == 'bgruening' }}
+    if: ${{ always() && (github.ref == 'refs/heads/master' || github.ref == 'refs/heads/main' ) }}
     runs-on: ubuntu-latest
+    environment: toolshed-deployment
     steps:
     # report to the PR if deployment failed
     - name: Get PR object


.github/workflows/pr.yaml
```diff
@@ -366,8 +366,9 @@ jobs:
   deploy:
     name: Deploy
     needs: [setup, lint, combine_outputs]
-    if: ${{ (github.ref == 'refs/heads/master' || github.ref == 'refs/heads/main' ) && github.repository_owner == 'bgruening' }}
+    if: ${{ github.ref == 'refs/heads/master' || github.ref == 'refs/heads/main' }}
     runs-on: ubuntu-latest
+    environment: toolshed-deployment
```
bernt-matthias, line 371: How is this working?

arash77: This is the documentation of how to use it:
https://docs.github.com/en/actions/how-tos/deploy/configure-and-manage-deployments/manage-environments

bernt-matthias: I think the intention of `github.repository_owner == 'bgruening'` was to avoid accidental deployment (which probably does not work anyway -- since secrets will not be configured -- I guess). Would this be covered?

arash77: Yes, it was, but we wanted to avoid this (`github.repository_owner == 'bgruening'`).
But if it’s not even needed, we can just completely remove them from tools_iuc and here to make it easier to update this file regularly.

bgruening: If you 1) have a fork, 2) in your github account globally TS credentials and 3) you commit to master/main ... then an upload will be triggered, isn't it? Or is this too wild, and we should not care about this case?

bernt-matthias: Maybe we follow "better safe than sorry" here as well :)

bgruening: Can we force that the credentials coming from a repo-secret and not a global secret?

arash77: Using an environment enforces this. (but not blocking to use other secrets)
Adding "Environment secrets" instead of repo/org secrets is the safest option in my opinion.

bgruening: Thanks! That makes then perfect sense.

@bernt-matthias if you agree we can test this here, but ideally would need to update IUC as well. Do you have any suggestion where do document this?

bernt-matthias: Code comment? Or in the template repo readme.

arash77: Is `CONTRIBUTING.md` file suitable for that?

bernt-matthias: I would add it here https://github.com/galaxyproject/planemo-ci-action#lint-mode, i.e. that the two variables influence the linting step as well.

arash77: Is this for environments or https://github.com/galaxyproject/planemo-ci-action/pull/65 one!?

PR #1723, opened 2025-11-25
Update flexynesis

nilchia:
Updates Flexynesis to version 1.1.6

Key changes:
* In GNN, users can now use their own network, instead of STRING
* Added the function to create the network from an expression matrix
* updated arguments for Cox function
* Restric visualization functions to models that generate embeddings (Xref: https://github.com/BIMSBbioinfo/flexynesis/issues/124)


tools/flexynesis/flexynesis_plot.xml
```diff
@@ -141,12 +141,9 @@
...
-                #if str($plot_conditional.crossval_conditional.crossval) == 'yes':
-                    --crossval
-                #end if
-                #if str($plot_conditional.crossval_conditional.crossval) == 'yes':
-                    --n_splits $plot_conditional.crossval_conditional.n_splits
-                #end if
+                --n_splits $plot_conditional.n_splits
+                --low_variance_threshold $plot_conditional.low_variance_threshold
+                --cox_penalizer $plot_conditional.cox_penalizer
                 --random_state 42
```
pavanvidem, line 147: why do you need an argument to the python script if it is always fixed?

tools/flexynesis/flexynesis_plot.py
```diff
@@ -1065,128 +1344,266 @@ def main():
...
+        help="Number of top important features to include in Cox model. Default is 5",
+    )
+    parser.add_argument(
+        "--n_splits",
+        type=int,
+        default=5,
+        help="Number of folds for cross-validation. Default is 5",
+    )
+    parser.add_argument(
+        "--random_state",
```
pavanvidem, line 1454: This is not changed from any Galaxy tools params. Can be removed.

tools/flexynesis/flexynesis_plot.xml
nilchia: So the user also knows from command that what param was used. They don't need to go and check the script.

tools/flexynesis/flexynesis.xml
```diff
@@ -489,20 +419,20 @@
...
         </conditional>
-        <param name="safetensors_out_bool" type="boolean" truevalue="True" falsevalue="False" checked="true" label="Save the trained model?" help="The model will be saved as Safetensors format." />
     </inputs>
     <outputs>
         <collection name="results" type="list" label="${tool.name} on ${on_string}: results">
             <discover_datasets pattern="(?P&lt;name&gt;.+)\.tabular$" format="tabular" directory="output"/>
         </collection>
         <data name="model" format="safetensors" from_work_dir="output/job.final_model.safetensors" label="${tool.name} on ${on_string}: trained_model">
-            <filter>safetensors_out_bool == True</filter>
+            <filter>training_type['model_class']['safetensors_out_bool'] == True</filter>
```
pavanvidem: `safetensors_out_bool` was never used.  `--safetensors` flag is always on which is creating an extra model file.

tools/flexynesis/flexynesis_plot.py
nilchia: What do you mean? I didn't understand.

tools/flexynesis/flexynesis.xml
nilchia: Thanks a lot!!!!

nilchia:
>>/usr/local/lib/python3.13/site-packages/geomloss/samples_loss.py:58: SyntaxWarning: invalid escape sequence '\i'
>
>Can we also deactivate please the progress-bar?

That is coming from this line here, I assume: https://github.com/BIMSBbioinfo/flexynesis/blob/f79f25d11d3254f2a399d75e358c733aa889b8f3/flexynesis/utils.py#L1752

I can't find an argument to deactivate the progress bar there. :thinking: 
@bgruening

pavanvidem:
here an example: https://github.com/bgruening/galaxytools/blob/master/tools/whisper/whisper.xml#L79

PR #1725, opened 2025-12-04
Restrict number of GPUs

anuprulez:
- `--bwa-normalized-queue-capacity 1` and `--num-gpus 1` may restrict the number of GPUs used on the EU
- `--gpuwrite` and `--gpusort` may improve performance on GPUs

https://docs.nvidia.com/clara/parabricks/latest/documentation/tooldocs/man_fq2bam.html#useful-options-for-performance

@bgruening

### DIFF
--- a/tools/parabricks/parabricks_fq2bam.xml
+++ b/tools/parabricks/parabricks_fq2bam.xml
@@ -2,7 +2,7 @@
     <description>Run GPU-bwa mem mapper</description>
     <macros>
         <token name="@TOOL_VERSION@">4.6.0-2</token>
-        <token name="@VERSION_SUFFIX@">1</token>
+        <token name="@VERSION_SUFFIX@">2</token>
         <token name="@PROFILE@">24.2</token>
     </macros>
     <creator>


tools/parabricks/parabricks_fq2bam.xml
```diff
@@ -15,7 +15,7 @@
     <command detect_errors="aggressive">
     <![CDATA[
         #if $reference_source.reference_source_selector == 'cached':
-            #set $pbrun_command = 'pbrun fq2bam --low-memory --preserve-file-symlinks --ref ' + $reference_source.ref_file.fields.path
+            #set $pbrun_command = 'pbrun fq2bam --low-memory --preserve-file-symlinks --bwa-normalized-queue-capacity 1 --num-gpus 1 --gpuwrite --gpusort --ref ' + $reference_source.ref_file.fields.path
```
bgruening: num_gpus should probably be an ENV var, similar to GALAXY_SLOTS ... should bwa-normalized-queue also be configurable?

PR #1732, opened 2025-12-09
Added m4a as input and output options for ffmpeg

dadrasarmin:
Some tests from this tool that was previously wrapped by others are already failing and I do not think that is related to this change.

I added m4a input, output, and test as requested by @Sch-Da here: https://github.com/galaxyproject/tools-iuc/issues/7401


tools/ffmpeg/ffmpeg_converter.xml
```diff
@@ -10,64 +10,77 @@
...
+                <when input="output_format" value="m4a" format="m4a"/>
             </change_format>
         </data>
     </outputs>
     <tests>
         <test expect_num_outputs="1">
             <param name="input_file" value="SampleVideo_360x240_cut.mp4" ftype="mp4"/>
             <param name="output_format" value="webm"/>
-            <output name="output_file" file="SampleVideo_360x240_cut.webm" ftype="webm"/>
+            <output name="output_file" file="SampleVideo_360x240_cut.webm" ftype="webm" compare="sim_size" delta="20000"/>
```
bgruening: can you please convert that to asserts, specify the size and the delta ... I think this makes it more reviewable. I don't know how big the file is and if such a large delta is justifyable ....

dadrasarmin: Done.

PR #1734, opened 2025-12-15
Add kMetaShot and DM

SantaMcCloud:
Test has to be done via VM since around 16 GiB RAM was needed. Here are the test result from the VM;

**Test 1:**
image

**Test 2:**
image

**Test 3:**
image


data_managers/data_manager_kmetashot/.shed.yml
```diff
@@ -0,0 +1,10 @@
+categories:
+- Data Managers
+- Metagenomics
+homepage_url: https://github.com/gdefazio/kMetaShot
+description: Data manager for kMetaShot reference data
+long_description: Data manager for kMetaShot reference data
+name: kmetashot_build_database
+owner: iuc
```
bgruening:
```suggestion
owner: bgruening
```

bgruening (approved):

PR #1735, opened 2025-12-18
add Psortb

pavanvidem:
Moved from tools-iuc (https://github.com/galaxyproject/tools-iuc/pull/7537) to here

### DIFF
--- a/tools/psortb/.shed.yml
+++ b/tools/psortb/.shed.yml
@@ -7,7 +7,7 @@ long_description: |
   Gram-positive, Gram-negative, and Archaea models.
 
 homepage_url: https://psort.org/
-remote_repository_url: https://github.com/bgruening/galaxytools/tree/main/tools/psortb
+remote_repository_url: https://github.com/bgruening/galaxytools/tree/master/tools/psortb
 type: unrestricted
 categories:
   - Proteomics


tools/psortb/psortb.xml
```diff
@@ -0,0 +1,115 @@
...
+				<option value="--positive" selected="true">Gram-positive (Bacteria)</option>
+				<option value="--negative">Gram-negative (Bacteria)</option>
+				<option value="--archaea">Archaea</option>
+			</param>
+		</section>
+
+		<param name="output_format" type="select" label="Output format" help="Choose PSORTb output format.">
+			<option value="normal" selected="true">Normal (human-readable)</option>
+			<option value="terse">3-column (terse)</option>
+			<option value="long">30-column (long)</option>
```
bgruening:
```suggestion
			<option value="normal">Normal (human-readable)</option>
			<option value="terse">3-column (terse)</option>
			<option value="long" selected="true">30-column (long)</option>
```

bgruening: I think I would like to have something machine readable as default, something that can be reused

tools/psortb/.shed.yml
```diff
@@ -0,0 +1,15 @@
+name: psortb
+owner: iuc
```
bgruening:
```suggestion
owner: bgruening
```

bgruening (approved):
Just two tiny suggestions.

PR #1740, opened 2026-01-02
vRhyme: Allow bam/sam inputs to accept multiple datasets without mapping over the tool

neo417:
The vRhyme readme at https://github.com/AnantharamanLab/vRhyme?tab=readme-ov-file#other-inputs highly suggests to use multiple samples for coverage calculations.

I could not find a way to do this right now without using the paired read collection and re-running the expensive bowtie2 alignment. From my experimentation I found that merging the BAM files does not address the problem, as it leads to different results to passing them separately.

PS: When using BAM files produced by bowtie2 is is critical that the input nucleotide scaffolds given to both tools do not contain spaces anywhere in the sequence headers. Otherwise vRhyme will not be able to match up the identifiers when extracting the coverage information from the BAM files.

### DIFF
--- a/tools/vrhyme/vrhyme.xml
+++ b/tools/vrhyme/vrhyme.xml
@@ -704,7 +704,7 @@ $new_name#slurp
             <output name="genes" ftype="fasta">
                 <assert_contents>
                     <has_n_lines n="1206" delta="5"/>
-                    <has_size value="85520" delta="1000"/>
+                    <has_size value="88520" delta="1000"/>
                     <has_text text="virus_1__13792:29888_1"/>
                 </assert_contents>
             </output>


tools/vrhyme/vrhyme.xml
```diff
@@ -6,6 +6,13 @@
     <expand macro="biotools"/>
     <expand macro="requirements"/>
     <command detect_errors="exit_code"><![CDATA[
+    #def makeNameUnique($name, $ext, $tracker={})
```
bgruening, line 9: @neo417 I'm not sure we should do this.

All this implicit conversion might be unexpected to the user and screw up more complex workflow that assume a consistent element_id. There are collection tools in Galaxy that can rename-collection identifiers and I think I would prefer this more explicit behaviour.

neo417: I thought it might be useful, as it strictly only modifies the name of the BAM file in cases where the tool would otherwise crash because of duplicated identifiers in a collection.
Additionally the identifiers only appear in the output in one place (the coverage file) where vRhyme already modifies them (prepend std_/avg_ and suffix _1 in case of paired read collections).

I will change it if you want, but then I am unsure how to write the test for multiple bam files without copying the test file to give it a different name (480kb).

bgruening: We can leave it. All good.

neo417:
Thank you!
If you don't mind, what is the process like for getting the update to [usegalaxy.eu](https://usegalaxy.eu/)?
If it is propagated automatically, how long does that usually take?

PR #1743, opened 2026-01-07
Handle RateLimitError in LLM Hub tool

arash77:
Enhance error handling by adding support for RateLimitError during API request retries of llm hub.

### DIFF
--- a/tools/llm_hub/llm_hub.py
+++ b/tools/llm_hub/llm_hub.py
@@ -5,7 +5,7 @@ import sys
 import time
 
 import yaml
-from openai import InternalServerError, OpenAI
+from openai import InternalServerError, OpenAI, RateLimitError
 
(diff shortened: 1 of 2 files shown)


bgruening (approved):

arash77:
Still needed! beyond the original `RateLimitError` backoff, this now also retries on `APITimeoutError` and `APIConnectionError`. We keep hitting timeout errors with this tool, especially on collections and large datasets (long-running generation, e.g. `glm-5.2` translation jobs), which previously escaped the retry loop and crashed the tool with an unhandled traceback. The client timeout is now configurable via `LITELLM_REQUEST_TIMEOUT` (default 600 s) and the SDK's internal retries are disabled so the loop owns backoff. Tool version is bumped to `2.41.1+galaxy1`.

bgruening (approved):

PR #1748, opened 2026-01-12
Update 3dtrees_tile_merge v1.0.2

kgerb:


tools/3dtrees_tile_merge/til_mer.xml
```diff
@@ -54,7 +53,7 @@
         </conditional>
     </inputs>
     <outputs>
-        <data name="output_tile" format="zip" label="Prepared Files" from_work_dir="prepared_files.zip">
+        <data name="output_tile" format="zip" label="prepared_files" from_work_dir="prepared_files.zip">
```
bgruening: Same here is not is not very nice for humans.

kgerb: We derived the naming from the labels and the spelling was hard to work with in the database - changed on our end :)

PR #1749, opened 2026-01-12
Update: 3dtrees_SegmentAnyTree v1.1.0+1

kgerb:
Fixed naming in xml file - nothing too special


tools/3dtrees_segmentanytree/segmentanytree.xml
```diff
@@ -41,12 +41,12 @@
...
     </inputs>
     <outputs>
-        <data name="output" format="zip" label="Processed Files">
+        <data name="output" format="zip" label="processed_files">
             <change_format>
                 <when input="input.ext" value="laz" format="laz"/>
             </change_format>
         </data>
-        <data name="resource_usage" format="txt" label="Resource Usage">
+        <data name="resource_usage" format="txt" label="resource_usage">
```
bgruening, line 49: But this is less human readable. Why would you want that?

PR #1766, opened 2026-02-03
New tool addition: search_proteins tool

SaimMomin12:
The PR is still in a preliminary state and is waiting on upstream.

https://github.com/RolandFaure/search_protein

### DIFF
--- a/tools/search_proteins/search_database.xml
+++ b/tools/search_proteins/search_database.xml
@@ -24,7 +24,8 @@
         <param name="query_embeddings_names" type="data" format="txt" label="Query embedding names" help="Output from the Embed Query tool" />
         <param name="database" type="select" label="FAISS Database" help="Select FAISS database">
             <options from_data_table="faiss_protein_databases">
-                <filter type="regexp" column="version" value="^1\.0$"/>
+                <filter type="sort_by" column="3"/>
+                <filter type="static_value" column="version" value="@TOOL_VERSION@"/>
             </options>
         </param>
         <param name="outfmt" type="select" optional="true" label="MMseqs2 output format" help="Format for MMseqs2 alignment output">
--- a/tools/search_proteins/test-data/faiss_database.loc
+++ b/tools/search_proteins/test-data/faiss_database.loc
@@ -3,4 +3,4 @@
 # - db-name
 # - version
 # - /path/to/data 
-faiss-demo-db-20260203	FAISS Test Database	1.0	${__HERE__}/test-db/
+faiss-demo-db-20260203	FAISS Test Database	1.2.0	${__HERE__}/test-db/


tools/search_proteins/macros.xml
```diff
@@ -0,0 +1,29 @@
+<macros>
+    <token name="@TOOL_VERSION@">1.0.0</token>
```
SaimMomin12:
```suggestion
    <token name="@TOOL_VERSION@">1.2.0</token>
```

Adhere to container version

tools/search_proteins/embed_query.xml
```diff
@@ -0,0 +1,65 @@
...
+            --query_sequences '$query_sequences'
+            --output results
+            #if $force_cpu:
+                --force_cpu
+            #end if
+            -F
+    ]]></command>
+    <inputs>
+        <param name="query_sequences" type="data" format="fasta" label="Query sequences (FASTA)" help="Input FASTA file with protein sequences to embed" />
+        <param name="force_cpu" type="boolean" truevalue="true" falsevalue="false" checked="true" label="Force CPU usage" help="Force CPU usage even if GPU is available. GPU acceleration takes ~1s/query vs 120s/query on CPU" />
```
bgruening:
```suggestion
```

tools/search_proteins/embed_query.xml
```diff
@@ -0,0 +1,65 @@
...
+    </macros>
+    <expand macro="requirements"/>
+    <command detect_errors="exit_code"><![CDATA[
+        mkdir -p results &&
+        python /app/embed_query.py
+            --query_sequences '$query_sequences'
+            --output results
+            #if $force_cpu:
+                --force_cpu
+            #end if
```
bgruening:
```suggestion
```

remove that, but include a check for an ENV variable, if that ENV is set, it will use CPUs ...

Maybe $GALAXY_LOGAN_PROTEIN_SEARCH_FORCE_CPU ...?

tools/search_proteins/.shed.yml
```diff
@@ -0,0 +1,31 @@
...
+homepage_url: https://github.com/RolandFaure/search_protein
+long_description: |
+    search_protein is a two-step pipeline for fast, embedding-based protein
+    sequence similarity searching.  Query sequences are first converted into
+    dense vector representations using the GLM2 language model, then searched
+    against a pre-built FAISS index of a reference proteome.  Candidate hits
+    are validated and aligned with MMseqs2.  The approach scales sub-linearly
+    with the number of queries, making it practical for large-scale metagenomic
+    and comparative-proteomics workflows.
+remote_repository_url: https://github.com/bgruening/galaxytools/tree/master/tools/search_protein
```
SaimMomin12:
```suggestion
remote_repository_url: https://github.com/bgruening/galaxytools/tree/master/tools/search_proteins
```

tools/search_proteins/embed_query.xml
bgruening: This will not work. `\${}` is a bash variable only available during job-runtime. But the if-condition runs during job-creation time.

Use a bash-if check for an existing ENV vairable.

tools/search_proteins/tool_data_table_conf.xml.sample
```diff
@@ -0,0 +1,6 @@
+<tables>
+    <table name="faiss_protein_databases" comment_char="#">
+        <columns>value, name, dbkey, path</columns>
```
bgruening: can you add a version here as well, and filter in the tool for version = 1 .... so we are future proof.

tools/search_proteins/embed_query.xml
SaimMomin12: For example following?

```
    if [ "${GALAXY_LOGAN_PROTEIN_SEARCH_FORCE_CPU:-True}" = "True" ]; then
        FORCE_CPU="--force_cpu"
    else
        FORCE_CPU=""
    fi

    python /app/embed_query.py \
        --query_sequences "$query_sequences" \
        --output results \
        $FORCE_CPU \
        -F
```

bgruening: I think this is better:

```
if [ -n "$MY_VAR" ]; then
    echo "Variable exists and is not empty."
else
    echo "Variable is either unset or empty."
fi

File: tools/search_proteins/search_database.xml [hunk 6]
```diff
@@ -24,7 +24,7 @@
         <param name="query_embeddings_names" type="data" format="txt" label="Query embedding names" help="Output from the Embed Query tool" />
         <param name="database" type="select" label="FAISS Database" help="Select FAISS database">
             <options from_data_table="faiss_protein_databases">
-                <filter type="sort_by" column="3"/>
```
bgruening: we can still sort, please add that back

tools/search_proteins/search_database.xml
```diff
@@ -0,0 +1,109 @@
...
+                --outfmt '$outfmt'
+            #end if
+    ]]></command>
+    <inputs>
+        <param name="query_sequences" type="data" format="fasta" label="Original query sequences (FASTA)" help="Same FASTA file used in the embedding step, needed for MMseqs2 alignment" />
+        <param name="query_embeddings_npy" type="data" format="npy" label="Query embeddings (npy)" help="Output from the Embed Query tool" />
+        <param name="query_embeddings_names" type="data" format="txt" label="Query embedding names" help="Output from the Embed Query tool" />
+        <param name="database" type="select" label="FAISS Database" help="Select FAISS database">
+            <options from_data_table="faiss_protein_databases">
+                <filter type="regexp" column="version" value="^1\.0$"/>
```
bgruening: something like `                <filter type="static_value" value="@TOOL_VERSION@" column="db_version"/>` should work, please also add back the sort_by

PR #1776, opened 2026-02-14
update whisperx to 3.8.1

bgruening:


bgruening:
@arash77 any idea why the test fails?

Or better how could the test pass in your first version :)

bgruening:
I don't know, that was my questino. The CI def. run ... but it turned green. Can it be that huggingface downloaded the model on the fly?

arash77:
I'm trying to test it, but it requires a GPU! I'm also not sure if the GPU is enabled on GitHub Actions.

PR #1778, opened 2026-02-16
uniprot: allow to select additional output fields

bernt-matthias:
when mapping to uniprot DBs


bgruening:
Nice, that will be very useful! Tests are relevant.

PR #1634, opened 2025-06-06
add YOLO related tools for image segmentation

sunyi000:
@bgruening just adding some yolo tools for building a workflow in Galaxy.
could you have a look? thank you very much

PR #1790, opened 2026-02-25
Yolo class

TuturBaba:
This pull request aims to add class support to the tool. In training, it is now possible to provide the list of classes. I removed the class input from prediction because it did not provide class names and was only used to know the number of classes. It has been replaced with the classes already known by the model.

### DIFF
--- a/tools/image_processing/yolo/yolo_training.xml
+++ b/tools/image_processing/yolo/yolo_training.xml
@@ -1,4 +1,4 @@
-<tool id="yolo_training" name="Perform YOLO training" version="@TOOL_VERSION@+galaxy6" profile="24.2">
+<tool id="yolo_training" name="Perform YOLO training" version="@TOOL_VERSION@+galaxy5" profile="24.2">
     <description>with ultralytics</description>
     <macros>
         <import>macros.xml</import>


tools/image_processing/yolo/yolo_prediction.xml
```diff
@@ -1,4 +1,4 @@
-<tool id="yolo_predict" name="Perform YOLO image labeling" version="@TOOL_VERSION@+galaxy@VERSION_SUFFIX@" profile="24.2">
+<tool id="yolo_predict" name="Perform YOLO image labeling" version="@TOOL_VERSION@+galaxy5" profile="24.2">
```
kostrykin, line 1: I think you should rather increment the value of the `@VERSION_SUFFIX@` macro?

tools/image_processing/yolo/yolo_training.xml
```diff
@@ -1,4 +1,4 @@
-<tool id="yolo_training" name="Perform YOLO training" version="@TOOL_VERSION@+galaxy5" profile="24.2">
+<tool id="yolo_training" name="Perform YOLO training" version="@TOOL_VERSION@+galaxy6" profile="24.2">
```
kostrykin:
```suggestion
<tool id="yolo_training" name="Perform YOLO training" version="@TOOL_VERSION@+galaxy@VERSION_SUFFIX@" profile="24.2">
```

tools/image_processing/yolo/yolo_training.xml
```diff
@@ -66,7 +67,7 @@
     ]]>
     </command>
     <inputs>
-        <param name="input_images" type="data" format="jpg" multiple="true"  label="Input images"/>
+        <param name="input_images" type="data" format="jpg,png,tiff" multiple="true"  label="Input images"/>
```
kostrykin: Can you please add (small) tests for PNG and TIFF?

See also: https://github.com/bgruening/galaxytools/pull/1634#discussion_r2144459995

If I remember correctly, TIFF didn't work for predictions, at least, so it's worth testing.

TuturBaba: I thought I had tested both PNG and TIFF, but apparently not. You are right — TIFF files do not work properly for predictions in the current state.

One possible solution would be to add a small preprocessing step to convert TIFF images to RGB before running inference. I already tried this locally and it seems to work quite well.

However, I am not sure whether this is the right approach. Since I am not very familiar with TIFF, I am questioning whether converting to RGB could decrease the interest of using it in the first place.

tools/image_processing/yolo/yolo_training.xml
TuturBaba: Personally, I don’t like using a macro for the version number. If you change the macro for YOLO training and prediction, then json2yolosegment.xml will also change automatically.

Adding a new version without any changes to the tool is not an idea I really like. But if necessary, I can make the changes.

kostrykin: > Adding a new version without any changes to the tool is not an idea I really like.

True. What's the stance of @bgruening here?

tools/image_processing/yolo/yolo_training.xml
kostrykin: > One possible solution would be to add a small preprocessing step to convert TIFF images to RGB before running inference. I already tried this locally and it seems to work quite well.

This sounds like a very valid approach.

You just need to be careful that a TIFF might already be RGB, or, in fact, have any number of channels (like 2 or 4 or 5 or more). So if you do not want to deal with this complexity (which I deem somewhat out of place here), I suggest to rather restrict the inputs to a valid set of file formats/representations. There are two questions here:

1. The input must be RGB? Add a validator that checks for 3 channels (or 4 if RGBA is also allowed).
2. The input must be PNG? Do not allow TIFF input, so it is in the responsibility of the user to convert (maybe they want to only use one channel from the image, or one slice from a 3-D stack, or one frame from a time-series).

For (1), I think you should be able to use something like:
```xml
<validator
    type="dataset_metadata_in_range"
    metadata_name="channels"
    min="3"
    max="4"
    message="Image must have 3 (RGB) or 4 (RGBA) channels."/>
```

Of course, you _can_ do the conversion (which probably is more user-friendly). But then you have to take care of such cases like time-series, 3-D stacks, and so on, when dealing with TIFFs.

> I am questioning whether converting to RGB could decrease the interest of using it in the first place.

Not sure what you mean?

tools/image_processing/yolo/yolo_training.xml
bgruening: If we only bump on tool in a suite we need to do it as @TuturBaba did.

But in general macros should be used, just to have one spot where the bot and humans can bump numbers

tools/image_processing/yolo/yolo_training.xml
TuturBaba: >>I am questioning whether converting to RGB could decrease the interest of using it in the first place.

>Not sure what you mean?

I’m not sure either :)

I tested both RGB and RGBA images. The tool works correctly with RGB images but does not support RGBA images. Therefore, I added a validator with min=3 and max=3 to enforce strictly 3-channel RGB images.


And I have a question about Galaxy EU, my last pull request is not there, even though it was pushed three weeks ago.

Do you know why, or is there something I should do?

kostrykin: Do you have a link for the Galaxy EU PR?

bgruening: In those cases please check 1) if your changes reached the ToolShed (if not deployment failed here on CI), 2) if your tool is listed in https://github.com/usegalaxy-eu/usegalaxy-eu-tools ... if this is also 3, please ping me then something went wrong during installation on EU.

TuturBaba: It is the YOLO training tool from this PR:
[https://github.com/bgruening/galaxytools/pull/1706#issue-3570596070](https://github.com/bgruening/galaxytools/pull/1706#issue-3570596070)

In the ToolShed, the version has not been updated.

TuturBaba: Normally Yvan found the error:  [https://github.com/bgruening/galaxytools/pull/1706#issuecomment-4004159903](https://github.com/bgruening/galaxytools/pull/1706#issuecomment-4004159903)

TuturBaba: I found a problem while testing my tool. To confirm that everything was working correctly, I tried to use the images from the tutorial.

I noticed that my validator does not accept them (black and white JPG images with 1 channel), but the YOLO function does.

So it seems that YOLO accepts black and white JPG images, but not black and white TIFF images.
Is it possible to run the validator only on TIFF images?

Error message:
 ```
RuntimeError: Given groups=1, weight of size [16, 3, 3, 3], expected input[3, 1, 640, 640] to have 3 channels, but got 1 channel instead.
```

Comment (bgruening) [reply tools/image_processing/yolo/yolo_training.xml]: That is a question for @kostrykin ... But I don't think so. You could use a conditional ask for the input type and offer a different type-data param with different validator.

Comment (kostrykin) [reply tools/image_processing/yolo/yolo_training.xml]: I'm not sure that this is possible, but you can try fiddling around with `<validator type="expression"><![CDATA[expr]]></validator>` where `expr` is a Pythonic expression and `value` is the selected dataset, I think. For example, using `value.metadata.channels == 3` for `expr` will give you a validator that only passes if the image has 3 channels. The question now is whether the datatype is available from `value`. I absolutely found no documentation of this, but you can try searching GitHub repos, or just experiment, maybe you are lucky with something like `value.datatype` or `value.format` or `value.ext` etc.

Comment (TuturBaba) [reply tools/image_processing/yolo/yolo_training.xml]: @kostrykin Normally It's work
```xml
<validator type="expression" message="PNG/JPG acceptés, TIFF uniquement si 3 canaux (RGB)">
   value is not None and (value.ext in ('png', 'jpg', 'jpeg') or (value.ext in ('tiff', 'tif') and getattr(value.metadata, 'channels', 0) == 3))
</validator>
```

kostrykin: > Normally It's work

You mean it's working? That would be super cool 🥳

tools/image_processing/yolo/yolo_training.xml
bgruening:
```suggestion
<tool id="yolo_training" name="Perform YOLO training" version="@TOOL_VERSION@+galaxy5" profile="24.2">
```

bgruening (approved):

PR #1791, opened 2026-02-26
Update remurna.xml

SaimMomin12:
Fixes: https://github.com/bgruening/galaxytools/pull/1789 

FOR CONTRIBUTOR:


tools/rna_tools/remurna/remurna.xml
bgruening, line 18: do we need this which here?

SaimMomin12: Yes, I tried without it and the tool fails :(

bgruening: Good so I had a reason for it :)

PR #1818, opened 2026-03-19
Rag tool

LucaBEMan:
FOR CONTRIBUTOR:

---

## Description

This PR adds a new Galaxy tool: **RAG Retriever**.

The tool performs document retrieval for Retrieval-Augmented Generation (RAG) workflows using LlamaIndex and HuggingFace sentence-transformers embeddings. It extracts the most relevant text chunks from input documents based on semantic similarity to a user query.

The output is a context file that can be used as input for downstream LLM tools.

---

## Example Workflow

This tool is intended to be used together with the **LLM Hub** tool.

Typical workflow:

1. Upload one or more documents (PDF, JSON, TXT, CSV, Markdown, etc.)
2. Upload an embedding model archive (`.tar.gz` / `.tgz`) containing a HuggingFace sentence-transformers model
3. Run **RAG Retriever** to extract relevant context chunks
4. Pass the generated `rag_context.txt` file to the **LLM Hub** tool
5. Ask a question using the retrieved context

Pipeline overview:

Documents + Embedding Model Archive + Question → RAG Retriever → Context → LLM Hub → Answer

**Example Galaxy workflow combining RAG Retriever and LLM Hub:**

Galaxy workflow

## Notes

- The tool does not generate answers itself, but prepares high-quality context for LLMs.
- Supports multiple input formats including PDF, JSON, TXT, CSV, and Markdown.
- Accepts a HuggingFace sentence-transformers embedding model as a `.tar.gz` / `.tgz` archive input.
- The archive-based model input preserves the original model directory structure required for loading custom embedding models.


tools/rag/.shed.yml
```diff
@@ -0,0 +1,18 @@
...
+  or summarization tasks.
+  The tool supports GPU acceleration when available (CUDA) and falls back to CPU otherwise.
+remote_repository_url: https://github.com/bgruening/galaxytools/tree/master/tools/rag
+homepage_url: https://github.com/bgruening/galaxytools/tree/master/tools/rag
+type:
+categories:
+  - Machine Learning
+  - Text Processing
+maintainers:
+  - Luca Riethmüller
```
anuprulez:
```suggestion
  - LucaBEMan
  - arash77
  - anuprulez
  
  
```

tools/rag/rag_retriever.xml
```diff
@@ -0,0 +1,186 @@
+<tool id="rag_retriever" name="RAG Retriever" version="1.0.0">
```
anuprulez:
```suggestion
<tool id="rag_retriever" name="RAG Retriever" version="1.0.0" profile="24.2">
```

tools/rag/download_embeddings.py
```diff
@@ -0,0 +1,32 @@
...
+    python download_embeddings.py
+"""
+
+import os
+import sys
+from pathlib import Path
+
+from sentence_transformers import SentenceTransformer
+
+models = [
```
anuprulez: Can we take these models as inputs to the tool? Otherwise we are stuck with these models and if users want, they cannot use their favorite embedding models. Now its really easy to accept these models. Entire hugging face in available inside Galaxy which can be directly imported from file uploader to any history and then to the tool. Maybe in future, a better embedding model becomes available.

Image

Image

LucaBEMan: I’m not sure if I fully understand your suggestion.
Do you mean that the tool should accept any Hugging Face embedding model as a dataset input (e.g., uploaded or provided within a Galaxy history)?
If so, is there already an existing Galaxy tool that can download a Hugging Face model and make it available as a dataset, or would this require adding a separate tool for that purpose?

tools/rag/tool-data/huggingface.loc.sample.loc.sample
bgruening, line 1: what was wrong with the huggingface loc file?

tools/rag/download_embeddings.py
LucaBEMan: It would be possible to manually download the .tar archive of a Hugging Face model and provide it as input to the tool. The main challenge is that the internal directory structure of these models can vary between repositories, and when downloading models through Galaxy, the files are provided as separate datasets without preserving the original folder structure. However, libraries like SentenceTransformer expect a specific directory layout (e.g., modules.json, tokenizer files, model weights, etc.) in order to correctly load the model, so it is not reliable to pass the individual files from Galaxy directly to the tool. A more robust approach would be to accept a single archived model (e.g., .tar or .tar.gz) as input, extract it within the tool, and then pass the extracted directory path to the embedding loader, ensuring that the original structure is preserved and allowing arbitrary Hugging Face models to be used.

tools/rag/tool-data/huggingface.loc.sample.loc.sample
LucaBEMan: There is nothing inherently wrong with the HuggingFace .loc file approach. However, it requires embedding models to be pre-downloaded and registered by a Galaxy administrator, which limits flexibility. With the new approach, users can provide models dynamically as archives, allowing them to choose any embedding model they want without requiring admin setup. That said, as mentioned above, there are currently some issues with this approach that still need to be resolved.

bgruening: Can we add both approaches? Please keep in mind that we also need to be careful about security here. I guess we are not restricting the user-models to safetensors? If not, we add a security risk here.

LucaBEMan: I have updated the environment to PyTorch v2.6.0 to mitigate the security risks of user-provided .bin models, as this version enforces mandatory safety checks for the pickle-based format ([CVE-2025-32434](https://nvd.nist.gov/vuln/detail/CVE-2025-32434)).

LucaBEMan: Regarding the question if we can have both approaches: Generally, this is possible, but they behave differently in a Galaxy workflow.For the .tar archive approach, the model is provided as a direct file input within the workflow, as you can see in the image in the PR description.
However, for the .loc file approach, the model must be pre-selected during the tool configuration (as shown in the other attached image). This is because Galaxy data tables cannot be integrated as dynamic file inputs in a workflow. So the choice depends on whether you want a portable workflow (with tar) or a fixed model set by the admin (with .loc).
Bildschirmfoto 2026-03-23 um 13 46 33

tools/rag/.shed.yml
```diff
@@ -0,0 +1,21 @@
...
+homepage_url: https://github.com/bgruening/galaxytools/tree/master/tools/rag
+type:
+categories:
+  - Machine Learning
+maintainers:
+  - LucaBEMan
+  - arash77
+  - anuprulez
+  
+
```
arash77:
```suggestion

```

tools/rag/rag_retriever.xml
```diff
@@ -0,0 +1,163 @@
...
+      area="true">
+      <validator type="empty_field" />
+    </param>
+
+    <param name="model_archive" type="data" format="tar.gz,tgz,tar,data"
+      label="Embedding Model Archive" />
+
+    <param name="top_k"
+      type="integer"
+      value="5"
```
arash77:
```suggestion
      value="5"
      min="1"
```

tools/rag/rag_retriever.xml
```diff
@@ -0,0 +1,163 @@
+<tool id="rag_retriever" name="RAG Retriever" version="1.0.0" profile="24.2">
+  <description>Retrieve relevant chunks from PDFs and output a context file (question + chunks) for
+    LLM Hub</description>
```
arash77:
```suggestion
<tool id="rag_retriever" name="RAG Retriever" version="@TOOL_VERSION@+galaxy@VERSION_SUFFIX@" profile="24.2">
  <description>Retrieve relevant chunks from documents and output a context file for LLM Hub</description>

  <macros>
    <token name="@TOOL_VERSION@">1.0.0</token>
    <token name="@VERSION_SUFFIX@">0</token>
  </macros>
```

tools/rag/rag_retriever.xml
```diff
@@ -0,0 +1,163 @@
...
+  </outputs>
+
+  <tests>
+    <test expect_failure="true" expect_exit_code="2">
+      <param name="files" value="test.txt" ftype="txt" />
+      <param name="question" value="Test" />
+      <param name="model_archive" value="test.txt" />
+      <param name="top_k" value="5" />
+      <assert_stderr>
+        <has_text text="gzip: stdin: unexpected end of file" />
```
arash77:
```suggestion
        <has_text_matching expression="gzip|not a gzip file|unexpected end of file" />
```

tools/rag/test-data/test.txt
```diff
@@ -0,0 +1 @@
+This is just a test file!
```
arash77, line 1:
```suggestion
This is just a test file!
```

tools/rag/rag_retriever.xml
```diff
@@ -0,0 +1,163 @@
...
+        <citation type="bibtex">
+            @misc{rag_retriever,
+                author = {RAG Retriever},
+                title = {RAG Retriever - Retrieval-Augmented Generation Document Retrieval},
+                howpublished = {\url{https://github.com/bgruening/galaxytools/tree/master/tools/rag}},
+                year = {2025}
+            }
+        </citation>
+    </citations>
+</tool>
```
arash77:
```suggestion
    <citation type="bibtex">
        @misc{rag_retriever,
            author = {RAG Retriever},
            title = {RAG Retriever - Retrieval-Augmented Generation Document Retrieval},
            howpublished = {\url{https://github.com/bgruening/galaxytools/tree/master/tools/rag}},
            year = {2025}
        }
    </citation>
</citations>
</tool>
```

tools/rag/rag_retriever.xml
```diff
@@ -0,0 +1,168 @@
+<tool id="rag_retriever" name="RAG Retriever" version="@TOOL_VERSION@+galaxy@VERSION_SUFFIX@" profile="24.2">
+  <description>Retrieve relevant chunks from documents and output a context file for LLM Hub</description>
+
+  <macros>
+    <token name="@TOOL_VERSION@">1.0.0</token>
+    <token name="@VERSION_SUFFIX@">0</token>
+  </macros>
+
+  <requirements>
+    <container type="docker">quay.io/luca_r/rag-tool:v1.0.3</container>
```
arash77: After https://github.com/usegalaxy-eu/docker-rag has a release you can use this:
```suggestion
    <container type="docker">quay.io/galaxy/rag-retriever:@TOOL_VERSION@</container>
```

arash77: I pinned the latest version of `llama-index` in the `Dockerfile` to `0.14.18` so we could use it in `quay.io/galaxy/rag-retriever:0.14.18`, but I think @bgruening needs to grant access first for it to be shown publicly.

LucaBEMan: Is the image already available?

tools/rag/tool-data/huggingface.loc.sample.loc.sample
bgruening: > However, for the .loc file approach, the model must be pre-selected during the tool configuration (as shown in the other attached image). 

This is not entirely true. You can click this <-> symbol and connect it with a workflow parameter. This workflow parameter can then be configured to expose all elements from the data table. The user starting the workflow can than choose between all available models in the loc file.

tools/rag/rag_retriever.xml
bgruening: Yes

tools/rag/tool-data/huggingface.loc.sample.loc.sample
LucaBEMan: Thanks, that is a good point. You are right that the tool parameter itself can be connected to a workflow parameter via the <-> symbol. However, what I am currently missing is how this is represented on the workflow level: in the workflow editor I do not see a dedicated workflow input type that directly handles a data table selection. I only see inputs such as text, float, dataset, etc. So while the parameter can be exposed, it is not clear to me how the workflow input would preserve the actual data table choice instead of behaving like a plain text-style parameter. Could you clarify how this is expected to work in practice for a .loc-backed select parameter?

bgruening: grafik

On the right-side "Restrict Text Values"

tools/rag/rag_retriever.xml
```diff
@@ -0,0 +1,202 @@
+<tool id="rag_retriever" name="RAG Retriever" version="@TOOL_VERSION@+galaxy@VERSION_SUFFIX@"
+  profile="24.2">
+  <description>Retrieve relevant chunks from documents and output a context file</description>
+
+  <macros>
+    <token name="@TOOL_VERSION@">1.0.0</token>
+    <token name="@VERSION_SUFFIX@">0</token>
+  </macros>
+
```
bgruening, line 10: please include required_files https://docs.galaxyproject.org/en/latest/dev/schema.html#tool-required-files-include

and also add creators: https://docs.galaxyproject.org/en/latest/dev/schema.html#tool-creator-person

tools/rag/rag_retriever.xml
```diff
@@ -0,0 +1,202 @@
...
+  <requirements>
+    <container type="docker">quay.io/luca_r/rag-tool:v1.0.5</container>
```
bgruening: change this container please to @arash77 version

tools/rag/rag_retriever.xml
```diff
@@ -0,0 +1,202 @@
...
+      <param name="source_type" type="select" label="Embedding Model Source">
+        <option value="preinstalled">Use an installed model</option>
+        <option value="upload">Upload a custom model archive</option>
+      </param>
+
+      <when value="preinstalled">
+        <param name="embedding_model" type="select" label="Select Model">
+          <options from_data_table="huggingface">
+            <column name="name" index="1" />
+            <column name="value" index="6" />
```
bgruening, line 60: @arash77 we should filter here a bit more? Or do we allow all models?

tools/rag/rag_retriever.xml
```diff
@@ -0,0 +1,202 @@
...
+
+    <param name="question"
+      type="text"
+      optional="false"
+      label="Question"
+      area="true">
+      <validator type="empty_field" />
+    </param>
+
+    <param name="top_k" type="integer" value="5" min="1" label="Top K chunks" />
```
bgruening: a bit more help would be useful

tools/rag/rag_retriever.xml
arash77: Yes, I think we can filter by free_tag to search specifically for embedding models.

tools/rag/rag_retriever.xml
LucaBEMan: I added both

tools/rag/rag_retriever.xml
LucaBEMan: I changed the container but added an extra macro for the image version because its not matching with the tool version

tools/rag/rag_retriever.xml
LucaBEMan: I have added the filter as suggested, but I am not sure if this aligns with what you had in mind.

tools/rag/rag_retriever.xml
LucaBEMan: I added some help to the parameter

tools/rag/tool-data/huggingface.loc.sample.loc.sample
LucaBEMan: Get it. Yes both is now possible

anuprulez:
These are my settings:

datatable1

datatable2

tableentry

data1


Tool UI

preinstalled

anuprulez:
@arash77 was it possible for you to test the PR? thanks!

arash77:
> @arash77 was it possible for you to test the PR? thanks!

Yes, tested it successfully! Here's a summary:

**1. Preinstalled model (data table):**
- Tested `sentence-transformers/all-MiniLM-L6-v2` and `BAAI/bge-small-en`
- Both loaded correctly from the `huggingface` data table, ran in the Docker container, and returned the expected output

**2. Upload custom model archive:**
- Created a `.tgz` archive of `all-MiniLM-L6-v2`, uploaded it, and ran the tool
- The `tar -xf ... --strip-components=1` extraction logic works correctly

**One thing worth flagging:** When a user uploads a `.tar.gz` file to Galaxy history using auto-detection, Galaxy assigns it the `gz` extension, but the tool's accepted formats are `tar.gz,tgz,tar,data`, which does **not** include plain `gz`. So the uploaded archive won't appear in the dataset selector. Users have to explicitly set the format to `tgz` when uploading. It might be worth either:
- Adding `gz` to the `format` attribute, or
- Adding a note in the help text saying users should set the file type to `tgz` when uploading

tools/rag/rag_retriever.xml
arash77: Let's also filter by version here.

LucaBEMan:
> > @arash77 was it possible for you to test the PR? thanks!
> 
> Yes, tested it successfully! Here's a summary:
> 
> **1. Preinstalled model (data table):**
> 
> * Tested `sentence-transformers/all-MiniLM-L6-v2` and `BAAI/bge-small-en`
> * Both loaded correctly from the `huggingface` data table, ran in the Docker container, and returned the expected output
> 
> **2. Upload custom model archive:**
> 
> * Created a `.tgz` archive of `all-MiniLM-L6-v2`, uploaded it, and ran the tool
> * The `tar -xf ... --strip-components=1` extraction logic works correctly
> 
> **One thing worth flagging:** When a user uploads a `.tar.gz` file to Galaxy history using auto-detection, Galaxy assigns it the `gz` extension, but the tool's accepted formats are `tar.gz,tgz,tar,data`, which does **not** include plain `gz`. So the uploaded archive won't appear in the dataset selector. Users have to explicitly set the format to `tgz` when uploading. It might be worth either:
> 
> * Adding `gz` to the `format` attribute, or
> * Adding a note in the help text saying users should set the file type to `tgz` when uploading

I will implement your first suggestion and add gz to the allowed formats to solve the visibility issue with auto-detection.

However, I wanted to point out that by allowing gz, users could also select 'raw' (single-file) gzip files, which would fail during the tar -xf extraction step. To mitigate this, I will add a note to the help section clarifying that the tool specifically requires a model archive (tar/tgz) and not just a single compressed file. I think this is the best trade-off for user-friendliness.

LucaBEMan:
> > @arash77 was it possible for you to test the PR? thanks!
> 
> Yes, tested it successfully! Here's a summary:
> 
> **1. Preinstalled model (data table):**
> 
> * Tested `sentence-transformers/all-MiniLM-L6-v2` and `BAAI/bge-small-en`
> * Both loaded correctly from the `huggingface` data table, ran in the Docker container, and returned the expected output
> 
> **2. Upload custom model archive:**
> 
> * Created a `.tgz` archive of `all-MiniLM-L6-v2`, uploaded it, and ran the tool
> * The `tar -xf ... --strip-components=1` extraction logic works correctly
> 
> **One thing worth flagging:** When a user uploads a `.tar.gz` file to Galaxy history using auto-detection, Galaxy assigns it the `gz` extension, but the tool's accepted formats are `tar.gz,tgz,tar,data`, which does **not** include plain `gz`. So the uploaded archive won't appear in the dataset selector. Users have to explicitly set the format to `tgz` when uploading. It might be worth either:
> 
> * Adding `gz` to the `format` attribute, or
> * Adding a note in the help text saying users should set the file type to `tgz` when uploading

Or we remove tar.gz completely as allowed format

arash77:
> Or we remove tar.gz completely as allowed format

I think this is a better solution for now.
Later, we could add the `tar.gz` data format to Galaxy itself if needed and adjust the tool based on that.

tools/rag/rag_retriever.xml
LucaBEMan: > Let's also filter by version here.

Should I restrict the tool to a specific version (e.g., only version 1) and ignore all others, or what is the intended logic for this filter?

arash77: Yes, please specify model version. This ensures that if the tool is updated and the older model becomes unusable, switching remains easy and older jobs stay reproducible.

LucaBEMan: > Yes, please specify model version. This ensures that if the tool is updated and the older model becomes unusable, switching remains easy and older jobs stay reproducible.

"I have now updated the tool to strictly filter for version '1' in the data table. This follows your recommendation to ensure reproducibility and stability for future updates.
If you had a different versioning format in mind (e.g., 'v1.0' instead of just '1'), please let me know.

LucaBEMan:
> > Or we remove tar.gz completely as allowed format
> 
> I think this is a better solution for now. Later, we could add the `tar.gz` data format to Galaxy itself if needed and adjust the tool based on that.

I removed it

arash77 (approved):
Thank you @LucaBEMan!

SaimMomin12 (approved):

PR #1825, opened 2026-03-30
Updating tools/rna_tools/ribotaper from version 1.3.1a to 1.3.1

gxydevbot:
Hello! This is an automated update of the following tool: **tools/rna_tools/ribotaper**. I created this PR because I think the tool's main dependency is out of date, i.e. there is a newer version available through conda.

I have updated tools/rna_tools/ribotaper from version 1.3.1a to 1.3.1.

**Project home page:** https://ohlerlab.mdc-berlin.de/software/RiboTaper_126

For any comments, queries or criticism about the bot, not related to the tool being updated in this PR, please create an issue [here](https://github.com/planemo-autoupdate/autoupdate/issues/new).

If you want to skip this change, close this PR without deleting the branch. It will be reopened if another change is detected.

Any commit from another author than 'planemo-autoupdate' will prevent more auto-updates.

To ignore manual changes and allow autoupdates, delete the branch.


tools/rna_tools/ribotaper/ribotaper_part3_main.xml
```diff
@@ -56,7 +56,7 @@
     </outputs>
     <tests>
         <test expect_num_outputs="8">
-            <param name="annotation_path" value="annotation_path.tgz"/>
+            <param name="annotation_path" value="annotation_path.tgz" ftpye="tar"/>
```
bgruening: tgz is not tar afaik

pavanvidem: The issue is that it is not a tgz, but the extension is tgz. Then it chooses the wrong command for unpacking. Specifying the ftype as tar fixed the issue.

bgruening: äh, ok, why then not renaming the file so the ending fits the content :)

pavanvidem: I was testing and it is a lazy fix ;) I will rename the file.

pavanvidem:
> .. WARNING (BioToolsValid): No entry ribotaper in bio.tools.

[it exists](https://bio.tools/ribotaper) and the previous linting worked

bgruening:
File-size also fails, and you can add a label to skip linting of the URLs

pavanvidem:
strange! why didn't it complain here? https://github.com/bgruening/galaxytools/actions/runs/24710190892

pavanvidem:
I checked again and the files are actually gzipped. 

Is there a way to skip the file size check?

pavanvidem:
@bgruening I think the issue is that with .tgz extension, it is already decompressing before running the tool. Hence, `tar -z` did not work.  With `.tar.gz` it is not the case.

PR #1826, opened 2026-03-30
Update TabPFN tool version and profile

anuprulez:


tools/tabpfn/tabpfn.xml
```diff
@@ -60,7 +60,7 @@
...
         <param name="test_data" type="data" format="tabular" label="Test data" help="Please provide test data for evaluating model. It may or may not contain labels/class/target in the last column"/>
         <param name="testhaslabels" type="boolean" truevalue="true" falsevalue="false" checked="false" label="Does test data contain labels?" help="Set this parameter when test data contains labels"/>
-        <param name="non_commercial" type="boolean" truevalue="true" falsevalue="false" checked="false" optional="false" label="Is the use of TabPFN non-commercial?" help="Set this parameter only for non-commercial purposes, and please reach out to sales@priorlabs.ai for commercial use"/>
+        <param name="non_commercial" type="boolean" truevalue="true" falsevalue="false" checked="true" optional="false" label="Is the use of TabPFN non-commercial?" help="Set this parameter only for non-commercial purposes, and please reach out to sales@priorlabs.ai for commercial use. Please read the license agreement: https://huggingface.co/Prior-Labs/tabpfn_2_5/blob/main/LICENSE"/>
```
bgruening: I guess the default should be that it is commercial and the tool can not be submitted. This way every user need to read this parameter.

And it only makes sense with a validator: https://github.com/galaxyproject/tools-iuc/blob/main/tools/meme/fimo.xml#L141

anuprulez: yes, that makes sense. Fixed it here: https://github.com/bgruening/galaxytools/pull/1826/changes/bdc0e1188eabaa5c1b0c3eadc5f5eca3345004db

tools/tabpfn/tabpfn.xml
```diff
@@ -1,8 +1,14 @@
-<tool id="tabpfn" name="Tabular data prediction using TabPFN" version="@TOOL_VERSION@+galaxy@VERSION_SUFFIX@" profile="23.0">
+<tool id="tabpfn" name="Tabular data prediction using TabPFN" version="@TOOL_VERSION@+galaxy@VERSION_SUFFIX@" profile="24.2">
     <description>with PyTorch</description>
     <macros>
-        <token name="@TOOL_VERSION@">2.0.9</token>
+        <token name="@TOOL_VERSION@">7.0.0</token>
         <token name="@VERSION_SUFFIX@">0</token>
+        <token name="@CHECK_NON_COMMERCIAL_USE@"><![CDATA[
```
bgruening: is that still needed?

tools/tabpfn/tabpfn.xml
```diff
@@ -17,14 +23,46 @@
...
+            <param name="source_type" type="select" label="Pretrained model source">
+                <option value="preinstalled">Use an installed model</option>
+                <option value="upload">Upload a custom model archive</option>
+            </param>
+            <when value="preinstalled">
+                <param name="pretrained_model" type="select" label="Select model">
+                <options from_data_table="huggingface">
+                    <column name="name" index="1" />
+                    <column name="value" index="6" />
+                    <filter type="static_value" value="tabular_prediction" column="4" />
```
bgruening: I mentioned that somewhere, we need to document those "ontology" that we are using to create this loc file somewhere. also ping @arash77

tools/tabpfn/tabpfn.xml
```diff
@@ -17,14 +23,46 @@
...
+                </options>
+                <validator type="no_options"
+                    message="No pretrained models are installed on this Galaxy instance." />
+                </param>
+            </when>
+            <when value="upload">
+                <param name="model_checkpoint" type="data" format="ckpt"
```
bgruening: What kind of filetype is that, is that safe?

bgruening: If not safe, lets drop it?

tools/tabpfn/tabpfn.xml
anuprulez: `fimo` and `flexynesis` tools use it from `macros`. Let me try without it. One observation is when the `non-commercial` button is unchecked, `run_tool` button is deactivated even after filling out all required datasets. When I check it, the `run_tool` becomes enabled. However, when I uncheck it again, `run_tool` still remains enabled.

anuprulez: Ok, its not needed. Neither calling `@CHECK_NON_COMMERCIAL_USE@` just before tool execution.

tools/tabpfn/tabpfn.xml
anuprulez: `ckpt` are model checkpoints (like training checkpoints). They are not that safe. But, they are directly getting processed by TabPFN and no where else these models paths are used. Unfortunately, they don't have `safetensors` format for models. Their latest set of models https://huggingface.co/Prior-Labs/tabpfn_2_6/tree/main also don't have `safetensors`

`TabPFNClassifier(random_state=42, model_path=args["model_path"])`

`TabPFNRegressor(random_state=42, model_path=args["model_path"])`

One possibility would be to just let user use the `ckpt` models provided by us via Huggingface table and exclude the feature `<option value="upload">Upload a custom model archive</option>` 

what do you think?

anuprulez: Optionally we can try to convert the `ckpt` files to `safetensors` ourselves in the tool and then pass them to TabPFN methods.

bgruening: As you like, the more unsafe file we can avoid the better.

anuprulez: i tried to convert the `.ckpt` files to `safetensors` but the TabPFN methods require not just tensor files but also metadata :(
 
 they have an open issue for this: https://github.com/PriorLabs/TabPFN/issues/204

anuprulez: I think let's drop the upgrade of this tool

tools/tabpfn/tool-data/huggingface.loc.sample
```diff
@@ -0,0 +1,19 @@
...
+prior-labs/tabpfn_2_5/tabpfn-v2.5-regressor-v2.5_quantiles	tabpfn-v2.5-regressor-v2.5_quantiles	tabular_prediction	tabular_prediction	tabular_prediction	1	prior-labs/tabpfn_2_5/tabpfn-v2.5-regressor-v2.5_quantiles.ckpt
+prior-labs/tabpfn_2_5/tabpfn-v2.5-regressor-v2.5_real	tabpfn-v2.5-regressor-v2.5_real	tabular_prediction	tabular_prediction	tabular_prediction	1	prior-labs/tabpfn_2_5/tabpfn-v2.5-regressor-v2.5_real.ckpt
+prior-labs/tabpfn_2_5/tabpfn-v2.5-regressor-v2.5_real-variant	tabpfn-v2.5-regressor-v2.5_real-variant	tabular_prediction	tabular_prediction	tabular_prediction	1	prior-labs/tabpfn_2_5/tabpfn-v2.5-regressor-v2.5_real-variant.ckpt
+prior-labs/tabpfn_2_5/tabpfn-v2.5-regressor-v2.5_small-samples	tabpfn-v2.5-regressor-v2.5_small-samples	tabular_prediction	tabular_prediction	tabular_prediction	1	prior-labs/tabpfn_2_5/tabpfn-v2.5-regressor-v2.5_small-samples.ckpt
+prior-labs/tabpfn_2_5/tabpfn-v2.5-regressor-v2.5_variant	tabpfn-v2.5-regressor-v2.5_variant	tabular_prediction	tabular_prediction	tabular_prediction	1	prior-labs/tabpfn_2_5/tabpfn-v2.5-regressor-v2.5_variant.ckpt
```
arash77: It would be better to use the official pipeline used on Hugging Face for these models, [tabular-classification](https://huggingface.co/models?pipeline_tag=tabular-classification), instead of `tabular_prediction`.

arash77: Also, the domain could be "tabular" and the free_tag could be "tabpfn." Also, we have been using the path to the directory rather than a specific file so far.

anuprulez: Thanks @arash77 for the suggestions but we are not going ahead with this PR because of `.ckpt` model files are not safe and they don't have models in `safetensors` format.

`tabular-classification` may not be entirely correct because there are models for both, classification and regression tasks. That's why I had used `tabular_prediction`.

anuprulez:
I am closing the PR because safe models are not yet supported in TabPFN.

tools/tabpfn/tool-data/huggingface.loc.sample
anuprulez: New changes: 
- Remove the feature to upload `.ckpt` files. Only pre-installed models are available
- `pipeline_tag, domain, free_tag` : classification	tabular	prediction (for classification models)
- `pipeline_tag, domain, free_tag`: regression	tabular	prediction (for regression models)
- Add to the helptext
**Help improve TabPFN**
                - The TabPFN project seeks new training datasets for improving models, and if you want to help this project, please visit https://ux.priorlabs.ai/playground.

anuprulez:
some more changes for this tool?

tools/tabpfn/tool-data/huggingface.loc.sample
```diff
@@ -0,0 +1,19 @@
...
+prior-labs/tabpfn_2_5/tabpfn-v2.5-regressor-v2.5_low-skew	tabpfn-v2.5-regressor-v2.5_low-skew	regression	tabular	prediction	1	prior-labs/tabpfn_2_5/tabpfn-v2.5-regressor-v2.5_low-skew.ckpt
+prior-labs/tabpfn_2_5/tabpfn-v2.5-regressor-v2.5_quantiles	tabpfn-v2.5-regressor-v2.5_quantiles	regression	tabular	prediction	1	prior-labs/tabpfn_2_5/tabpfn-v2.5-regressor-v2.5_quantiles.ckpt
+prior-labs/tabpfn_2_5/tabpfn-v2.5-regressor-v2.5_real	tabpfn-v2.5-regressor-v2.5_real	regression	tabular	prediction	1	prior-labs/tabpfn_2_5/tabpfn-v2.5-regressor-v2.5_real.ckpt
+prior-labs/tabpfn_2_5/tabpfn-v2.5-regressor-v2.5_real-variant	tabpfn-v2.5-regressor-v2.5_real-variant	regression	tabular	prediction	1	prior-labs/tabpfn_2_5/tabpfn-v2.5-regressor-v2.5_real-variant.ckpt
+prior-labs/tabpfn_2_5/tabpfn-v2.5-regressor-v2.5_small-samples	tabpfn-v2.5-regressor-v2.5_small-samples	regression	tabular	prediction	tabular_prediction	1	prior-labs/tabpfn_2_5/tabpfn-v2.5-regressor-v2.5_small-samples.ckpt
+prior-labs/tabpfn_2_5/tabpfn-v2.5-regressor-v2.5_variant	tabpfn-v2.5-regressor-v2.5_variant	regression	tabular	prediction	1	prior-labs/tabpfn_2_5/tabpfn-v2.5-regressor-v2.5_variant.ckpt
```
arash77: Four issues bundled in this file:

**1. Two malformed rows (8 columns instead of 7)**
- `tabpfn-v2.5-classifier-v2.5_real` (line 6): a stray `tabular_prediction` token is inserted between `name` and `pipeline_tag`, producing 8 TAB-separated fields instead of 7. Galaxy's data table parser will silently mismap every column from that row onward.
- `tabpfn-v2.5-regressor-v2.5_small-samples` (line 18): a stray `tabular_prediction` appears as an 8th field at the end — same issue.

**2. Non-official HuggingFace `pipeline_tag` values**
`classification` and `regression` are not in the HuggingFace tag vocabulary. The correct machine-readable tags are `tabular-classification` and `tabular-regression`. See the [HuggingFace models page](https://huggingface.co/models?pipeline_tag=tabular-classification) and the [shared schema reference](https://github.com/galaxyproject/galaxy-hub/pull/3923).

**3. Generic `free_tag = prediction`**
The `free_tag` column is the primary XML filter that isolates this tool's rows in the shared `huggingface.loc` file. A generic value like `prediction` risks collisions with future tools. The convention is a unique lowercase per-tool identifier — `tabpfn` is the natural choice.

**4. No documentation header**
The Flux tool's `.loc.sample` and the [schema reference post](https://github.com/galaxyproject/galaxy-hub/pull/3923) recommend a comment block explaining the schema, column meanings, download instructions, and a worked example so Galaxy administrators can populate the file correctly without guessing.

```suggestion
# This is a sample file distributed with Galaxy that is used to define HuggingFace
# pre-trained models for TabPFN tabular prediction, using 7 tab-separated columns.
# Lines starting with "#" are ignored by Galaxy.
#
# Columns (TAB-separated):
#   value	name	pipeline_tag	domain	free_tag	version	path
#
#   value         Unique row ID across the whole huggingface.loc table
#   name          Human-readable label shown in the Galaxy select widget
#   pipeline_tag  Official HuggingFace pipeline tag; TabPFN classifiers use
#                 "tabular-classification", regressors use "tabular-regression"
#                 (https://huggingface.co/models?pipeline_tag=tabular-classification)
#   domain        Coarse data domain (tabular / image / text / audio / ...)
#   free_tag      Per-tool-family namespace used as the primary XML filter;
#                 TabPFN rows use "tabpfn"
#   version       Tool version these rows belong to; used as a secondary XML filter
#   path          Path to the model checkpoint (.ckpt) or directory on this server
#
# Administrators: copy this file to your Galaxy tool-data directory as
# "huggingface.loc" and update each path to point to the actual model files.
# Models can be downloaded from: https://huggingface.co/prior-labs/tabpfn_2_5
#
# For the full shared schema reference see:
#   https://galaxyproject.org/news/2026-04-13-huggingface-data-table/
#
# Example (replace the path with the real location on your server):
# prior-labs/tabpfn_2_5/tabpfn-v2.5-classifier-v2.5_default	tabpfn-v2.5-classifier-v2.5_default	tabular-classification	tabular	tabpfn	1	/data/models/tabpfn-v2.5-classifier-v2.5_default.ckpt
#
# Classification models
prior-labs/tabpfn_2_5/tabpfn-v2.5-classifier-v2.5_default	tabpfn-v2.5-classifier-v2.5_default	tabular-classification	tabular	tabpfn	1	prior-labs/tabpfn_2_5/tabpfn-v2.5-classifier-v2.5_default.ckpt
prior-labs/tabpfn_2_5/tabpfn-v2.5-classifier-v2.5_default-2	tabpfn-v2.5-classifier-v2.5_default-2	tabular-classification	tabular	tabpfn	1	prior-labs/tabpfn_2_5/tabpfn-v2.5-classifier-v2.5_default-2.ckpt
prior-labs/tabpfn_2_5/tabpfn-v2.5-classifier-v2.5_large-features-L	tabpfn-v2.5-classifier-v2.5_large-features-L	tabular-classification	tabular	tabpfn	1	prior-labs/tabpfn_2_5/tabpfn-v2.5-classifier-v2.5_large-features-L.ckpt
prior-labs/tabpfn_2_5/tabpfn-v2.5-classifier-v2.5_large-features-XL	tabpfn-v2.5-classifier-v2.5_large-features-XL	tabular-classification	tabular	tabpfn	1	prior-labs/tabpfn_2_5/tabpfn-v2.5-classifier-v2.5_large-features-XL.ckpt
prior-labs/tabpfn_2_5/tabpfn-v2.5-classifier-v2.5_large-samples	tabpfn-v2.5-classifier-v2.5_large-samples	tabular-classification	tabular	tabpfn	1	prior-labs/tabpfn_2_5/tabpfn-v2.5-classifier-v2.5_large-samples.ckpt
prior-labs/tabpfn_2_5/tabpfn-v2.5-classifier-v2.5_real	tabpfn-v2.5-classifier-v2.5_real	tabular-classification	tabular	tabpfn	1	prior-labs/tabpfn_2_5/tabpfn-v2.5-classifier-v2.5_real.ckpt
prior-labs/tabpfn_2_5/tabpfn-v2.5-classifier-v2.5_real-large-features	tabpfn-v2.5-classifier-v2.5_real-large-features	tabular-classification	tabular	tabpfn	1	prior-labs/tabpfn_2_5/tabpfn-v2.5-classifier-v2.5_real-large-features.ckpt
prior-labs/tabpfn_2_5/tabpfn-v2.5-classifier-v2.5_real-large-samples-and-features	tabpfn-v2.5-classifier-v2.5_real-large-samples-and-features	tabular-classification	tabular	tabpfn	1	prior-labs/tabpfn_2_5/tabpfn-v2.5-classifier-v2.5_real-large-samples-and-features.ckpt
prior-labs/tabpfn_2_5/tabpfn-v2.5-classifier-v2.5_variant	tabpfn-v2.5-classifier-v2.5_variant	tabular-classification	tabular	tabpfn	1	prior-labs/tabpfn_2_5/tabpfn-v2.5-classifier-v2.5_variant.ckpt

# Regression models
prior-labs/tabpfn_2_5/tabpfn-v2.5-regressor-v2.5_default	tabpfn-v2.5-regressor-v2.5_default	tabular-regression	tabular	tabpfn	1	prior-labs/tabpfn_2_5/tabpfn-v2.5-regressor-v2.5_default.ckpt
prior-labs/tabpfn_2_5/tabpfn-v2.5-regressor-v2.5_low-skew	tabpfn-v2.5-regressor-v2.5_low-skew	tabular-regression	tabular	tabpfn	1	prior-labs/tabpfn_2_5/tabpfn-v2.5-regressor-v2.5_low-skew.ckpt
prior-labs/tabpfn_2_5/tabpfn-v2.5-regressor-v2.5_quantiles	tabpfn-v2.5-regressor-v2.5_quantiles	tabular-regression	tabular	tabpfn	1	prior-labs/tabpfn_2_5/tabpfn-v2.5-regressor-v2.5_quantiles.ckpt
prior-labs/tabpfn_2_5/tabpfn-v2.5-regressor-v2.5_real	tabpfn-v2.5-regressor-v2.5_real	tabular-regression	tabular	tabpfn	1	prior-labs/tabpfn_2_5/tabpfn-v2.5-regressor-v2.5_real.ckpt
prior-labs/tabpfn_2_5/tabpfn-v2.5-regressor-v2.5_real-variant	tabpfn-v2.5-regressor-v2.5_real-variant	tabular-regression	tabular	tabpfn	1	prior-labs/tabpfn_2_5/tabpfn-v2.5-regressor-v2.5_real-variant.ckpt
prior-labs/tabpfn_2_5/tabpfn-v2.5-regressor-v2.5_small-samples	tabpfn-v2.5-regressor-v2.5_small-samples	tabular-regression	tabular	tabpfn	1	prior-labs/tabpfn_2_5/tabpfn-v2.5-regressor-v2.5_small-samples.ckpt
prior-labs/tabpfn_2_5/tabpfn-v2.5-regressor-v2.5_variant	tabpfn-v2.5-regressor-v2.5_variant	tabular-regression	tabular	tabpfn	1	prior-labs/tabpfn_2_5/tabpfn-v2.5-regressor-v2.5_variant.ckpt
```

tools/tabpfn/tool_data_table_conf.xml.sample
```diff
@@ -0,0 +1,7 @@
+<tables>
+    <!-- Locations of all fasta files under genome directory -->
```
arash77: This comment was copied from another tool (likely the genome reference data tables). It is misleading in the context of HuggingFace models.

```suggestion
    <!-- huggingface models -->
```

tools/tabpfn/tabpfn.xml
```diff
@@ -18,17 +18,37 @@
...
+            <param name="source_type" type="select" label="Pretrained model source">
+                <option value="preinstalled">Use an installed model</option>
+            </param>
+            <when value="preinstalled">
+                <param name="pretrained_model" type="select" label="Select model">
+                <options from_data_table="huggingface">
+                    <column name="name" index="1" />
+                    <column name="value" index="6" />
+                    <filter type="static_value" value="prediction" column="4" />
+                </options>
```
arash77, line 47: Two issues here:

**1. `free_tag` filter value is `prediction` but the `.loc.sample` registers a different value**
After fixing `free_tag` to `tabpfn` in the data file (see other comment), this filter must match. Even without that change, `prediction` is too generic — once another tool shares the `huggingface.loc` file it could accidentally see TabPFN's rows.

**2. No version filter**
The [shared schema convention](https://github.com/galaxyproject/galaxy-hub/pull/3923) recommends filtering by both `free_tag` (column 4) **and** `version` (column 5) so that when a future tool version adds new model rows the old rows are automatically excluded for older tool instances.

```suggestion
                <options from_data_table="huggingface">
                    <column name="name" index="1" />
                    <column name="value" index="6" />
                    <filter type="static_value" value="tabpfn" column="4" />
                    <filter type="static_value" value="1" column="5" />
                </options>
```

arash77 (changes requested):
Thanks for the PR! I found a few schema conformance issues to address before merging. All three files have one-click suggestions attached — see the inline comments for details.

Summary of issues:

**`tool-data/huggingface.loc.sample`**
- Two rows have 8 tab-separated columns instead of 7 (`_real` classifier and `_small-samples` regressor) — a stray `tabular_prediction` token in each breaks the column mapping for that row
- `pipeline_tag` uses `classification` / `regression` which are not in the HuggingFace tag vocabulary; the correct tags are `tabular-classification` / `tabular-regression`
- `free_tag = prediction` is too generic; per the shared schema convention it should be a unique per-tool identifier (`tabpfn`)
- No documentation header to guide administrators setting up the file

**`tool_data_table_conf.xml.sample`**
- The comment `<!-- Locations of all fasta files under genome directory -->` was copied from a genome reference table and is wrong; replaced with `<!-- huggingface models -->`

**`tabpfn.xml`**
- The model select filter references `free_tag = prediction` (which must match the `.loc` file)
- Missing version filter on column 5 — recommended by the [shared schema reference](https://github.com/galaxyproject/galaxy-hub/pull/3923) to future-proof row isolation

arash77 (approved):

anuprulez:
thanks a lot @arash77

bgruening:
Pretend that I'm already out and find someone that merges this PR, please :)

SaimMomin12 (approved):

PR #1834, opened 2026-04-08
Image Magick conversion: fix missing font error for svg conversion

Delphine-L:
FOR CONTRIBUTOR:
* [ ] I have read the [CONTRIBUTING.md](https://github.com/galaxyproject/tools-iuc/blob/master/CONTRIBUTING.md) document and this tool is appropriate for the tools-iuc repo.
* [ ] License permits unrestricted use (educational + commercial)


bgruening:
Can you please add a test for this? Shouldn't the font file be the same in the container for all installations?

SaimMomin12:
@Delphine-L do you need any help here?

PR #1838, opened 2026-04-16
bump AGAT to v1.7.0

d-callan:
theres a bug fix in 1.7 that i want, but looks like some linting issues were standing in the way of that propagating automagically [here](https://github.com/bgruening/galaxytools/pull/1507) so im attempting to fix them :crossed_fingers: 

FOR CONTRIBUTOR:


bgruening:
Thanks @d-callan 

I see those errors here:

> cpu parameter missing in the configuration file.
minimum_chunk_size parameter missing in the configuration file.
url_encode_out parameter missing in the configuration file.

tools/agat/agat.xml
```diff
@@ -170,6 +170,11 @@ progress_bar: true
 log: true
 debug: false
 tabix: false
+cpu: 1
```
bgruening: Is one CPU usually enough? Otherwise we can add there some sed-able placeholder like REPLACE_ME_CORE_COUNT and then we replace that in the command section with GALAXY_SLOTS

d-callan: ive no good guess tbh. i like this idea.

bgruening (approved):

bgruening:
merci!

Issue #1808, opened 2026-03-13
tiara tool still not working reliably


wm75:
@Minamehr @bgruening after #1526 there still seem to be issues remaining with the tiara xml.

Apparently their can be cases when no files get produced for a selected taxonomy filter and in that case the mv/rename part of the wrapper is failing.

I don't know much about the tool, but if it's producing .dat files, then it's likely just keeping the input file suffix?
In that case a better approach than trying to rename all outputs, would be to symlink the input to get a proper suffix from the start?

PR #1846, opened 2026-04-21
Fix Tiara wrapper when outputs are missing

Minamehr:
@wm75 @bgruening following the #1808.

FOR CONTRIBUTOR:

### DIFF
--- a/tools/tiara/macros.xml
+++ b/tools/tiara/macros.xml
@@ -1,6 +1,7 @@
 <macros>
     <token name="@TOOL_VERSION@">1.0.3</token>
-    <token name="@VERSION_SUFFIX@">galaxy0</token>
+    <token name="@VERSION_SUFFIX@">2</token>
+    <token name="@PROFILE@">25.1</token>
     <xml name="requirements">
         <requirements>
             <requirement type="package" version="@TOOL_VERSION@">tiara</requirement>
--- a/tools/tiara/macros.xml
+++ b/tools/tiara/macros.xml
@@ -5,7 +5,6 @@
     <xml name="requirements">
         <requirements>
             <requirement type="package" version="@TOOL_VERSION@">tiara</requirement>
-            <requirement type="package" version="3.8">python</requirement>
         </requirements>
     </xml>
     <xml name="biotools">
--- a/tools/tiara/tiara.xml
+++ b/tools/tiara/tiara.xml
@@ -1,4 +1,4 @@
-<tool id="tiara" name="tiara" version="@TOOL_VERSION@+galaxy2" profile="21.05">
+<tool id="tiara" name="tiara" version="@TOOL_VERSION@+galaxy@VERSION_SUFFIX@" profile="@PROFILE@">
     <description>Deep-learning-based approach for identification of eukaryotic sequences in the metagenomic data </description>
     <macros>
         <import>macros.xml</import>
--- a/tools/tiara/tiara.xml
+++ b/tools/tiara/tiara.xml
@@ -7,7 +7,6 @@
     <expand macro="requirements"/>
     <command detect_errors="exit_code"><![CDATA[ 
         mkdir -p ./txt &&
-        mkdir -p ./fasta &&
 
         tiara 
         -t \${GALAXY_SLOTS:-4} 
@@ -33,9 +32,6 @@
         #if $taxonomy_filter
             --tf #for $tf in $taxonomy_filter
                 $tf
-            #end for
-            #for $tf in $taxonomy_filter
-              && find ./txt/ -name ${tf}*.dat -exec mv {} ./fasta/${tf}.fasta ';'
             #end for
         #end if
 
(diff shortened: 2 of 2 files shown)


tools/tiara/tiara.xml
```diff
@@ -1,16 +1,18 @@
-<tool id="tiara" name="tiara" version="@TOOL_VERSION@+galaxy1" profile="21.05">
+<tool id="tiara" name="tiara" version="@TOOL_VERSION@+galaxy2" profile="21.05">
```
SaimMomin12:
```suggestion
<tool id="tiara" name="tiara" version="@TOOL_VERSION@+galaxy@VERSION_SUFFiX@" profile="@PROFILE@">
```
Can we introduce a VERSION_SUFFIX token in macros and bump the PROFILE to more recent version 25.1?

tools/tiara/tiara.xml
```diff
@@ -78,38 +80,47 @@
         </section>
     </inputs>
     <outputs>
-        <collection name="output" type="list" label="${tool.name} on ${on_string}: classified sequences in txt and Fasta Output">
-            <discover_datasets pattern="__name_and_ext__"  ext="fasta,txt" directory="results" />
+        <collection name="output_fasta" type="list" label="${tool.name} on ${on_string}: Fasta Output">
+            <discover_datasets pattern="__name_and_ext__"  ext="fasta" directory="fasta" />
```
wm75: Is the `find` business in the command section really necessary?

Would sth like:

```suggestion
            <discover_datasets pattern="(?P&lt;designation&gt;.+)\.dat$" directory="txt" format="fasta"/>
```

not work directly?

Minamehr: I tested this, but it failed. It works in principle, but due to how tiara names the outputs, it breaks the expected identifiers, so I kept the current approach for now.

wm75: Can you be a bit more specific on this?
What are the names of the outputs, what would you like the identifiers to be and what are you getting instead. I feel this really *should* be solvable with the framework instead of via the command section.

Minamehr: You’re right, my bad. The issue was Tiara’s output naming. Fixed now and updated tests.

tools/tiara/tiara.xml
```diff
@@ -78,38 +76,47 @@
...
     </inputs>
     <outputs>
-        <collection name="output" type="list" label="${tool.name} on ${on_string}: classified sequences in txt and Fasta Output">
-            <discover_datasets pattern="__name_and_ext__"  ext="fasta,txt" directory="results" />
+        <collection name="output_fasta" type="list" label="${tool.name} on ${on_string}: Fasta Output">
+            <discover_datasets pattern="(?P&lt;designation&gt;[^_]+)_dataset_.*\.dat$"  format="fasta" directory="txt" />
+        </collection>
+        <collection name="output_txt" type="list" label="${tool.name} on ${on_string}: Classified sequences in txt">
+            <discover_datasets pattern="(?P&lt;designation&gt;.+)\.txt$"  format="txt" directory="txt" />
         </collection>
```
wm75: Are there ever more txt files than the:

- main results and the
- log output

?

If not, then I don't think this is a good use of a collection and both outputs should just be added as individual datatsets. You can also discover them with `from_work_dir` and their known names.

tools/tiara/macros.xml
wm75: Why is there a Python requirement here? When I'm installing the bioconda package I'm getting Python 3.8 anyway.

Am I missing sth here or is this just some historic artifact?

wm75 (commented):
Much better already, thanks!

tools/tiara/tiara.xml
```diff
@@ -32,21 +33,18 @@
...
-              && ls -l ./results/
-              &&  mv ./results/${tf}*.dat ./results/${tf}.fasta
-            #end for
         #end if
 
     ]]></command>
     <inputs>
         <param name="input" type="data" format="fasta" label="input fasta,fasta.gz file"/>
-        <param name="taxonomy_filter" type="select" multiple="true" optional="true" label="Write sequences to fasta,fasta.gz files specified in the arguments to this option." help="all refers to all classes present in input fasta (to separate fasta files).">
+        <param name="taxonomy_filter" type="select" multiple="true" optional="true" label="Write sequences to fasta,fasta.gz files specified in the arguments to this option." help="Select one or more target types of sequences to write to separate FASTA files.">
```
wm75: Instead of making this param optional (and producing an empty collection if the user selects nothing), I would rather make the param required, but put it inside a conditional and after a choice param that asks whether split fasta output should be produced at all.
Then, in outputs, you'd create the collection only if the user selected Yes.

tools/tiara/tiara.xml
wm75: Also a more specific output format for the main results is `tabular`.

bgruening (approved):

PR #1850, opened 2026-04-23
Refine font size parameter descriptions in wordcloud.xml

Sch-Da:
As suggested by @RZ9082, I updated the help text for the min_font_size parameter to clarify its effect on word frequency scaling. I deleted extra spaces and renamed the label as well.

@RZ9082 could you please check if this makes sense and is syntactically correct here? Thanks a lot for pointing this out!


tools/wordcloud/wordcloud.xml
RZ9082:
```suggestion
<token name="@TOOL_VERSION@">1.9.6</token>
<token name="@VERSION_SUFFIX@">0</token>
```

@Sch-Da  could you please bump the tool version and set the suffix back to 0?

RZ9082 (commented):
Looks good, thank you!

Sch-Da:
Thanks a lot for updating the tests @RZ9082

PR #1864, opened 2026-05-13
Add PECAT tool

SaimMomin12:
FOR CONTRIBUTOR:


tools/pecat/.shed.yml
```diff
@@ -0,0 +1,13 @@
+categories:
+- Assembly
+description: Phased error correction and assembly tool for long reads
+long_description: |
+  PECAT is a phased error correction and assembly pipeline for diploid genomes
+  from PacBio CLR, Oxford Nanopore (ONT), or PacBio HiFi long-read data. It
+  performs haplotype-aware error correction, phased string-graph assembly, and
+  Racon consensus polishing.
+name: pecat
+owner: iuc
```
SaimMomin12:
```suggestion
owner: bgruening
```

tools/pecat/.shed.yml
```diff
@@ -0,0 +1,13 @@
...
+description: Phased error correction and assembly tool for long reads
+long_description: |
+  PECAT is a phased error correction and assembly pipeline for diploid genomes
+  from PacBio CLR, Oxford Nanopore (ONT), or PacBio HiFi long-read data. It
+  performs haplotype-aware error correction, phased string-graph assembly, and
+  Racon consensus polishing.
+name: pecat
+owner: bgruening
+homepage_url: https://github.com/lemene/PECAT
+remote_repository_url: https://github.com/galaxyproject/tools-iuc/tree/main/tools/pecat
```
RZ9082:
```suggestion
remote_repository_url: https://github.com/bgruening/galaxytools/tree/master/tools/pecat
```

tools/pecat/macros.xml
```diff
@@ -0,0 +1,194 @@
...
+        <citations>
+            <citation type="doi">10.5281/zenodo.8380113</citation>
+            <citation type="bibtex">
+                @misc{githubPECAT,
+                    author = {Nie, Fan},
+                    title = {PECAT: Phased Error Correction and Assembly Tool},
+                    publisher = {GitHub},
+                    journal = {GitHub repository},
+                    url = {https://github.com/lemene/PECAT}}
+            </citation>
```
RZ9082:
```suggestion
            <citation type="doi">10.1101/2022.09.25.509436 </citation>
```

tools/pecat/pecat.xml
```diff
@@ -0,0 +1,335 @@
...
+    <command detect_errors="aggressive"><![CDATA[
+        #if $reads_input.is_of_type('fasta.gz')
+            #set $ext = '.fasta.gz'
+        #elif $reads_input.is_of_type('fastqsanger.gz')
+            #set $ext = '.fastq.gz'
+        #elif $reads_input.is_of_type('fastqsanger')
+            #set $ext = '.fastq'
+        #else
+            #set $ext = '.fasta'
+        #end if
```
RZ9082: This block can be moved to macros as it is identical with the one in config file

tools/pecat/pecat.xml
```diff
@@ -0,0 +1,335 @@
...
+
+polish_use_reads=$platform.polish.polish_use_reads
+polish_map_options=$platform.polish.polish_map_options
+polish_filter_options=$platform.polish.polish_filter_options
+polish_cns_options=$platform.polish.polish_cns_options
+        ]]></configfile>
+    </configfiles>
+    <inputs>
+        <param name="reads_input" type="data" format="fasta,fasta.gz,fastqsanger,fastqsanger.gz" label="Long reads" help="PacBio CLR, Oxford Nanopore (ONT), or PacBio HiFi long reads"/>
+        <param argument="genome_size" type="text" value="" label="Estimated genome size (bp)" help="Integer number of bases (e.g. 130000000 for a 130 Mb genome)">
```
RZ9082: type should be integer? default, min, max?

tools/pecat/pecat.xml
```diff
@@ -0,0 +1,335 @@
...
+            #set $ext = '.fasta.gz'
+        #elif $reads_input.is_of_type('fastqsanger.gz')
+            #set $ext = '.fastq.gz'
+        #elif $reads_input.is_of_type('fastqsanger')
+            #set $ext = '.fastq'
+        #else
+            #set $ext = '.fasta'
+        #end if
+        ln -s '$reads_input' 'reads${ext}' &&
+        cp '$pecat_cfg' pecat_config.txt &&
```
RZ9082:
```suggestion
        ln -s '$pecat_cfg' pecat_config.txt &&
```

tools/pecat/pecat.xml
```diff
@@ -0,0 +1,335 @@
...
+                <param name="preset" value="clr"/>
+                <section name="prep">
+                    <param name="prep_min_length" value="3000"/>
+                    <param name="prep_output_coverage" value="80"/>
+                </section>
+                <section name="corr">
+                    <param name="corr_iterate_number" value="1"/>
+                    <param name="corr_block_size" value="4000000000"/>
+                    <param name="corr_rd2rd_options" value="-x ava-pb"/>
+                    <param name="corr_filter_options" value="--filter0=:al=5000:alr=0.5:aalr=0.5:oh=1000:ohr=0.1"/>
```
RZ9082, line 101: I think it is better to split them into individual parameters/arguments. This applies also to the other similar cases in this wrapper.

tools/pecat/macros.xml
```diff
@@ -0,0 +1,194 @@
...
+                <option value="0" selected="@USE_READS_0_SELECTED@">Raw reads</option>
+                <option value="1" selected="@USE_READS_1_SELECTED@">Corrected reads</option>
+            </param>
+            <param argument="polish_map_options" type="text" value="@MAP_OPTIONS@" label="minimap2 options for assembly-to-reads mapping (polish_map_options)">
+                <expand macro="opts_sanitizer"/>
+            </param>
+            <param argument="polish_filter_options" type="text" value="@FILTER_OPTIONS@" optional="true" label="Polish filter options (polish_filter_options)">
+                <expand macro="opts_sanitizer"/>
+            </param>
+            <param argument="polish_cns_options" type="text" value="" optional="true" label="Racon options (polish_cns_options)">
```
RZ9082, line 127: Defaults?

RZ9082 (commented):
Thanks @SaimMomin12, I have left some comments/suggestions.

tools/pecat/pecat.xml
SaimMomin12: Genome size cannot really be pre-set, since this parameter varies from organism to organism. The only requirement is that it should accept a positive integer value.

tools/pecat/pecat.xml
SaimMomin12: I would not invest too much effort into splitting them, since these parameters are only exposed as advanced options. At this stage, we mainly need the tool to support a user’s publication, so keeping the implementation simpler is probably sufficient for now.

tools/pecat/macros.xml
SaimMomin12: If the values are pre-set, they would end up in the config file and effectively make polishing mode the default behavior, which I do not think we want. Since these are intended as advanced options, it is probably better to keep them explicitly user-controlled.

tools/pecat/pecat.xml
RZ9082: In that case please set a minimum.

RZ9082 (approved):

PR #1869, opened 2026-05-18
Fix mOTUs wrapper

Minamehr:
FOR CONTRIBUTOR:

This PR updates the mOTUs profiler wrapper.
The mOTUs database is large, so only the existing lightweight wrapper test is kept in the tool XML, but Tested single-end, paired-end, and paired-end collection inputs locally.


tools/motus/macros.xml
```diff
@@ -12,11 +12,11 @@
     </xml>
     <token name="@TOOL_VERSION@">3.1.0</token>
     <token name="@PROFILE@">21.05</token>
-    <token name="@VERSION_SUFFIX@">0</token>
+    <token name="@VERSION_SUFFIX@">2</token>
```
SaimMomin12, line 15:
```suggestion
    <token name="@VERSION_SUFFIX@">1</token>
```

tools/motus/macros.xml
```diff
@@ -12,11 +12,11 @@
     </xml>
     <token name="@TOOL_VERSION@">3.1.0</token>
     <token name="@PROFILE@">21.05</token>
```
SaimMomin12:
```suggestion
    <token name="@PROFILE@">25.0</token>
```

tools/motus/motus_profiler.xml
```diff
@@ -84,18 +89,27 @@
...
                             <param name="reverse_reads" type="data" format="fastq,fastq.gz" label="Reverse reads" help="-r Input file(s) for reads in reverse orientation, fastq(.gz)-formatted"/>
                         </when>
+                        <when value="paired_collection">
+                            <param name="paired_collection"
+                                type="data_collection"
+                                collection_type="paired"
+                                format="fastq,fastq.gz"
+                                label="Paired-end read collection"
+                                help="A paired-end collection containing forward and reverse reads for one sample."/>
+                        </when>
```
SaimMomin12, line 108: Can we do `planemo format motus_profiler.xml` on this file?

Minamehr: Sure,I will fix that, thanks.

tools/motus/macros.xml
Minamehr: Thanks! The original `motus_profiler.xml` was already using `version="@TOOL_VERSION@+galaxy1"`. Since I changed it to use the shared macro, `version="@TOOL_VERSION@+galaxy@VERSION_SUFFIX@"`, I think `@VERSION_SUFFIX@` needs to stay at `2` so the resolved version becomes `3.1.0+galaxy2`.

tools/motus/motus_profiler.xml
```diff
@@ -84,18 +89,22 @@
...
                             <option value="paired">Paired-end</option>
+                            <option value="paired_collection">Paired-end collection</option>
                             <option value="single" selected="true">Single</option>
                         </param>
                         <when value="paired">
                             <param name="forward_reads" type="data" format="fastq,fastq.gz" label="Forward reads" help="-f Input file(s) for reads in forward orientation, fastq(.gz)-formatted"/>
                             <param name="reverse_reads" type="data" format="fastq,fastq.gz" label="Reverse reads" help="-r Input file(s) for reads in reverse orientation, fastq(.gz)-formatted"/>
                         </when>
+                        <when value="paired_collection">
+                            <param name="paired_collection" type="data_collection" collection_type="paired" format="fastq,fastq.gz" label="Paired-end FASTQ collection"/>
```
bernt-matthias: Use fastqsanger and fastqsanger.gz instead of fastq,fastq.gz (here and everywhere else). If needed add fastqillumina and fastqillumina.gz.

tools/motus/motus_profiler.xml
```diff
@@ -127,51 +136,48 @@
...
-        <conditional name="cami_format_option">
-            <param argument="cami_options" type="select" label="Choose cami_format">
-                <option value="yes">yes</option>
-                <option value="no" selected="True">no</option>
+        <conditional name="output_type">
+            <param name="selector" type="select" label="Main output type">
+                <option value="default" selected="true">Default mOTU profile</option>
+                <option value="biom">BIOM format</option>
+                <option value="ref_motus">Only species with reference genomes (ref-mOTUs)</option>
+                <option value="full_species_name">Print full species names</option>
```
bernt-matthias: Can you mention which command line parameters is covered by the options, e.g. add `(-u)` to the end of the option text.

tools/motus/motus_profiler.xml
```diff
@@ -127,51 +136,48 @@
...
-        <param argument="biom_format" type="boolean" label="print result in BIOM format" truevalue="-B" falsevalue="" checked="false"/>
-        <param argument="full_rank_taxonomy" type="boolean" label="print the full rank taxonomy" truevalue="-q" falsevalue="" checked="false"/>
-        <param argument="all_taxonomic_level" type="boolean" label="print all taxonomic levels together" help="-A print all taxonomic levels together (kingdom to mOTUs, override -k)." truevalue="-A" falsevalue="" checked="false"/>
-        <param argument="save_bam" type="boolean" label="BAM output" help="-I save the result of BWA in BAM format (output of motus map_tax)." truevalue="-I" falsevalue="" checked="false"/>
-        <param argument="save_mgc" type="boolean" label="Mgc output" help="-M save the mgc reads count(output of motus calc_mgc)." truevalue="-M" falsevalue="" checked="false"/>
-        <conditional name="cami_format_option">
-            <param argument="cami_options" type="select" label="Choose cami_format">
-                <option value="yes">yes</option>
-                <option value="no" selected="True">no</option>
+        <conditional name="output_type">
```
bernt-matthias, line 140: The downside of this change is that the user can only select one output. Is this intended?

Minamehr: thanks! I replaced `fastq/fastq.gz` with `fastqsanger/fastqsanger.gz`, added the command-line flags to the relevant option labels, and adjusted the output logic after checking the actual mOTUs behavior.

tools/motus/motus_profiler.xml
```diff
@@ -1,80 +1,85 @@
...
-            #set $DATABASE_DIR = 'database_dir/db_mOTU' ##the actual DB folder in the archive called db_mOTU
+            mkdir -p database_dir &&
+            tar -xvf '${db.database}' -C database_dir/ &&
+            DATABASE_DIR='database_dir' &&
         #end if
-        echo "Database Directory:" $DATABASE_DIR  &&
+        if [ -d "\${DATABASE_DIR}/db_mOTU" ]; then
+            DATABASE_DIR="\${DATABASE_DIR}/db_mOTU";
+        fi &&
+        echo "Database Directory: \${DATABASE_DIR}" &&
```
bernt-matthias:
```suggestion
        echo "Database Directory: \$DATABASE_DIR} &&
```

Can you remove this from all shell variables?

tools/motus/motus_profiler.xml
```diff
@@ -1,80 +1,85 @@
...
+        #elif $inputs.input_type.selector == 'mgc_read_count'
+            -m '${inputs.input_type.mgc_read_count}'
         #end if
+       
         #if $algorithm_advanced_options.advanced_options.customize_algorithm_options == 'customize'
-            -g $algorithm_advanced_options.advanced_options.marker_genes_cutoff
-            -l $algorithm_advanced_options.advanced_options.minimum_alignment_len
-            -v $algorithm_advanced_options.advanced_options.verbosity_level_selection
-            -y $algorithm_advanced_options.advanced_options.read_count_type_selection
+            -g '${algorithm_advanced_options.advanced_options.marker_genes_cutoff}'
```
bernt-matthias: Curly braces are also net needed for cheetah variables (except if you want to follow it by a character that could be part of the cheetah variable). Please remove them

tools/motus/motus_profiler.xml
```diff
@@ -84,18 +89,27 @@
...
                         </param>
                         <when value="paired">
                             <param name="forward_reads" type="data" format="fastq,fastq.gz" label="Forward reads" help="-f Input file(s) for reads in forward orientation, fastq(.gz)-formatted"/>
                             <param name="reverse_reads" type="data" format="fastq,fastq.gz" label="Reverse reads" help="-r Input file(s) for reads in reverse orientation, fastq(.gz)-formatted"/>
                         </when>
+                        <when value="paired_collection">
+                            <param name="paired_collection"
+                                type="data_collection"
+                                collection_type="paired"
+                                format="fastq,fastq.gz"
```
bernt-matthias: Please do not use unqualified fastq, but fastqsanger and/or fastqillumina .. whatever applies

mvdbeek:
The SameStr mOTUs data manager (`iuc/data_manager_samestr`) has the same `motus_db_versioned.path` assumption this PR is fixing for the profiler — and it trips over it. Its mOTUs branch hard-codes:

```
#set $db_path = $db_source.motus_db.fields.path
#set $markers_info1 = $db_path + '/db_mOTU/db_mOTU_taxonomy_ref-mOTUs.tsv'
...
cp '$db_path'/db_mOTU/db_mOTU_versions ...
```

i.e. it assumes `path` is the **parent** of `db_mOTU`. But the mOTUs DM's `value_translation` (`${GALAXY_DATA_MANAGER_DATA_PATH}/motus_database/${value}/db_mOTU`) and this PR's new `.loc.sample` (`path = /data/motus/db_mOTU`) make `path` the **`db_mOTU` folder itself** — so SameStr's `$path/db_mOTU/...` doubles and fails (`cp: can't stat .../db_mOTU/db_mOTU_versions: No such file`).

We hit this building an mOTUs-derived SameStr DB as a reference-data bundle for idc. It works on usegalaxy.eu only because that particular install happens to have `path` = the parent (a `db_mOTU/` subdir), which isn't universal.

Two questions:
1. Is `path` = the `db_mOTU` folder the intended **canonical** convention going forward (per the new `.loc.sample` + the `value_translation`)?
2. Would you be open to SameStr adopting the same robust probe you added here —
   ```
   if [ -d "$db_path/db_mOTU" ]; then db_path="$db_path/db_mOTU"; fi
   ```
   then reading the markers from `$db_path` directly — so both consumers agree regardless of which convention an install uses?

If that's the direction, we're happy to send the SameStr PR mirroring this. cc @SaimMomin12 (SameStr DM author + reviewer here) @Minamehr @bernt-matthias

SaimMomin12:
@Minamehr Can you please check with this?

mvdbeek:
Friendly ping on this — @Minamehr @SaimMomin12: any thoughts on whether `path = the db_mOTU folder` is the intended canonical convention (per this PR's `.loc.sample` + the DM `value_translation`)?

To make the concrete proposal easy to look at, I opened the analogous SameStr fix as a **draft**: galaxyproject/tools-iuc#8346 — it uses the exact same `if [ -d "$path/db_mOTU" ]` probe you added here, so both consumers agree regardless of which convention an install uses. Happy to adjust it to whatever convention you settle on.

bernt-matthias:
Having `db_mOTU/` subdir seems not necessary, but changing this might break existing installations? Maybe better change the dir at eu?

mvdbeek:
That would be the cleaner fix, but is this not breaking any tools ? Has mOTUs never worked against a reference created by the data manager ?

bernt-matthias:
If the tool never worked with the `db_mOTU/` in the path we can and should remove it. But I do not know this. 

@Minamehr ?

Minamehr:
Thanks, I checked the mOTUs 3.1.0 code again.
From the mOTUs side, -db should point to the actual database directory. mOTUs derives the database prefix from the basename of the path passed to -db and then looks for `<db_path>/<basename>_versions`.
For the current mOTUs database, this seems consistent with motus_db_versioned.path pointing to the db_mOTU directory itself.
I cannot say for all older installations, but I agree that using the same path normalization in SameStr sounds like the safest option.

mvdbeek:
Diagnosis, now confirmed with the actual usegalaxy.eu data:

**On disk** (`tree` of the DM output on EU): `/data/db/data_managers/motus_database/db_from_2026-04-27T094930Z/db_mOTU/` holds the database, all files prefixed `db_mOTU_`. The `db_from_…` directory contains nothing but `db_mOTU`.

**In the EU data table** (admin view): `path` = `/data/db/data_managers/motus_database/db_from_2026-04-27T094930Z` — the **parent**, without `/db_mOTU`.

**What the data manager actually writes**: I ran the DM under planemo against a real Galaxy (same identifier override as the EU row). Galaxy's `<move>` + `value_translation` produced

```
db_from_2026-04-27T094930Z	3.1.0	mOTUs DB version 3.1.0 (db_from_2026-04-27T094930Z)	<data_manager_path>/motus_database/db_from_2026-04-27T094930Z/db_mOTU
```

i.e. `path` ends in `/db_mOTU`. The EU row does not (and its `name` text was never produced by any released fetcher), so it was edited after the DM ran. Every other install that runs the DM gets the `/db_mOTU` form.

**mOTUs 3.1.0 requires the `db_mOTU` form**: `motus.py` takes the basename of `-db` as the file prefix and opens `<db>/<basename>_versions`. With the parent path it looks for `db_from_…_versions` and exits. So the released profiler (which passes `path` straight to `-db`) cannot have worked against the EU cached DB, while SameStr's hard-coded `$path/db_mOTU/…` works *only* there.

**Plan**
- Fix the EU `motus_db_versioned` entry to the DM location (`…/db_from_2026-04-27T094930Z/db_mOTU`). The released mOTUs profiler then works as-is; the probe in this PR remains harmless and protects other installs.
- That change breaks the **SameStr data manager**'s mOTUs branch on EU (`iuc/data_manager_samestr_db`, hard-coded `$path/db_mOTU/…`) — galaxyproject/tools-iuc#8346 fixes it with the same probe and should land first. The SameStr *tools* only read `samestr_db` and are unaffected, as is the already-built mOTUs-derived SameStr DB on EU.
- Keep the DM unchanged: `path` = the `db_mOTU` directory is the canonical convention, and dropping the subdir is not possible since mOTUs needs that basename.

(Side note for anyone checking via `/api/tool_data/motus_db_versioned`: for non-admins Galaxy strips `path` to its basename, so it shows `db_from_2026-04-27T094930Z` there — that is not the stored value.)

PR #1880, opened 2026-05-28
Addresses https://github.com/galaxyproject/tools-iuc/issues/7999

scottcain:
(non-alphanumeric characters in genome names causes failure)

FOR CONTRIBUTOR:

## Description

For HiCUP tools that have `--genome` as an option, HiCUP has a strict rule that only alphanumeric characters plus underscores are allowed, meaning that a genome with a label like NCBI RefSeq-style accession with a version, like GCA_003177105.1, causes the execution of the tool to stop.  This patch sanitizes those labels, replacing all non-allowed characters to underscores.

## About tests

I was unable to run complete tests on my M4 Mac, but tests with `--no_dependency_resolution` did pass after I removed some strict version pins in the macro. Specifically, The gmp=6.2.1 pin causes a hard conda conflict with current nettle (which requires gmp>=6.3.0). For bowtie2 and samtools, I was unable to install the pinned version but with the pin removed, it got newer versions that did install. Those changes to the macro are included in this PR. To the test file, I added a test that provides an versioned accession to test these changes, otherwise reusing the existing test data.


tools/hicup/test-data/digest_file_dotted_genome.txt
bernt-matthias, line 1: Can you remove lines, maybe just a few are sufficient for the test?

bernt-matthias (commented):
Ah, I thought `--genome` is the cached index file. 

Values from a `text` parameter can be sanitized much more easily: 

https://docs.galaxyproject.org/en/master/dev/schema.html#tool-inputs-param-sanitizer

you do not need to do this in the command section.

scottcain:
> Ah, I thought `--genome` is the cached index file.
> 
> Values from a `text` parameter can be sanitized much more easily:
> 
> https://docs.galaxyproject.org/en/master/dev/schema.html#tool-inputs-param-sanitizer
> 
> you do not need to do this in the command section.

Yeah, that's what I thought too until I dug in a little more deeply. I was unaware of the sanitizer thing; I'll update with that.

tools/hicup/test-data/digest_file_dotted_genome.txt
scottcain: Is there a way to say "the top ten lines of the result should look like this"?

bernt-matthias: You can use `compare="contains"` then the 10 lines must be included .. but not necessarily at the top. Otherwise have a look at https://docs.galaxyproject.org/en/master/dev/schema.html#tool-tests-test-output-assert-contents

scottcain: Actually, I changed to assert that the file contains the sanitized label and not the original label and removed the redundant output file.

tools/hicup/hicup_macros.xml
```diff
@@ -1,15 +1,15 @@
...
-    <token name="@WRAPPER_VERSION@">@VERSION@+galaxy1</token>
+    <token name="@WRAPPER_VERSION@">@VERSION@+galaxy2</token>
     <token name="@PROFILE@">22.01</token>
     <xml name="requirements_hicup">
         <requirements>
             <requirement type="package" version="@VERSION@">hicup</requirement>
-            <requirement type="package" version="2.4.5">bowtie2</requirement>
-            <requirement type="package" version="1.16.1">samtools</requirement>
+            <requirement type="package">bowtie2</requirement>
+            <requirement type="package">samtools</requirement>
```
bernt-matthias: You should keep the versions (if you want you can update to the latest)

scottcain: samtools is really problematic for testing of this package: HiCUP has a dependence on the BioPerl  samtools module, but that module requires samtools to be between >=0.1.19,<0.2.0a. This conflicts with the pin of this package of 1.16.1. Meaning tests could never pass (at least not in a normal testing environment).  Should I put it back to 1.16.1 and live with failing tests on my machine?

bgruening: bowtie2 is defined here: https://github.com/bioconda/bioconda-recipes/blob/master/recipes/hicup/meta.yaml so it can be removed I think.

bioperl as a dependency on bioperl-samtools https://github.com/bioconda/bioconda-recipes/blob/master/recipes/perl-bio-samtools/meta.yaml

And that one has also samtools defined and pinned, so I think you can drop this here also completely.

scottcain: Done, I dropped explicit dependencies on bowtie and samtools and tests still pass with `--no_dependency_resolution` which is what I would expect, but good that it does

scottcain:
@bernt-matthias I don't understand the nature of the two failing tests (check workflow success and combine chunked test results); one has so little output as to be useless to me, the other says to check a file that I don't know how to find.

bgruening:
@scottcain please run `planemo test --biocontainers ./tools/foo/var` ... thats all you need.

The failing tests are due to changes in the diff of the result files. You can see them here: https://github.com/bgruening/galaxytools/actions/runs/26597546236?pr=1880

PR #1886, opened 2026-06-03
Fix TypeError when getOverlay() returns None with no particles

dianichj:
`imagej2_analyze_particles_binary_jython_script.py` crashes with:
```
TypeError: 'NoneType' object is not iterable
```
at the line:
```python
for i, roi in enumerate(ov):
```
whenever `export_rois=true` and **zero particles** match the analysis criteria
(e.g. `size=50-Infinity` on an image whose binary mask has no blob ≥ 50 px²).

Root cause: `ImagePlus.getOverlay()` returns `None` (not an empty list) when
Analyze Particles produces no results. The script then tries to iterate over
`None`, which raises `TypeError`.

This is the real failure underlying Step 14 (`Generate ROIs of stained region`)
in the IWC histological staining workflow PR
[[galaxyproject/iwc#1206](https://github.com/galaxyproject/iwc/pull/1206)](https://github.com/galaxyproject/iwc/pull/1206).

## Fix

One-line change in `imagej2_analyze_particles_binary_jython_script.py`:
```python
# Before
ov = input_image_plus_copy.getOverlay()

# After
ov = input_image_plus_copy.getOverlay() or []
```

## Changes

- `imagej2_analyze_particles_binary_jython_script.py`: guard `getOverlay()` against `None`
- `imagej2_analyze_particles_binary.xml`: add regression test for zero-particles case
- `test-data/analyze_particles_rois_empty.tabular`: expected header-only output for the new test

## Test case added

```xml
<test expect_num_outputs="2">
    <param name="input" value="blobs.gif"/>
    <param name="size" value="100000-Infinity"/>
    <param name="show" value="Nothing"/>
    <param name="export_rois" value="true"/>
    <output name="results" file="analyze_particles_nothing.tabular"/>
    <output name="rois" file="analyze_particles_rois_empty.tabular"/>
</test>
```

`analyze_particles_rois_empty.tabular` contains only the header row:
```
shape	points	label	t	z
```
@mvdbeek


tools/image_processing/imagej2/imagej2_analyze_particles_binary.xml
mvdbeek, line 1: Can you bump up the version suffix macro ?

dianichj: I was wondering if I was missing this too, thanks!

bgruening:
> AssertionError: Test input file (confocal-series-first-channel.tif) cannot be found.

and it seems bio.tools has not entry for `imagej`

mvdbeek:
There is one now (https://bio.tools/imagej, maybe flaky network previously), i fixed the path to the test file.

bgruening (approved):

bgruening:
Will try to deploy it in the next hours.

mvdbeek:
There's no hurry, we're building on 26.1 on test. We'll update main hopefully this week and then we move over to .eu next week ?

mvdbeek:
The TS deploy however didn't work, that would be good to fix (maybe that was the deploy you meant ?) https://github.com/bgruening/galaxytools/actions/runs/27191718092/job/80275468017

PR #1892, opened 2026-06-11
Coco visualizer h264

TuturBaba:
I pushed a test on Galaxy Europe and see that the output video could not be played by media player, as the codec was not supported. So, following the same approach as SAM3, I add a re-encoding step.

### DIFF
--- a/tools/image_processing/yolo/coco_annotation_visualizer.py
+++ b/tools/image_processing/yolo/coco_annotation_visualizer.py
@@ -1,5 +1,6 @@
 import argparse
 import json
+import subprocess
 from pathlib import Path
 from typing import Dict, List, Optional, Tuple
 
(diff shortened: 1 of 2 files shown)


bgruening:
What format is not supported by Galaxy? or is it not supported by the browser? We can work at improving the Visualisation instead. 

I could be wrong, but maybe its https://github.com/galaxyproject/galaxy-visualizations/tree/main/packages/plyr

TuturBaba:
> What format is not supported by Galaxy? or is it not supported by the browser? We can work at improving the Visualisation instead.
> 
> I could be wrong, but maybe its https://github.com/galaxyproject/galaxy-visualizations/tree/main/packages/plyr

I think I have the answer. With help from Claude and other sources: "MP4V (MPEG-4 Part 2) is not supported by most browsers." On top of that, OpenCV on Linux doesn't include H.264 encoding due to licensing restrictions.

The solution is to re-encode using FFmpeg (locally). 
I could try switching to `cv2.VideoWriter_fourcc(*'H264')` directly, but it's not guaranteed to work in Galaxy. Keeping the FFmpeg re-encoding step is the safer and more portable approach.

Either way, the root cause is a browser compatibility issue, not a Galaxy-specific one.

bgruening:
Ok. Let me know when the PR is finished. One test is failing

TuturBaba:
> Ok. Let me know when the PR is finished. One test is failing

 @bgruening Sorry for the late ping! The PR is ready, we just need to add skip version check

bgruening (approved):

TuturBaba:
Thank you

Issue #1921, opened 2026-07-17
llm_hub: no way to attribute requests to a Galaxy user (blocks per-user quotas on a shared LiteLLM proxy)


gkr0110:
Hi! We're running LLM Hub on [usegalaxy.ch](https://stage-usegalaxy.ch) against a self-hosted LiteLLM proxy in front of the [CSCS](https://docs.cscs.ch/services/inference/) inference service. Two things made that harder than expected — happy to send PRs for either/both if you agree with the direction.

## 1. Requests can't be attributed to a Galaxy user

`llm_hub.py` builds the request as:

```python
api_params = {"model": model, "messages": messages, "stream": True}
if temperature is not None:
    api_params["temperature"] = temperature
```

There's no `user` field, and the tool never receives the Galaxy user id (the wrapper passes `context_files`, `prompt`, `model_id`, `model_type`, `temperature`, `provider`).

**Why it matters:** a Galaxy instance authenticates to the proxy with **one shared key**, so from LiteLLM's perspective every request looks identical. That means:

- spend/usage can only be tracked in aggregate, not per user;
- LiteLLM's [end-user budgets](https://docs.litellm.ai/docs/proxy/users) (`/customer/new`, `max_end_user_budget_id`) can't be used at all — they key off the OpenAI `user` field;
- there's no way to apply a fair-use quota, so one user can exhaust a shared instance-wide rate limit for everyone.

OpenAI's `user` field is designed for exactly this, and LiteLLM auto-creates an end-user record when it's present.

**Suggested fix:** pass `$__user_id__` from the wrapper and forward it as `user` when set. The opaque numeric id avoids putting an email in a field that gets forwarded upstream to the provider. Anonymous sessions would simply omit it (or share one bucket).

Happy to raise a PR — I have this working locally as an out-of-tree patch.

## 2. Credentials can only come from a config file, never the environment

```python
litellm_config_file = os.environ.get("LITELLM_CONFIG_FILE")
if not litellm_config_file:
    sys.exit("LITELLM_CONFIG_FILE environment variable is not set.")
...
litellm_api_key = source.get("LITELLM_API_KEY")
if not litellm_api_key:
    sys.exit(
        "LiteLLM API key is not configured! Please set LITELLM_API_KEY environment variable."
    )
```

The error message says *"Please set LITELLM_API_KEY environment variable"*, but the value is only ever read from the **config file** — setting that env var has no effect. That's a bit of a trap (it cost us a while to work out).

[...]

PR #1922, opened 2026-07-17
llm_hub: attribute requests to the Galaxy user

gkr0110:
Fixes the first half of #1921.

## Problem

A Galaxy instance authenticates to its LiteLLM proxy with a **single shared key**, so every request looks identical to the proxy. Today `llm_hub.py` builds:

```python
api_params = {"model": model, "messages": messages, "stream": True}
```

— no `user` field, and the wrapper never passes the Galaxy user id. Consequently:

- usage/spend is trackable only in aggregate, not per user;
- LiteLLM's [end-user budgets](https://docs.litellm.ai/docs/proxy/users) (`/customer/new`, `max_end_user_budget_id`) can't be used at all — they key off the OpenAI `user` field;
- a shared instance-wide rate limit can be exhausted by one user, for everyone.

## Change

Pass `$__user_id__` from the wrapper and forward it as the standard OpenAI `user` field when set.

```diff
  '$model.fields.provider'
+ '$__user_id__'
```
```diff
  api_params = {"model": model, "messages": messages, "stream": True}
+ if galaxy_user_id:
+     api_params["user"] = galaxy_user_id
```

Two deliberate details:

- **`$__user_id__`, not `$__user_email__`.** This value is forwarded to the upstream provider, so the opaque numeric id keeps personal data out of it. Admins who want a different identifier can map it proxy-side.
- **Anonymous sessions.** Cheetah renders the id as the literal string `"None"`, which is normalised to empty, so `user` is simply omitted. Requests are byte-for-byte unchanged for anyone not doing per-user metering.

`VERSION_SUFFIX` bumped to 3.

## Testing

Running as an out-of-tree patch on [stage-usegalaxy.ch](https://stage-usegalaxy.ch) against a self-hosted LiteLLM proxy in front of the [CSCS](https://docs.cscs.ch/services/inference/) inference service. With `user` present, LiteLLM auto-creates the end-user record on first call and `max_end_user_budget_id` then applies a default TPM/RPM budget — which is what makes per-user fair-use possible at all.

`py_compile` and XML parse both clean. No test-data changes: the field only affects the outbound request, and the existing tests don't exercise a proxy that reports attribution. Happy to add coverage or adjust the identifier choice if you'd prefer something else.


bgruening:
Thanks! When per-user quotas are needed this makes a lot of sense. I have one very tiny, unlikely concern. If multiple Galaxy instances are sharing one LiteLLM, we do this at EU, there is the unlikely that two users at two Galaxy instances have the same ID. Would it be possible to hash the ID together with a URL or some static unique Galaxy key and make the user-ID more unique? Maybe hashing the ID + email - in the hope that GDPR-compliant?

gkr0110:
Good point, thanks — done in the latest commit.

I went with the URL rather than the email, to keep it PII-free: the id is now sent as `sha256("$__galaxy_url__|$__user_id__")`. `$__galaxy_url__` is the instance's `galaxy_infrastructure_url`, so two instances sharing one proxy get different hashes and can't collide, and neither the URL nor the numeric id leaves for the provider — no personal data involved, which sidesteps the GDPR question your email idea raised.

One small addition while I was there: anonymous users (no `$__user_id__`) fall back to a per-instance shared `"anonymous"` bucket — `sha256("<url>|anonymous")` — so unauthenticated traffic is still metered as one pool per instance rather than unbounded. Easy to drop if you'd rather keep the PR strictly to logged-in users.

Only caveat: an instance that leaves `galaxy_infrastructure_url` at its default won't contribute a distinguishing prefix, but that's no worse than today and only affects unconfigured instances.

arash77:
Thanks @gkr0110, the approach is sound, namespacing the user id by the instance URL and hashing it is a clean, privacy-preserving per-user key for the proxy.

One detail caught my eye: both `== "None"` guards look like dead code. Galaxy never emits `"None"` for either variable:
- `$__user_id__` is the literal `"Anonymous"` for logged-out users and `f"{user.id}"` for logged-in ones (`User.user_template_environment`, doctest pins `'Anonymous'`) — never `"None"` or empty.
- `$__galaxy_url__` is `galaxy_infrastructure_url`, declared `type: str` (default `http://localhost:8080`) in the config schema, always a URL string, never `"None"`.

gkr0110:
Thanks for catching this @arash77 — you're right, good eyes. Fixed in 6c14c18: the `galaxy_user_id` guard now checks for the literal `"Anonymous"` (what Galaxy actually emits for logged-out sessions) instead of `"None"`, and the guard for `galaxy_url` was dead code, so I removed it.

I also double-checked the rest of the tool for similar "None" string guards on Cheetah-templated values — this was the only pair, so nothing else to fix.

gkr0110:
@bgruening With the anonymous-session guard fixed per @arash77 's review, is this good to merge? We'd also love to know if it could make it into your upcoming release — we're looking to try the latest version out on Galaxy (https://stage-usegalaxy.ch) and would appreciate having this land there.

bgruening (approved):

bgruening:
Thanks @gkr0110

PR #1927, opened 2026-07-24
add ketos_ocr suite (the model training part of kraken-ocr)

IvoLeist:
This PR starts the addition of the ketos_ocr suite.
Ketos is the model training part of [kraken](https://github.com/mittagessen/kraken), a turn-key OCR system optimized for historical and non-Latin script material.

Here the [`ketos compile`](https://kraken.re/main/user_guide/training_recognition.html#data-preparation) command is wrapped. Future PRs (WIP) will add other ketos functionalities such as `train`.
The idea of wrapping ketos' commands individually is to allow the use of each ketos command within a galaxy workflow.

FOR CONTRIBUTOR:

Notes for reviewers:
- I expect the linting and tests to fail with "Unknown datatype `page.xml` and `arrow`"
 since I assume that the GitHub CI is still at Galaxy 26.0.1. These errors should disappear once version 26.1 is in use.


IvoLeist:
Currently failing tests here (but passing locally):

ketos_train: 1, 29, and 30 

All three return:
```
/usr/local/lib/python3.13/site-packages/torch/_inductor/runtime/runtime_utils.py

22 def default_cache_dir() -> str:
> 23     sanitized_username = re.sub(r'[\\/:*?"<>|]', "_", getpass.getuser())
24     return os.path.join(
25         tempfile.gettempdir() if not is_fbcode() else "/var/tmp",
26         "torchinductor_" + sanitized_username,

/usr/local/lib/python3.13/getpass.py:175 in getuser

172     import pwd
173     return pwd.getpwuid(os.getuid())[0]
174 except (ImportError, KeyError) as e:
> 175     raise OSError('No username set in the environment') from e

OSError: No username set in the environment
```

IvoLeist:
After setting the USER it failed with:

> InductorError: InvalidCxxCompiler: No working C++ compiler found in torch._inductor.config.cpp.cxx: (None, 'g++')
> 
> Set TORCHDYNAMO_VERBOSE=1 for the internal stack trace (please do this especially if you're reporting a bug to PyTorch). For even more developer context, set 
> TORCH_LOGS="+dynamo"

IvoLeist:
Previously failing tests were fixed by adding the following:

```
 ## Disable PyTorch compilation by default; set TORCHDYNAMO_DISABLE=0 to enable it.
 export TORCHDYNAMO_DISABLE="\${TORCHDYNAMO_DISABLE:-1}"
```

tools/ketos_ocr/ketos_compile.xml
```diff
@@ -0,0 +1,219 @@
+<tool id="ketos_compile" name="Ketos Compile" version="@TOOL_VERSION@+galaxy@VERSION_SUFFIX@" profile="@PROFILE@">
+    <description>precompile OCR training data into an Arrow dataset</description>
+    <macros>
+        <import>macros.xml</import>
+        <import>macros_for_testing.xml</import>
+    </macros>
+    <expand macro="requirements">
+        <requirement type="package" version="4.1.2">yq</requirement>
```
bgruening, line 8: I'm super confused with this package. @pavanvidem said this is now called go-yq?

tools/ketos_ocr/stage_ground_truth_xml.py
bgruening, line 1: is this still used? you are using yq for it correct?

tools/ketos_ocr/ketos_train.xml
```diff
@@ -0,0 +1,1929 @@
...
+
+                @REJECT_MULTIPAGE_TIFF_BASED_ON_FILE_METADATA@
+
+                #if $training_data.format_type == 'path'
+                    ln -s '$image' 'ground_truth_${i}.${image.ext}' &&
+                    ln -s '$annotation' 'ground_truth_${i}.gt.txt' &&
+                #else
+                    mkdir -p 'ground_truth_${i}' &&
+                    ln -s '$image' 'ground_truth_${i}/input.${image.ext}' &&
+                    python '$__tool_directory__/stage_ground_truth_xml.py' '$annotation' 'ground_truth_${i}/annotation.xml' 'input.${image.ext}' &&
```
bgruening: yq maybe?

tools/ketos_ocr/ketos_compile.xml
```diff
@@ -0,0 +1,219 @@
...
+                <option value="page">PAGE XML</option>
+                <option value="path">Line image/transcription pairs</option>
+            </param>
+            <expand macro="compile_training_data_inputs"/>
+        </conditional>
+        <param argument="--force-type" type="select" optional="true" label="Force dataset type" help="Forces the dataset type to a specific value. Can be used to 'convert' a line strip-type collection to a baseline-style dataset, e.g. to disable centerline normalization.">
+            <option value="baseline">Baselines</option>
+            <option value="bbox">Bounding boxes</option>
+        </param>
+        <param name="empty_lines" type="select" label="Empty text lines" help="Whether to keep or skip empty text lines. Text-less datasets are useful for unsupervised pretraining but loading datasets with many empty lines for recognition training is inefficient">
```
bgruening: that could also be a boolean I guess ... matter of preference I guess

tools/ketos_ocr/ketos_train.xml
```diff
@@ -0,0 +1,1929 @@
...
+                        #end if
+                    #end for
+                #end if
+            #end if
+        ]]></configfile>
+    </configfiles>
+    <inputs>
+        <conditional name="training_data">
+            <param argument="--format-type" type="select" label="Ground-truth data format" help="Format used for both training data and potential explicit validation data.">
+                <option value="binary" selected="true">Compiled binary dataset</option>
```
bgruening: what is that, do we need to offer this?

bgruening (approved):

tools/ketos_ocr/ketos_compile.xml
IvoLeist: I had a look into it:
[go-yq](https://anaconda.org/channels/conda-forge/packages/go-yq/overview) points to https://github.com/mikefarah/yq = uses jq like syntax
[yq](https://anaconda.org/channels/conda-forge/packages/yq/overview)  points to https://github.com/kislyuk/yq = `jq` wrapper for YAML, XML, TOML documents

So I would continue to stick to the "original" `yq` and not its go re-implementation

tools/ketos_ocr/ketos_train.xml
bgruening: @IvoLeist is this still needed?

IvoLeist:
@bgruening thanks for your quick first comments :)

Here are my changes :rocket: 
- replace stage_ground_truth helper with yq
- skip empty lines is now a boolean
- training data format type "binary" is now more self explanatory -> Compiled Apache Arrow dataset

tools/ketos_ocr/ketos_train.xml
IvoLeist: This is to upload precompiled training-data in  Apache Arrow format as produced by `ketos compile`. 
It packages the line images and transcriptions for faster loading. Thus, yes we should offer this since it is should be the preferred input format for the training data. 

But good catch ! I made it now more self-explanatory by calling the option "Compiled Apache Arrow dataset"

tools/ketos_ocr/stage_ground_truth_xml.py
IvoLeist: good catch :) I had the yq solution already in ketos_compile but then forgot re-use it in ketos_train :sweat_smile:

bgruening:
Merci!

PR #1928, opened 2026-07-24
Hello gxy-tool-bot

d-callan:
@bgruening please see bot config. assumes GXY_TOOL_BOT_API_KEY is set already and holds the cz token. also, we need labels before this will work, i can add them and comment back here once they exist. 

i set it to deepseek-thinking w qwen fallback for now. can change it to try different things as needed. but thats what ive been doing the last 24 hrs or so in my iuc fork though and so far at least while its slower it times out less and never fails entirely for that. glm use looks ok today, but was really heavy yest so i tried something else. honestly i think if we get the harness good enough (eventually lol) the model will matter less

other things to note.. can change timeout times, temperature, how many iterations to let it go before it has to make a commit etc in the bot config yml also. and if you want to restrict to specific maintainers, can do that in the yml to w:
```
allowed_maintainers:
  - your-github-handle
```

if any of those buttons and knobs youd rather not be in the bot config yml where others can see and pr against etc, let me know. i figured favor transparency s much as possible until i was given a reason to not.

its likely as i keep fiddling w things the workflows might get modified or cleaned up some, but i think where it is now is ok enough to try to get broader feedback on and see .

### DIFF
--- a/.github/ISSUE_TEMPLATE/tool-request.yml
+++ b/.github/ISSUE_TEMPLATE/tool-request.yml
@@ -1,0 +1,45 @@
+name: Tool Request
+description: Request a new Galaxy tool wrapper to be generated
+labels: ["tool-request"]
+body:
+  - type: markdown
+    attributes:
+      value: |
+        Fill out this form to request a new Galaxy tool wrapper.
+        The bot will research the tool and post a plan as a comment.
+        A maintainer will then review and approve it for generation.
+  - type: input
+    id: tool-name
+    attributes:
+      label: Tool name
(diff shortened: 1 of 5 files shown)


d-callan:
labels exist. 

also worth noting i have this installing from pypi so its stable while i play, ill keep my iuc fork using latest

d-callan:
oh. also some of the reason its slower now is i let it run planemo test and lint itself directly to try to reduce the number of feedback iterations

bgruening (approved):
It's great as it is, let's ship it, test it and get some broader feedback. 

Thanks a lot @d-callan !

Issue #1924, opened 2026-07-24
Port Galaxy Aries tools


paulzierep:
We have a support request to deploy Galaxy Aries tools to EU. Therefore, we need to port them to a reliable source i.e. here, or IUC if the maintainers are willing to update to IUC standards.
First, lets find the wrappers:

- [x] LisSero: https://toolshed.g2.bx.psu.edu/view/estrain/lissero/6c8e5bfff8a8
- [x] Confindr: https://github.com/OLC-Bioinformatics/ConFindr?utm_source=chatgpt.com
- [x] ChewBBACA Allele Call: https://toolshed.g2.bx.psu.edu/repositories/878ed8350482b853 
- [ ] Chewtree: https://github.com/B-UMMI/chewTree?utm_source=chatgpt.com
- [ ] Grapetree: https://toolshed.g2.bx.psu.edu/repositories/c03f5600b8c00af1

```
LisSero, ChewBBACA Allele Call, and GrapeTree are Galaxy ToolShed wrappers and should be installable from the ToolShed.
ConFindr and chewTree are indeed standalone GitHub projects and, I’m going to ask for public wrappers to the people from GALAXY ARIES:.
For GrapeTree, I am referring to the standard Galaxy tool for visualising cgMLST/MST trees, not the Interactive Tool version.
```

paulzierep:
Questions: 
* LisSero wrapper wraps variant4b.py not LisSero is this intended ?
* https://usegalaxy.eu/?tool_id=toolshed.g2.bx.psu.edu/repos/iuc/chewbbaca_allelecall/chewbbaca_allelecall/3.3.10+galaxy2 is in IUC and on EU.
* https://toolshed.g2.bx.psu.edu/view/estrain/confindr does not exist anymore, cannot find the wrapper
* Chewtree link does not exist

paulzierep:
Grapetree updated with working test: https://github.com/bgruening/galaxytools/pull/1925

paulzierep:
LisSero with only variant4b.py script: https://github.com/bgruening/galaxytools/pull/1926

paulzierep:
We got Chewtree and Confindr locally, but needs licence information from the devs before we can publish here.

mtangaro:
For Chewbbca: https://github.com/aknijn/chewbbaca-galaxy
Shigatoxin is not in the list but is in AIRES: https://github.com/aknijn/shigatoxin-galaxy

paulzierep:
[...]

PR #1936, opened 2026-07-27
Add ecolishigatoxintyper

paulzierep:
FOR CONTRIBUTOR:


tools/ecolishigatoxintyper/scripts/duk
bgruening, line 1: why are we shipping binaries and tools in this PR?

paulzierep: I ported this tool from https://github.com/aknijn/shigatoxin-galaxy as support for the Aries tools (https://github.com/bgruening/galaxytools/issues/1924) - I have no idea about this tool myself, and did not check first what is in the scripts folder. 
The correct way would probably be to make a Bioconda env first. But I have no idea about perl and I am not even sure what the duk binary is, so I can only try :)

RZ9082: This looks more like a whole workflow with a bunch of tools rather than a single tool. I’m going to close this for now, but if you think otherwise, please feel free to reopen it.

PR #1943, opened 2026-07-30
llm_hub: surface context-window overflow as an actionable error instead of a traceback

arash77:
## Summary

Replaces the original approach (a `max_context` data-table column + local `tiktoken` estimate) with graceful handling of the rejection the proxy **already** returns.

The premise of the first version was that an oversized input is silently truncated. Testing against both live providers showed that is not what happens — the request is rejected up front, before any generation, with the exact numbers:

```
litellm.BadRequestError: Hosted_vllmException -
  Input length (270080) exceeds model's maximum context length (128000).
  Received Model Group=gpt-oss-120b-llmlb
```

Nothing is truncated and no tokens are generated. The real defect was that `llm_hub.py` did not catch this, so the job died with a raw Python traceback instead of a usable message.

## What changed

- **`llm_hub.py`**: catch `BadRequestError` as non-retryable and exit with an actionable message. Context-window overflows get a tailored hint (split the input, or pick a model with a larger window); other 400s report the upstream error verbatim.
- **`llm_hub.py`**: add a trailing `except APIError` clause. The OpenAI SDK maps only 400/401/403/404/409/422/429/5xx to named subclasses — everything else (HTTP 413 in particular) falls through to a bare `APIStatusError`, and `APIResponseValidationError` is an `APIError` that is not an `APIStatusError` at all. All of these previously escaped as tracebacks.
- **`llm_hub.xml`**: version suffix `0` → `1`.

## What was dropped, and why

- **The `max_context` column and the `genai_models` → `genai_models_extended` rename.** @bgruening: this addresses your objection directly — there is no data-table change at all now.
- **The `tiktoken` dependency.** The server's numbers are authoritative; a local `cl100k_base` estimate is both redundant and inaccurate (measured 25–34% under on JSON/HTML, 62% over on mixed-script text).

Beyond the table-rename problem, the column had no source to populate it from: the proxies' virtual keys are scoped to `llm_api_routes`, and `max_input_tokens` is unset on every model on both providers. Any value would have to be hand-derived per model, and re-derived whenever a provider changes its lineup.

## Retry behaviour is unchanged

Verified against the pinned SDK that `APITimeoutError`, `APIConnectionError`, `RateLimitError` and `InternalServerError` are all `APIError` subclasses caught by the earlier clauses, so timeouts, connection failures, 429 and 5xx still retry exactly as before. Only genuinely non-retryable statuses reach the new clause. All ten error types were checked against the real clause order.

## Testing

- `planemo lint` — no warnings or errors
- `planemo test` — `All 1 test(s) executed passed. llm_hub (Test #1): passed`
- End-to-end against both providers: oversized input now exits 1 with the actionable message and no traceback; a normal request still succeeds and writes `output.md`
- A real non-overflow 400 (`"g02v09-glm-5.2 is not a multimodal model"`) correctly falls through to the generic message rather than being mislabelled as an input-size problem

## Backward compatibility

No data table, `.loc`, or `tool_data_table_conf.xml` changes. Existing deployments need no action.

---

FOR CONTRIBUTOR:

### DIFF
--- a/tools/llm_hub/README.rts
+++ b/tools/llm_hub/README.rts
@@ -6,7 +6,7 @@ This Galaxy tool provides access to various Large Language Models (LLMs) through
 
 ### Multi-Server Setup
 
-Galaxy administrators can configure multiple LiteLLM servers using a YAML configuration file. Each server is identified by the provider field (4th column) in `genai_models.loc`.
+Galaxy administrators can configure multiple LiteLLM servers using a YAML configuration file. Each server is identified by the provider field (4th column) in `genai_models_extended.loc`.
 
 #### Configuration Steps:
 
@@ -30,7 +30,7 @@ Galaxy administrators can configure multiple LiteLLM servers using a YAML config
 
 3. Set the `LITELLM_CONFIG_FILE` environment variable to point to your configuration file.
 
-4. Configure your models in `genai_models.loc`. See `tool-data/genai_models.loc.sample` for the complete format specification and examples.
+4. Configure your models in `genai_models_extended.loc`. See `tool-data/genai_models_extended.loc.sample` for the complete format specification and examples.
 
 ### Backward Compatibility
 
@@ -41,4 +41,4 @@ LITELLM_BASE_URL: https://your-litellm-instance.com/
 LITELLM_API_KEY: sk-your-api-key-here
 ```
 
-In this case, all models will use the same server configuration regardless of the provider field in `genai_models.loc`.
+In this case, all models will use the same server configuration regardless of the provider field in `genai_models_extended.loc`.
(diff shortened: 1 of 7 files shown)


bgruening:
> Existing 6-column genai_models.loc deployments keep working provided admins update tool_data_table_conf.xml to the 7-column declaration and pad each row to 7 fields (empty max_context is fine). The existing tool test (no API key configured) is unchanged and still passes.

This is all automatics, we can not do that :(
We would need to change the table name again.

arash77:
> This is all automatics, we can not do that :( We would need to change the table name again.

Is it worth doing, then? Our providers could actually supply this via an API endpoint like `/model/info` (which carries `max_input_tokens`), but we would have to ask them to add it.

bgruening:
Is CZ supporting it and UFR not? I would ask UFR to add it.

arash77:
> Is CZ supporting it and UFR not? I would ask UFR to add it.

Neither supports it right now.

bgruening:
Interesting, ok lets ask them and see how complicated that is.

arash77:
ok RZ looked into it and it seems customizing the api pass through routes is a premium freature in LiteLLM!

PR #1947, opened 2026-08-03
Add langchain_text_splitters

IvoLeist:
This PR starts the addition of [langchain-text-splitters](https://docs.langchain.com/oss/python/integrations/splitters),  a script collection for breaking large text into smaller chunks for downstream LLM tools. In the GalaxyEU context these could be e.g. [LLM Hub](https://usegalaxy.eu/?tool_id=toolshed.g2.bx.psu.edu%2Frepos%2Fbgruening%2Fllm_hub%2Fllm_hub%2F2.41.1%2Bgalaxy3&version=latest) or [RAG Retriever](https://usegalaxy.eu/?tool_id=toolshed.g2.bx.psu.edu%2Frepos%2Fbgruening%2Frag_retriever%2Frag_retriever%2F1.0.0%2Bgalaxy1&version=latest).

**Leveraged Methods of `langchain-text-splitters` for this initial PR**

- **Character-based splitting:** Splits text at character-based separators. The target chunk size can be measured in characters or tokens.
    - [RecursiveCharacterTextSplitter](https://docs.langchain.com/oss/python/integrations/splitters/recursive_text_splitter): Recommended for most use cases as it attempts to preserve larger text units by trying an ordered list of separators, from coarser boundaries such as paragraph breaks to finer boundaries such as line breaks, or spaces, down to individual characters.
    - [CharacterTextSplitter](https://docs.langchain.com/oss/python/integrations/splitters/character_text_splitter): Splits text using one defined separator or character sequence, such as a paragraph break, line break, space, or custom separator.
- **Token-based splitting:** The [TokenTextSplitter](https://docs.langchain.com/oss/python/integrations/splitters/split_by_token) ignores any potential text structure and simply creates chunks of a specified token length.
- **Sentence-based splitting:** Detects sentence boundaries first and then fills the chunks with whole sentences, so a chunk never ends in the middle of a sentence. Pick this for prose. A single sentence that is already longer than the target chunk size becomes an oversized chunk of its own, because the sentence is never cut.
    - `NLTKTextSplitter`: Uses NLTK's Punkt sentence tokenizer, which is trained per language and therefore handles language-specific abbreviations such as `Dr.` in English or `z.B.` in German. Select the language of the input.
    - `SpacyTextSplitter`: Uses spaCy. The rule-based sentencizer is fast and needs no model, but it applies English tokenization rules. The `en_core_web_sm` model is more accurate on English text, and is roughly five times slower and needs about ten times more memory.

**Library for Token Counting**

OpenAI's fast Byte-Pair Encoding (BPE) tokenizer [tiktoken](https://github.com/openai/tiktoken) library is used to determine the number of tokens in a chunk for this proof of concept. *`langchain-text-splitters` is also allowing Hugging Face tokenizer to count length, consequently this is on the roadmap, but not (yet) part of this PR.

FOR CONTRIBUTOR:


tools/langchain_text_splitters/split_text.py
bgruening, line 1: there is nothing like that already available?

tools/langchain_text_splitters/macros_for_testing.xml
bgruening, line 1: I would move them into the main macro file.

tools/langchain_text_splitters/macros.xml
```diff
@@ -0,0 +1,112 @@
...
+    <token name="@PROFILE@">25.1</token>
+    <xml name="requirements">
+        <requirements>
+            <requirement type="package" version="@TOOL_VERSION@">langchain-text-splitters</requirement>
+            <requirement type="package" version="0.13.0">tiktoken</requirement>
+        </requirements>
+    </xml>
+    <xml name="creators">
+        <creator>
+            <person givenName="Ivo Christopher" familyName="Leist" email="leist@informatik.uni-freiburg.de" identifier="https://orcid.org/0000-0002-9772-443X"/>
```
bgruening, line 28: mabye also ack your funding, de.NBI or AI4SOCIAL

tools/langchain_text_splitters/test-data/.gitignore
bgruening, line 1: what is that?

bgruening (commented):
Remove the dev_utils?

IvoLeist:
@bgruening thanks for the quick first review :)

Here are the changes:
- Remove the folder dev_utils and its content
- Remove .gitignore inside the test-data folder
- Merge macros_for_testing.xml into macros.xml
- Remove macro for setting the TIKTOKEN_CACHE_DIR env var
  -> I guess we better set this env variable via tpv/tools.yml as we did with kranken-ocr (https://github.com/usegalaxy-eu/infrastructure-playbook/blob/master/files/galaxy/tpv/tools.yml#L897)
- Add "AI4Social+" to the creators
- Address flake8 warnings

tools/langchain_text_splitters/test-data/.gitignore
IvoLeist: The file I did not wanted to accidentally commit and thus had this .gitignore inside test-data was Daniela's use case I was describing in Zammad. But thanks good catch :) I removed the .gitignore and moved the referenced file somewhere else.

tools/langchain_text_splitters/split_text.py
IvoLeist: The closest existing Galaxy tool I found is [Split file to dataset collection](https://usegalaxy.eu/?tool_id=toolshed.g2.bx.psu.edu%2Frepos%2Fbgruening%2Fsplit_file_to_collection%2Fsplit_file_to_collection%2F0.5.2&version=latest) which can also split generic text files via a separator ([link to python code](https://github.com/bgruening/galaxytools/blob/master/tools/text_processing/split_file_to_collection/split_file_to_collection.py#L140)).

However, in the _Split file to dataset collection tool_ afaik you can only define the target chunksize with "Number of records per output file" whereby the splitting is based on one separator.

The `langchain_text_splitters` in contrast allow more "sophisticated" text chunking by e.g. recursively going through the defined separators to find the best fitting separator(s) so larger text units can be preserved, see [https://docs.langchain.com/oss/python/integrations/splitters/recursive_text_splitter](https://docs.langchain.com/oss/python/integrations/splitters/recursive_text_splitter). On top, target chunk sizes can be defined on the character or the token count level. And finally it is also possible to allow overlapping of chunks content, which is especially useful for retrieval-augmented generation (RAG) workflows as it can help to preserve context across chunk boundaries.

bgruening: Not a Galaxy tool, but any tool out their seems like a usecase the humanities did already 100 times no?

IvoLeist: Yes, there are more "old school" tools for text splitting/tokenization, such as [NLTK](https://www.nltk.org/api/nltk.tokenize.html#module-nltk.tokenize) or more modern [spaCy](https://spacy.io/usage/linguistic-features#tokenization). However, as I understood in these tools their tokenizers would not split inside a word.

Thus,  the PR as it is at the moment is clearly tailored to support downstream LLM workflows. Nevertheless, we could: 

- (a) Wrap spaCy/NLTK each in its own Galaxytool
- (b) Add to this PR the respective `langchain_text_splitters`' wrappers:
  - [SpacyTextSplitter](https://reference.langchain.com/python/langchain-text-splitters/spacy/SpacyTextSplitter) 
  -  [NLTKTextSplitter](https://reference.langchain.com/python/langchain-text-splitters/nltk/NLTKTextSplitter) 

?

anuprulez: I think (b) is more suitable and extensible for different types of data formats such as JSON, XML etc. We can extend (b) for more data formats.

arash77:
Thank you @IvoLeist, nice tool! A few issues to fix before merge, found by AI review (verified by running the script with the pinned deps).

## Must fix (silent wrong output)

1. **Line endings get changed.** `read_text()` turns `\r\n` and `\r` into `\n`. Windows or old Mac text files are silently altered, and chunk positions no longer match the input. Use `read_bytes().decode("utf-8")` instead. (`split_text.py:294`)

2. **TSV content is broken.** Tabs and newlines in a chunk are replaced with the literal text `\t` and `\n`, so reading the TSV back gives wrong content. A real tab and the text `\t` look the same, so it can't be undone. Let `csv.writer` do the quoting and remove the `.replace()` calls. (`split_text.py:278`)

## Should fix (crashes)

3. **Non-UTF-8 input crashes** with a raw error trace. Catch it and show a clear message. (`split_text.py:294`)

4. **Character + token mode + a special token + "reject special tokens" crashes** with a raw error trace. Only the pure-token path is tested for this. Show a clear message instead. (`split_text.py`)

5. **Offline use needs the tiktoken cache.** tiktoken downloads encoding files from the internet on first use and the conda package does not include them. I saw the plan to set `TIKTOKEN_CACHE_DIR` via TPV on EU — that covers EU, but the code comment still points to the deleted `dev_utils` script, and other deployers won't know about the cache requirement. Please update the comment and document it in the tool help. (`split_text.py:108`)

6. **`start_index` changes type.** Following up on the invalid start-index upstream bug you describe in the PR text: the tool currently replaces the number with a long string, so the JSON has integers for some chunks and a string for others. Downstream code that does math on it will crash. Use `null` instead, and keep the warning separate. (`split_text.py:340`)

## Nice to fix

7. **Empty or whitespace-only input gives 0 chunks**, a 0-byte TSV, and an empty collection. No warning. Add a guard. (`split_text.py`)

8. **Token-mode output doesn't say whitespace is kept.** As you note in the PR text, `strip_whitespace` is never applied by the token splitter. But the JSON metadata simply omits the field in token mode, so users can't tell from the output. Add `strip_whitespace: false` to the token-mode metadata, or document it. (`split_text.py:376`)

9. **JSON chunk key differs by mode** (`character_count` vs `token_count`). Downstream readers must branch on `length_unit`. Use one `length` key plus `length_unit`. (`split_text.py:348`)

10. **Minor cleanups:** the hand-rolled token counter copies library logic and may drift on upgrade; the `tiktoken_options` dict is shared by two callers that handle it differently; a dead `else None` branch in `build_splitter`; the whole input is token-counted again after splitting; bad custom-separator escapes crash (`\UFFFFFFFF`) or silently pass through (`\u300`); `--encoding-name` and `--model-name` are not mutually exclusive. The two similar tiktoken blocks in the XML could also be pulled into a shared macro so they don't drift apart later.

## Tests to add

Empty input, whitespace-only input, non-UTF-8 input, CRLF/CR input, character+token mode with a disallowed special token, and a TSV round-trip check (the current `tsv_escape` test only compares bytes to the broken expected file).

tools/langchain_text_splitters/split_text.py
bgruening: fyi: https://github.com/galaxyproject/tools-iuc/pull/8006

arash77:
I fixed the points from my review locally and ran the tests. 16/16 passed before, 20/20 after. Two more real bugs came up while testing.

## Correction: do not use `csv.writer` for item 2

I was wrong there. Galaxy's `tabular` format does not understand CSV quotes, so a chunk with a newline would become several rows and break the table.

Keep the escaping, but make it undoable: escape `\` first, then `\t`, `\r`, `\n`. Then write the row with a plain `"\t".join(...)`.

Note also that after the line ending fix (item 1) a chunk can hold a real `\r`, which most TSV readers treat as a new row. So `\r` needs escaping too.

## Two new bugs

**`csv.writer` breaks any chunk containing a `"`.** This is the one I would fix first. `csv.writer` adds quotes by itself, so `He said "hi" ok` is written as `"He said ""hi"" ok"`. Normal text is full of quotation marks, so the content column is wrong for many real documents, and no unescaping brings it back. Dropping `csv` (see above) fixes this too.

**A separator starting with `-` breaks the tool.** The separator is passed as `--separator '$value'`. A user splitting a markdown file on `---` gets `error: expected one argument` and exit code 2, with nothing useful shown. Writing it as `--separator='$value'` fixes it.

## Minor

- The bad start index is not limited to the case in the PR text. After changing it to `null`, two old tests failed: recursive+gpt2 and token+gpt-5, both with overlap 1. Chunk 2 also gets `-1`. The old `not_has_text` check only passed because the code wrote a long string there. So it hits any token split with overlap. Worth adding to the upstream issue.
- Outputs are written with the system locale encoding while the input is read as UTF-8. Python normally turns a `C` locale into UTF-8, so this only breaks on a machine set to something like ISO-8859-1. Unlikely on our instances, but `encoding="utf-8"` on the writes is a one word fix.
- Item 1 cannot be tested in Galaxy: upload turns `\r\n` into `\n` by default. Same for non-UTF-8 input. Both still matter for files made by other tools, but can only be tested on the script.
- Item 7 changes behaviour: input with only spaces used to run and give empty output, now it stops.
- Items 6, 8 and 9 change the output format. Feel free to take only the bug fixes.
- The `predefined:` separator inside the repeat has no test. All repeat tests use a custom one.

IvoLeist:
> * The bad start index is not limited to the case in the PR text. After changing it to `null`, two old tests failed: recursive+gpt2 and token+gpt-5, both with overlap 1. Chunk 2 also gets `-1`. The old `not_has_text` check only passed because the code wrote a long string there. So it hits any token split with overlap. Worth adding to the upstream issue.

Very interesting finding :nerd_face:  I went through some langchain issues/stopped PRs regarding it (see e.g. https://github.com/langchain-ai/langchain/issues/29884) and apparently this is not easy to fix. So :crossed_fingers:  that they succeed at one point !  Since this start_index information is nice to have (from the users perspective) but in my opinion not critical I would (for now) comment out the failing asserts including a reference to the upstream bug issue. So we/potential future maintainers could reactivate these assert easily ;D

tools/langchain_text_splitters/test-data/tsv_escape.txt
```diff
@@ -1,2 +1,3 @@
 alpha	beta
+C:\path
```
bgruening, line 2: ?

tools/langchain_text_splitters/test-data/tsv_escape_expected.tsv
```diff
@@ -1 +1 @@
-1	alpha\tbeta\nomega\n	17
+1	alpha\tbeta\nC:\\path\nomega\n	25
```
bgruening, line 1: What is this windows path thing?

arash77: `C:\path` is just the shortest recognizable string that contains a literal backslash, and the backslash is what makes the TSV escaping reversible.

arash77: `back\slash` does not work by the way; the collection assertion is a regex, so `\s` there is the whitespace class. `C:\path` is fine because `\p` means nothing in either layer, so let's keep it.

IvoLeist:
Latest update:
- Add NLTKTextSplitter and SpacyTextSplitter (addresses comment: https://github.com/bgruening/galaxytools/pull/1947#discussion_r3714026485
- Add @arash77  as creator

*Note this is so far only a proof of concept. There are no spaCy or NLTK tests yet. So not merge ready.

IvoLeist:
Note, before Arash's PR https://github.com/conda-forge/nltk_data-feedstock/pull/10 is not merged the NLTK TextSplitter can only be tested locally:

First you would need to build the `nltk_data` conda package by cloning the PR`s branch:
https://github.com/arash77/nltk_data-feedstock/tree/update-nltk-data-2026.07.01

Then you could build the package with:
```shell
cd nltk_data-feedstock
python build-locally.py
```
> [!NOTE]
> You need some patience building it can take a while....
> If you encounter a WARNING: Number of parsed outputs does not match detected raw metadata blocks no worries the package has most likely been built correctly

You can run the tests as follows:
```shell
planemo test \
    --conda_dependency_resolution \
    --conda_auto_install \
    --conda_ensure_channels \
    "file:///home/<user>/<path_to_feedstock>/build_artifacts,conda-forge,bioconda" 
```
It should return: 
> All 26 test(s) executed passed.

Or test it in your local galaxy instance by setting the following in your config/galaxy.yml
```yaml
conda_ensure_channels: file:///home/<user>/<path_to_feedstock>/build_artifacts,conda-forge,bioconda
conda_auto_install: true
```
before running it with:
`planemo serve --galaxy_root /home/<user>/<path_to_your_local_galaxy_instance>`

IvoLeist:
To keep everyone in the loop: 
In the meantime Arash's https://github.com/conda-forge/nltk_data-feedstock/pull/10 PR got merged and @bgruening re-triggered the due to GithubAction outage skipped Linux  built: [conda-forge/nltk_data-feedstock/actions/runs/31182366681/job/92878488138](https://github.com/conda-forge/nltk_data-feedstock/actions/runs/31182366681/job/92878488138) :rocket:

IvoLeist:
Latest changes:
- Add strip_whitespace option for SpacyTextSpliiter; 
  - Refer to it in the tool help 
  - Change the params of one test case to strip_whitespace=True
- Wrap strip_whitespace in macro
- Add explanation why strip_whitespace is fixed to false for the NLTK text splitter
- Populate the NLTK language options with all the in punkt_tab available 19 languages
- Set the chunk_overlap default to "0"
- Add a TODO related to galaxy adding by default a "\n" to each upload when there is no terminal new line 
  - Made one test (Number: 9) fail on purpose so we remember to address this before the merge

Regarding the still open TODO this is what I learned so far.
As I have understood this here below is adding the "\n" during galaxy file upload

```python
if last_block and last_block[-1] != NEWLINE_BYTE:
  converted_newlines = True
  i += 1
  fp.write(b"\n")
```
https://github.com/galaxyproject/galaxy/blob/release_26.1/lib/galaxy/datatypes/sniff.py#L166-L169

A simple: `return text.removesuffix("\n") ` leads to 6 six failing tests (6,12,14,15,16,17)

However, just not removing the "\n" potentially messes with the splitting as indicated by our 
> Warning: The NLTK sentence splitter dropped the 1 character(s) after the last sentence

 To reproduce use sentence_nltk_english.txt of our test-data, with target chunk size in characters: 30
 Or run `planemo test --test_index 9`

bgruening:
Does your test data contain windows line breaks?

IvoLeist:
> Does your test data contain windows line breaks?

No, I just double checked:
```shell
cd test-data/
file *    
```
> custom_separator.txt:      ASCII text
> dash_separator.txt:        ASCII text
> langchain_docs_sample.txt: Unicode text, UTF-8 text
> overlap_words.txt:         ASCII text
> pipe_separator.txt:        ASCII text
> recursive_default.txt:     ASCII text
> sentence_nltk_english.txt: ASCII text
> sentence_nltk_german.txt:  ASCII text
> sentence_overlap.txt:      ASCII text
> sentence_spacy.txt:        ASCII text
> sentence_tokens.txt:       ASCII text

`grep -rl $'\r' -- test-data/ | wc -l`
0
`grep -rl $'\r$' -- test-data/ | wc -l`
0

tools/langchain_text_splitters/langchain_text_splitters.xml
IvoLeist: Thanks to @arash77's latest changes (https://github.com/IvoLeist/galaxytools/pull/5)
in which the text stripping logic is moved into main() this limitation of not allowing the removal of leading/trailing whitespaces for the NLTK splitter should no longer be there.

Thus, this help text needs to be updated as soon there are the respective test cases for the NLTK splitter. 

One general remark to be discussed: Since our latest implementation no longer utilises the `strip_whitespace` logic of Langchain I think the help text should reflect this.

tools/langchain_text_splitters/split_text.py
IvoLeist: spaCy/NLTK splitters are now fully functional :rocket:

bgruening (approved):

Sch-Da:
really cool, thanks so much @IvoLeist @arash77 @anuprulez and @bgruening !

PR #1953, opened 2026-08-12
Add miRga3 and DM

SantaMcCloud:
FOR CONTRIBUTOR:


SantaMcCloud:
@bgruening this PR can be reviewed now since it should work, 

When i did the DM my head was more by the exam than doing this....it was a simple mistake after all but it should work now!

bgruening:
Nice, thanks a lot!

PR #1954, opened 2026-08-12
Add tDRmapper

SantaMcCloud:
FOR CONTRIBUTOR:


tools/rna_tools/tdrmapper/tdrmapper.xml
```diff
@@ -0,0 +1,128 @@
...
+        #set $sam = re.sub('[^\w\-_\.]', '_', $sample.element_identifier)
+        ln -s '$sample' '$sam' &&
+
+        TdrMappingScripts.pl
+        '$ref'
+        '$sam'
+
+        &&
+
+        ls . &&
```
bgruening: is that really needed?

tools/rna_tools/tdrmapper/tdrmapper.xml
```diff
@@ -0,0 +1,128 @@
...
+        mv '${sam}.hq_cs.mapped' '$mapped' &&
+        mv '${sam}.hq_cs.mapped.covg.txt' '$mapped_covg' &&
+        mv '${sam}.hq_cs.mapped.covgPre.txt' '$mapped_covg_pre' &&
+        mv '${sam}.hq_cs.mapped.speciesInfo.txt' '$spec_info' &&
+        mv '${sam}.hq_cs.mapped.top50covg.txt' '$top_50' &&
+        mv '${sam}.hq_cs.mapped.top50covgPre.txt' '$pre_top_50'
+
+    ]]></command>
+    <inputs>
+        <param name="reference" type="data" format="fasta" label="Input reference tRNA"/>
```
bgruening: does that also work with fasta.gz? Can we make it work with fasta.gz?

tools/rna_tools/tdrmapper/tdrmapper.xml
SantaMcCloud: It is not needed but it might help with Debugging since the tool has print outputs so the User might know what files were generated.

tools/rna_tools/tdrmapper/tdrmapper.xml
SantaMcCloud: I didnt test it but we can make it work with unzip it befor using it. 

Do you want only gz files or are both fine?

bgruening: both is ok.

tools/rna_tools/tdrmapper/tdrmapper.xml
bgruening: but users can not do anyting woth this information.
The `ls` might be good for developers but imho not for users. 

```suggestion
```

SantaMcCloud:
@bgruening should be good now after the test. I did add fasta.gz suport via command line

bgruening:
Merci!

PR #1967, opened 2026-08-26
[kraken _segment/_ocr] improve help texts and add citations of the on galaxy EU provided models

IvoLeist:
The changes below are based on the initial feedback of @Sch-Da 
- Point to Zenodo and Kraken's model catalog in the model selection help and each tool help
~- Add to the help section the supported model formats~
~- Add segmentation/ocr models citations of the models provided on Galaxy EU~

FOR CONTRIBUTOR:

Notes for reviewers:
All tests pass locally, they fail in GitHub CI due to the new datatypes added recently to the branch release_26.1 https://github.com/galaxyproject/galaxy/pull/23088. I expect all tests to pass here as soon the GitHub CI is no longer using Galaxy 26.0.1 (?) for testing.


tools/kraken_ocr/kraken_ocr.xml
```diff
@@ -191,29 +191,34 @@
...
+kraken is an open-source, turn-key automatic text recognition (ATR)
 system optimized for historical and non-Latin script writing.
 This tool only exposes Kraken's OCR command.
 
-OCR is the process of extracting text from images. 
-Single-line images can be processed directly. 
-For document images, a segmentation file in [ALTO](https://www.loc.gov/standards/alto/) 
-or [PAGE XML](https://github.com/PRImA-Research-Lab/PAGE-XML/wiki) format is required. 
-This additional input file can be produced by the Kraken segmentation tool, 
+OCR is the process of extracting text from images.
```
Sch-Da:
```suggestion
Optical Character Recognition (OCR) is the process of extracting text from images.
```
Maybe in case this has not been mentioned before, we could add this in the text?
But probably you already did :)

tools/kraken_ocr/macros.xml
```diff
@@ -67,6 +67,15 @@
 
             ocr
     ]]></token>
+    <token name="@MODEL_SELECTION_HELP@"><![CDATA[
+either upload your own model in
+[mlmodel](https://apple.github.io/coremltools/mlmodel/index.html) or
+[safetensors](https://huggingface.co/docs/safetensors/index) format,
```
Sch-Da: nice addition, thanks!

tools/kraken_ocr/macros.xml
```diff
@@ -67,6 +67,15 @@
...
+or select a model available on the Galaxy instance.]]></token>
+    <token name="@MODEL_SELECTION_METADATA_HELP_TEXT@"><![CDATA[
+For model metadata, such as covered languages, scripts and licensing information, 
+consult Zenodo, Kraken's model catalog (https://htrmopo.inria.fr),
```
Sch-Da: Is it possible to make this a hyperlink? If not or if I am missing it, never mind.
Thanks a lot for all your implementations @IvoLeist

tools/kraken_ocr/kraken_ocr.xml
IvoLeist: Thanks, good catch ! I included your suggestion :)

tools/kraken_ocr/macros.xml
IvoLeist: Now in the selection box help as well as in help section there is [ Kraken's model catalog](https://htrmopo.inria.fr). Good catch :)

tools/kraken_ocr/macros.xml
```diff
@@ -87,9 +100,42 @@
...
+    </macro>
+    <macro name="segmentation_models_citations">
+        <!-- this macro is galaxy instance specific, 
+            here pre-trained segmentation models provided by Galaxy EU -->
+        <citation type="doi">10.5281/zenodo.10890967</citation>
+    </macro>
+    <macro name="ocr_models_citations">
+        <!-- this macro is galaxy instance specific, 
+            here pre-trained ocr models provided by Galaxy EU -->
+        <citation type="doi">10.5281/zenodo.12743230</citation>
```
bgruening: I don't think we can do this, sorry.

Models can be added and removed by admins and should not be hardcoded to the tools.
The tool citations are for the tools, not the data.

The only way I can suggest is to put the DOI into the location file and mention in the tool help that users should cite the models that they are using as well.

IvoLeist: Fair enough, thanks for the quick clarification :) Actually the DOI is already in the location file but in its own column and not as part of the `display_name`. On top, I  would refrain from including the DOI in the `display_name` since you cannot copy a value in the selection box. 

So for now I have removed the model citations until we come up with a better idea :thinking: 
Nevertheless, the friendly reminder that users should cite the used model is already included :v:

Sch-Da: Thanks for including the citation to the used model. 
My reason to suggest adding the DOI - even though I agree that the idea in the display name is not the neatest one, at least it would contain some info for reference? Because How do I know what to quote without even having that?
This is way trickier than I had hoped for

bgruening: We can not copy from the display name, but we can copy from the `i` dataset page, where the DOI would be also listed, maybe ... we need to check.

Another idea, that I also don't like is that we could create a new dataset that contains a text with the DOI, cite this ... We don't do this for any other tool ...

IvoLeist: I guess in theory we could make the e.g. DOI somehow end up in the input parameters
image

But as discussed with @Sch-Da it might be confusing for users to show them an input parameter which the  user never have set but got populated "automagically" for them? Or are there any other tools which do this?
Thus, for now maybe indeed the best option is to include the DOI/GitHub Repo in the selection box: 

If DOI present:
image

No DOI found:
image

Note, the DOI/GitHub repo in the `display_name` column is not populated automatically
but would need to be included manually in the respective .loc file

tools/kraken_ocr/macros.xml
```diff
@@ -67,6 +67,19 @@
...
+[safetensors](https://huggingface.co/docs/safetensors/index) format,
+or select a model available on the Galaxy instance.]]></token>
+    <token name="@MODEL_METADATA_HELP_PART1@">For model metadata, such as covered languages, scripts, and license information, consult Zenodo or</token>
+    <token name="@MODEL_METADATA_HELP_PART2@">Please remember to cite the model you use.</token>
+    <token name="@MODEL_METADATA_HELP_HTML@"><![CDATA[
+        @MODEL_METADATA_HELP_PART1@ <a href="https://htrmopo.inria.fr">Kraken's model catalog</a>. @MODEL_METADATA_HELP_PART2@
+    ]]></token>
+    <token name="@MODEL_METADATA_HELP_MD@"><![CDATA[
+        @MODEL_METADATA_HELP_PART1@ [Kraken's model catalog](https://htrmopo.inria.fr). @MODEL_METADATA_HELP_PART2@
+    ]]></token>
```
bgruening: Do you think this is easy to maintain?
Maybe here its ok to just copy and paste the text in both tools? I find it hard to understand how this is composed now.

Have you tried https://planemo.readthedocs.io/en/latest/writing_advanced.html#parameterizing-xml-macros-with-tokens ... this lets you write some text and pass variables to macros, maybe that makes it easier to read and maintain?

bgruening (approved):

bgruening:
Thanks @IvoLeist

Sch-Da:
thank you @IvoLeist !

PR #1968, opened 2026-08-26
Add paramter to generate aux files

razmia02:
FOR CONTRIBUTOR:
* [ ] License permits unrestricted use (educational + commercial)

### DIFF
--- a/tools/salmon/salmonquant.xml
+++ b/tools/salmon/salmonquant.xml
@@ -19,11 +19,12 @@
             @salalign@
         #end if
         @salquant@
-        #if $quant_type.qtype == "reads" and $quant_type.bam_options.writeMappings:
-            && @bam_sort@ ./output/samout.sam
+        #if $generate_aux
+        #if $quant_type.qtype == "reads"
+        && tar -cf aux_files.tar -C output aux_info libParams lib_format_counts.json
+        #else
+        && tar -cf aux_files.tar -C output aux_info libParams
         #end if
-        #if $quant_type.qtype == "alignment" and $quant_type.sampleOut:
-            && @bam_sort@ ./output/postSample.bam
         #end if
         ]]>
     </command>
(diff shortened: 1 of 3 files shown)


tools/salmon/salmonquant.xml
```diff
@@ -19,11 +19,12 @@
...
         @salquant@
-        #if $quant_type.qtype == "reads" and $quant_type.bam_options.writeMappings:
-            && @bam_sort@ ./output/samout.sam
+        #if $generate_aux
+        #if $quant_type.qtype == "reads"
+        && tar -cf aux_files.tar -C output aux_info libParams lib_format_counts.json
+        #else
+        && tar -cf aux_files.tar -C output aux_info libParams
         #end if
-        #if $quant_type.qtype == "alignment" and $quant_type.sampleOut:
```
bgruening, line 25: I'm wondering when those outputs gets lost, but it seems we are not needing them.

razmia02: To clarify the motivation: the current Salmon wrapper only exposes quant.sf, discarding the aux_info/ and libParams/ directories Salmon produces alongside it. Those auxiliary files are what MultiQC's Salmon module actually parses (meta_info.json, lib_format_counts.json, flenDist.txt), so without them MultiQC on Galaxy has nothing to report on for Salmon. This is also why the Salmon integration in galaxyproject/tools-iuc's MultiQC wrapper currently sits unfinished/commented out.
There was actually a previous attempt at this exact thing in [PR#1083](https://github.com/bgruening/galaxytools/pull/1083), which stalled before merging, partly over concerns about exposing too many files and cluttering the tool interface. I've tried to address that directly here: rather than exposing each aux file as its own dataset, this bundles just the minimal set MultiQC needs into a single opt-in tarball output (generate_aux, off by default), so users who don't need it see no change at all.
Happy to adjust the approach if you'd prefer a different output structure entirely.

bgruening (approved):
Thanks a lot!

PR #1992, opened 2026-09-17
Export encoded vocabulary

anuprulez:


tools/sklearn/label_encoder.xml
```diff
@@ -22,17 +23,30 @@
     </inputs>
     <outputs>
         <data name="outfile" format="tabular"/>
+        <data name="outfile_encoding_vocabulary" format="json"/>
```
nilchia: no label for the outputs? What would be the default label then?

nilchia (approved):

tools/sklearn/label_encoder.xml
anuprulez: thanks, fixed in https://github.com/bgruening/galaxytools/pull/1992/commits/b4ee47778084d42fb6b3e29e09e7386bec25c33b

nilchia (approved):

anuprulez:
thanks @nilchia

anuprulez:
@bgruening do we know what is the problem ?  It seems tools from @Sch-Da also have similar issues 

thank you!

